PR #2, opened 2018-08-14
replace generic updatestate with run

leijurv:


HowardStark:
After very, very long talks we arrived at a conclusion that while everyone agrees the system now is less than perfect, we can’t find a suitable replacement that encapsulates the functionality, readability, and use ability of the current system. 

Anything else to add on this discussion @leijurv?

PR #66, opened 2018-08-21
Map Generation

HowardStark:
### Overview

Simple implementation of map generation that, at minimum, achieves feature parity with the detail of the built-in Minecraft map item.

### DIFF
--- a/src/main/java/baritone/bot/Baritone.java
+++ b/src/main/java/baritone/bot/Baritone.java
@@ -23,7 +23,12 @@ import baritone.bot.behavior.impl.MemoryBehavior;
 import baritone.bot.behavior.impl.PathingBehavior;
 import baritone.bot.event.GameEventHandler;
 import baritone.bot.utils.InputOverrideHandler;
+import net.minecraft.client.Minecraft;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -47,6 +52,7 @@ public enum Baritone {
     private InputOverrideHandler inputOverrideHandler;
     private Settings settings;
     private List<Behavior> behaviors;
+    private File dir;
 
     /**
      * Whether or not Baritone is active
@@ -64,6 +70,13 @@ public enum Baritone {
         behaviors.add(PathingBehavior.INSTANCE);
         behaviors.add(LookBehavior.INSTANCE);
         behaviors.add(MemoryBehavior.INSTANCE);
+
+        this.dir = new File(Minecraft.getMinecraft().gameDir, "baritone");
+        if (!Files.exists(dir.toPath())) {
+            try {
+                Files.createDirectories(dir.toPath());
+            } catch (IOException ignored) {}
+        }
 
         this.active = true;
         this.initialized = true;
@@ -100,4 +113,6 @@ public enum Baritone {
     public static Settings settings() {
         return Baritone.INSTANCE.settings; // yolo
     }
+
+    public final File getDir() { return this.dir; }
 }
--- a/src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
+++ b/src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
@@ -31,6 +31,7 @@ import baritone.bot.pathing.movement.ActionCosts;
 import baritone.bot.pathing.movement.CalculationContext;
 import baritone.bot.pathing.movement.Movement;
 import baritone.bot.utils.pathing.BetterBlockPos;
+import baritone.map.MapChunk;
 import net.minecraft.util.math.BlockPos;
 
 import java.util.ArrayList;
@@ -170,6 +171,10 @@ public class ExampleBaritoneControl extends Behavior {
                     return;
                 }
             }
+        }
+        if(msg.toLowerCase().equals("map")) {
+            MapChunk chunk = new MapChunk(world().getChunk(player().getPosition()));
+            chunk.writeImage();
         }
         if (Baritone.settings().byLowerName.containsKey(msg.toLowerCase())) {
             Settings.Setting<?> setting = Baritone.settings().byLowerName.get(msg.toLowerCase());
--- a/src/main/java/baritone/map/MapChunk.java
+++ b/src/main/java/baritone/map/MapChunk.java
@@ -1,0 +1,72 @@
+package baritone.map;
+
+import baritone.bot.Baritone;
+import baritone.bot.utils.BlockStateInterface;
+import baritone.bot.utils.pathing.BetterBlockPos;
+import net.minecraft.block.material.MapColor;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.math.BlockPos.MutableBlockPos;
+import net.minecraft.world.chunk.Chunk;
+import org.lwjgl.util.Color;
+
+import javax.imageio.ImageIO;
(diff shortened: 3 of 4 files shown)


src/main/java/baritone/map/MapChunk.java
```diff
@@ -0,0 +1,98 @@
...
+     * in Minecraft's maps.
+     *
+     * @param x relative x coordinate to the start of the chunk
+     * @param z relative z coord.
+     * @return Integer RGB value with contextual shading
+     */
+    public int getColor(int x, int z) {
+        int chunkX = chunk.getPos().getXStart() + x;
+        int chunkZ = chunk.getPos().getZStart() + z;
+        MutableBlockPos blockPos = new MutableBlockPos(chunkX, chunk.getHeight(new BetterBlockPos(chunkX, 0, chunkZ)), chunkZ);
```
leijurv: This can just be BlockPos, no?

src/main/java/baritone/map/MapChunk.java
```diff
@@ -0,0 +1,98 @@
...
+        MapColor color = BlockStateInterface.get(blockPos).getMapColor(chunk.getWorld(), blockPos);
+
+        // The chunk heightMap returns the first non-full block above the surface, including bushes.
+        // We have to check this and shift our block down by one if the color is "air" (as in, there's no real block there)
+        int colorValue = color != MapColor.AIR ? color.colorValue : BlockStateInterface.get(blockPos.down()).getMapColor(chunk.getWorld(), blockPos).colorValue;
+        int red = (colorValue >> 16) & 0xFF;
```
leijurv: Idk, I'm a big fan of something like:

```
Color color = new Color(colorValue);
int red = color.getRed();
int green = color.getGreen();
int blue = color.getBlue();
```

just for clarity on what's going on, bitshifts and &s are a little unclear

src/main/java/baritone/map/MapChunk.java
```diff
@@ -0,0 +1,98 @@
...
+        int green = (colorValue >> 8) & 0xFF;
+        int blue = colorValue & 0xFF;
+
+        // Now we get the proper block for the position one to the north.
+        BlockPos offset = blockPos.offset(EnumFacing.NORTH);
+        offset = new BlockPos(offset.getX(), chunk.getHeight(offset), offset.getZ());
+        // And once again, check to make sure we have an actual colored block an not "air"
+        if(BlockStateInterface.get(offset).getMapColor(chunk.getWorld(), offset) == MapColor.AIR)
```
leijurv: b r a c e s

src/main/java/baritone/map/MapChunk.java
```diff
@@ -0,0 +1,98 @@
...
+            offset = offset.down();
+
+        // We start the heightCoef at 0.8 to darken the colors similar to vanilla Minecraft's. 
+        float heightCoef = 0.8f + (Math.signum(blockPos.getY() - offset.getY()) * 0.2f);
+        int newColor = Math.min(255, (int) (red * heightCoef));
```
leijurv: Perhaps something like

```
red = Math.min(255, (int) (red * heightCoeff));
green = Math.min(255, (int) (green * heightCoeff));
blue = Math.min(255, (int) (blue * heightCoeff));

return new Color(red, green, blue).getRGB();
```

src/main/java/baritone/bot/Baritone.java
```diff
@@ -67,6 +73,13 @@ public synchronized void init() {
             registerBehavior(MemoryBehavior.INSTANCE);
         }
 
+        this.dir = new File(Minecraft.getMinecraft().gameDir, "baritone");
```
leijurv: Same directory as chuck caching, so that it's separate per-world pls

src/main/java/baritone/map/MapChunk.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public MapChunk(Chunk chunk) {
+        this.chunk = chunk;
+    }
+
+    public void writeImage() {
+        Path image = getImagePath();
+        if(!Files.exists(image.getParent())) {
+            try {
+                Files.createDirectory(image.getParent());
```
leijurv: Redundant with the createDirectories call in Baritone.java, no? Probably just this one would work

src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
```diff
@@ -177,6 +178,10 @@ public void onSendChatMessage(ChatEvent event) {
                 }
             }
         }
+        if(msg.toLowerCase().equals("map")) {
+            MapChunk chunk = new MapChunk(world().getChunk(player().getPosition()));
```
leijurv: We use `playerFeet()` instead of `player().getPosition()` everywhere else because for some strange reason, `getPosition()` adds 0.5 to x, y, and z before casting to integer. `playerFeet` gives the exact block the player sees when you look straight down, `getPosition` will give the one that's next to it if you're in the top (coordinate-wise) half of the block.

src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
```diff
@@ -177,6 +178,10 @@ public void onSendChatMessage(ChatEvent event) {
...
+            chunk.writeImage();
```
leijurv:
```
event.cancel();
return;
```

src/main/java/baritone/map/MapChunk.java
HowardStark: No, this is creating the "map" directory.

src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
HowardStark: This is _purely_ for testing. You're right that it's not great, but this will be removed in a later update.

src/main/java/baritone/bot/utils/ExampleBaritoneControl.java
HowardStark: Again, just for testing. There won't be a command in the final one.

src/main/java/baritone/map/MapChunk.java
HowardStark: Let's not introduce personal preference in a PR (unless there are literally no other cases of no-bracket if's in the project left, in which case my bad).

src/main/java/baritone/map/MapChunk.java
HowardStark: This, in the context of https://github.com/cabaletta/baritone/pull/66#discussion_r211692811 is a good change.

leijurv:
code LGTM from a testing point of view, once it's integrated with chunk caching, merge it. 

I'm not sure if that integration should be just storing the PNGs in the same folder, or maybe shoving them into the same file as the 2-bit representation. 

Interestingly enough, what if chunk caching stored the block types of the very top layer, which is 16x16=256 blocks. You get the heightmap from the 2-bit shape. So no need to store all the PNGs at all.
In fact, storing the blocktypes of the very top block in each column would be great for more accurate pathing huh.

HowardStark:
I've created issue #71 to implement the above. Sick idea.

HowardStark:
Example of map rendering progress;

full-map

leijurv:
The web UI should allow you to click a spot and have the client start pathing to it. It should show your current position and draw your current path segment. 

Also it would be cool if you could see your status (inventory, hunger, health). 

It would also be really cool if the web UI could connect to multiple Baritone instances =O

PR #96, opened 2018-08-26
Cleaned up 7 duplicate switch statements

ZeroMemes:
I get aids every time I look at it. The behavior that is yielded from my changes is the same as it was before.


ZeroMemes:
YOU KNOW I HAD TO

PR #140, opened 2018-09-07
MC version support

leijurv:


src/main/java/baritone/utils/ToolSet.java
```diff
@@ -86,15 +87,22 @@ private double calculateStrVsBlock(byte slot, IBlockState state) {
         // Calculate the slot with the best item
         ItemStack contents = player().inventory.getStackInSlot(slot);
 
+        // In 1.10 null item stacks were a thing, no such thing as empty ones.
+        if (contents == null)
+            return 0;
```
leijurv: This is incorrect, 0 means instant break.

PR #141, opened 2018-09-07
Multiple Version Support

ZeroMemes:
Currently supports 1.9.4, 1.10.2, 1.11.2, and 1.12.2


Mhowser:
Any chance on seeing support for 1.13+ versions?

0-x-2-2:
@Mhowser [ForgeGradle](https://github.com/MinecraftForge/ForgeGradle/tree/FG_3.0) 3.0 is very unstable and [MixinGradle](https://github.com/SpongePowered/MixinGradle) won't work with new forgegradle till they update it and [Mixin](https://github.com/SpongePowered/Mixin) also needs a update to work with [ModLauncher](https://github.com/cpw/modlauncher) aka the replacement for [LegacyLauncher/LaunchWrapper](https://github.com/Mojang/LegacyLauncher). so it won't be for a while and it will probably be after forge 1.13 is released some time in 2019.

Mhowser:
Ah gotcha, thanks for explaining.

ZeroMemes:
Due to the 1.13.2 update, the approach in this PR would not be viable. Ideally, we could have adapters for MovementHelper given the massive changes made by updating to 1.13.2, but I'm not going to continue this branch any further.

PR #221, opened 2018-10-13
Proguard Gradle Build Magic

ZeroMemes:
replace dummy scripts with universal distribution meme that works on all platforms!!!


ZeroMemes:
😏

Issue #227, opened 2018-10-21
Tool selection choice is ineffecient


ave4224:
If two items break the block with equal speed, the leftmost one is used. Sometimes a pick will be used on dirt.

PR #228, opened 2018-10-21
Fixed Toolset issue #227

ave4224:


src/main/java/baritone/behavior/MineBehavior.java
```diff
@@ -53,7 +53,8 @@
     private List<BlockPos> locationsCache;
     private int quantity;
 
-    private MineBehavior() {}
```
leijurv: no, we aren't changing the code style

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -62,7 +62,8 @@
 
     private final LinkedBlockingQueue<PathEvent> toDispatch = new LinkedBlockingQueue<>();
 
-    private PathingBehavior() {}
+    private PathingBehavior() {
```
leijurv: no, we aren't changing the code style

src/main/java/baritone/cache/CachedRegion.java
```diff
@@ -95,9 +95,7 @@ public final boolean isCached(int x, int z) {
                 if (locs == null) {
                     continue;
                 }
-                for (BlockPos pos : locs) {
-                    res.add(pos);
-                }
+                res.addAll(locs);
```
leijurv: This is a key inner loop for minebehavior, this is actually faster to iterate over the linked list one at a time, as opposed to addAll which converts to a [] internally.

src/main/java/baritone/cache/CachedWorld.java
```diff
@@ -120,8 +120,7 @@ public final boolean isCached(int blockX, int blockZ) {
                     int regionZ = zoff + playerRegionZ;
                     CachedRegion region = getOrCreateRegion(regionX, regionZ);
                     if (region != null) {
-                        // TODO: 100% verify if this or addAll is faster.
-                        region.getLocationsOf(block).forEach(res::add);
+                        res.addAll(region.getLocationsOf(block));
```
leijurv: Same as CachedRegion, this is actually faster

src/main/java/baritone/pathing/calc/AStarPathFinder.java
```diff
@@ -40,9 +40,9 @@
  */
 public final class AStarPathFinder extends AbstractNodeCostSearch implements Helper {
 
-    private final Optional<HashSet<Long>> favoredPositions;
```
leijurv: Why would you replace an optional with a nullable lol

src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -74,23 +74,13 @@ public CalculationContext(ToolSet toolSet) {
...
     public boolean canBreakAt(int x, int y, int z) {
-        if (!allowBreak()) {
-            return false;
-        }
-        if (isPossiblyProtected(x, y, z)) {
-            return false;
-        }
-        return true;
+        return allowBreak() && !isPossiblyProtected(x, y, z);
     }
```
leijurv: These changes to canBreakAt and canPlaceThrowawayAt literally make it harder to read, no.

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -197,7 +197,7 @@ static boolean isDoorPassable(BlockPos doorPos, BlockPos playerPos) {
     }
 
     static boolean isGatePassable(BlockPos gatePos, BlockPos playerPos) {
-        if (playerPos.equals(gatePos)) {
+        if (!playerPos.equals(gatePos)) {
```
leijurv: wtf?

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -295,10 +295,7 @@ static boolean canWalkOn(int x, int y, int z, IBlockState state) {
             }
             return true;
         }
-        if (block instanceof BlockStairs) {
-            return true;
-        }
-        return false;
+        return block instanceof BlockStairs;
```
leijurv: I know codacy loves this, but in the context of a chain of

```
if(cond1)
    return true
if(cond2)
    return true
if(cond3)
    return true
```

it's better if they all follow the pattern, instead of having a special case for the last one

src/main/java/baritone/pathing/path/PathExecutor.java
```diff
@@ -426,7 +424,7 @@ private void sprintIfRequested() {
         player().setSprinting(false);
     }
 
-    private static boolean canSprintInto(IMovement current, IMovement next) {
+    private static boolean canSprintInto(IMovement current, IMovement next) { // TODO: Move this into IMovement
```
leijurv: no, why?

src/main/java/baritone/utils/ExampleBaritoneControl.java
```diff
@@ -77,7 +77,7 @@ public void onSendChatMessage(ChatEvent event) {
         }
     }
 
-    public boolean runCommand(String msg) {
+    private boolean runCommand(String msg) {
```
leijurv: Impact uses this for .b

src/main/java/baritone/utils/ExampleBaritoneControl.java
```diff
@@ -298,7 +298,8 @@ public boolean isInGoal(BlockPos pos) {
                 MineBehavior.INSTANCE.mine(quantity, block);
                 logDirect("Will mine " + quantity + " " + blockTypes[0]);
                 return true;
-            } catch (NumberFormatException | ArrayIndexOutOfBoundsException | NullPointerException ex) {}
```
leijurv: no, we aren't changing the code style

src/main/java/baritone/utils/PathRenderer.java
```diff
@@ -58,7 +58,8 @@
     private static final Tessellator TESSELLATOR = Tessellator.getInstance();
     private static final BufferBuilder BUFFER = TESSELLATOR.getBuffer();
 
-    private PathRenderer() {}
+    private PathRenderer() {
```
leijurv: no, we aren't changing the code style

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,87 +23,140 @@
...
+    private final Function<Block, Double> backendCalculation = this::getBestDestructionTime;
+
+    /**
+     * Using the best tool on the hotbar, how long would it take to mine this block
+     *
+     * @param state the blockstate to be mined
+     * @return how long it would take in ticks
+     */
+    public double getStrVsBlock(IBlockState state) {
+        return breakStrengthCache.computeIfAbsent(state.getBlock(), backendCalculation) * potionAmplifier();
```
leijurv: No need to put this into a `private final Function<Block, Double>`, you can just do `breakStrengthCache.computeIfAbsent(state.getBlock(), this::getBestDestructionTime)` and then it's a synthetic method directly to the class instead of to the Function synthetic interface in the `backendCalculation` field

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,87 +23,140 @@
...
+        IBlockState blockState = b.getDefaultState();
         for (byte i = 0; i < 9; i++) {
-            double v = calculateStrVsBlock(i, b);
+            ItemStack itemStack = player().inventory.getStackInSlot(i);
+            double v = calculateStrVsBlock(itemStack, blockState);
             if (v > value || value == -1) {
                 value = v;
                 best = i;
+                materialCost = getMaterialCost(itemStack);
+            } else if (Math.abs(v - value) < 0.00001) {
```
leijurv: Why not == if they're supposed to be identically calculated?

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,87 +23,140 @@
...
+                int c = getMaterialCost(itemStack);
+                if (c < materialCost) {
```
leijurv: If materialCost starts at -2 won't this never happen or something

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,87 +23,140 @@
...
      * A cache mapping a {@link Block} to how long it will take to break
      * with this toolset, given the optimum tool is used.
      */
-    private Map<Block, Double> breakStrengthCache = new HashMap<>();
+    private final Map<Block, Double> breakStrengthCache = new HashMap<>();
+
+    /**
+     * My buddy leijurv owned me so we have this to not create a new lambda instance.
+     */
+    private final Function<Block, Double> backendCalculation = this::getBestDestructionTime;
```
leijurv: delet this

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,87 +23,140 @@
...
-    public double getStrVsBlock(IBlockState state) {
-        Double strength = this.breakStrengthCache.get(state.getBlock());
-        if (strength != null) {
-            // the function will take this path >99% of the time
-            return strength;
-        }
-        double str = calculateStrVsBlock(getBestSlot(state), state);
-        this.breakStrengthCache.put(state.getBlock(), str);
-        return str;
+    public Double getBestDestructionTime(Block b) {
```
leijurv: why `Double`?

src/main/java/baritone/pathing/calc/AStarPathFinder.java
ZeroMemes: This is probably fine because we're using this ``Optional`` as a parameter, which is not the intended use of the class. Alternatively, we could pass ``orElse(null)`` to the method that has the Optional parameter.

src/main/java/baritone/cache/CachedWorld.java
ZeroMemes: IntelliJ ``// noinspection ...`` when

src/main/java/baritone/utils/ExampleBaritoneControl.java
ZeroMemes: We should find an elegant way to expose this in the API

src/main/java/baritone/utils/ToolSet.java
leijurv: Also, I'm not 100% sure about having potion effects here like this. I think it's good that it's it's own function but I think it can be cached, no? What's the point of pulling it out of the cache? Like, if you have haste at the start of path calculation but it runs out within the 500ms before the timeout? This toolset object is used only for one calculation, because we can't assume that the tools they have on their hotbar (or their potion effects) haven't changed from the last path calculation

src/main/java/baritone/pathing/calc/AStarPathFinder.java
leijurv: Ok I take this one back, this is probably better

src/main/java/baritone/utils/ToolSet.java
ave4224: The cache is persistant across paths. Each segment should use updated potion statuses. I'll work on a better way of making this work. Ideally it would know how long it would take to get to a node and whether the effect is still active by then, but that's a bit over complicated. However storing the result for block values persistantly is an issue across segments. Each segment should recalculate the toolset map.

src/main/java/baritone/utils/ToolSet.java
ave4224: That's an artifact.

ave4224:
This has gotten way too polluted with the reformatted code. I'm closing this and starting over with JUST the toolkit changes.

PR #233, opened 2018-10-21
Toolkit fixes

ave4224:


src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,109 @@
...
-    private Map<Block, Double> breakStrengthCache = new HashMap<>();
+    private final Map<Block, Double> breakStrengthCache;
+
+    /**
+     * My buddy leijurv owned me so we have this to not create a new lambda instance.
+     */
+    private final Function<Block, Double> backendCalculation;
+
+    public ToolSet() {
+        super();
```
leijurv: what? lol 
helper has no constructor?

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,109 @@
...
+        breakStrengthCache = new HashMap<>();
+
+        if (Baritone.settings().considerPotionEffects.get()) {
```
leijurv: This is not worth it, returning a double gets boxed into a Double to pass through this function, we can just have an if like this to apply the potion modifier or not, no need to do this fancy business with composing the amplifier. It is quite functional though.

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,109 @@
...
+        IBlockState blockState = b.getDefaultState();
         for (byte i = 0; i < 9; i++) {
-            double v = calculateStrVsBlock(i, b);
+            ItemStack itemStack = player().inventory.getStackInSlot(i);
+            double v = calculateStrVsBlock(itemStack, blockState);
             if (v > value || value == -1) {
                 value = v;
                 best = i;
+                materialCost = getMaterialCost(itemStack);
+            } else if (Math.abs(v - value) < 0.00001) {
```
leijurv: literally don't do this... if it's computed the exact same way you'll get the exact same output

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -84,49 +135,55 @@ public double getStrVsBlock(IBlockState state) {
...
+        }
+        return speed;
+    }
+
+    /**
+     * Calculates any modifier to breaking time based on status effects.
+     *
+     * @return a double to scale block breaking speed.
+     */
+    private double potionAmplifier() {
```
leijurv: epic!

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,109 @@
...
      *
      * @param b the blockstate to be mined
-     * @return a byte indicating the index in the tools array that worked best
+     * @return A byte containing the index in the tools array that worked best
      */
-    public byte getBestSlot(IBlockState b) {
+    public byte getBestSlot(Block b) {
         byte best = 0;
         double value = -1;
+        int materialCost = -2;
```
leijurv: why?? nothing will ever be less than this, does this even work??

src/main/java/baritone/utils/ToolSet.java
ave4224: Here for style?

src/main/java/baritone/utils/ToolSet.java
ave4224: The reason is this happens once ever. Otherwise on every block type you have to check the potion status.

ave4224: It also means you get to skip one or two condition jumps on every block strength calculation.

src/main/java/baritone/utils/ToolSet.java
ave4224: This works because it still changes on the first time v > value

src/main/java/baritone/utils/ToolSet.java
ave4224: I'm not sure what you're asking me to do... This is the final part of your review.

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,108 @@
...
      * @param b the blockstate to be mined
-     * @return a byte indicating the index in the tools array that worked best
+     * @return A byte containing the index in the tools array that worked best
      */
-    public byte getBestSlot(IBlockState b) {
+    public byte getBestSlot(Block b) {
         byte best = 0;
-        double value = -1;
+        double value = Double.NEGATIVE_INFINITY;
+        int materialCost = Integer.MIN_VALUE;
```
leijurv: If you're certain your code will never use a value, initialize it like `int materialCost;`, it's an indication to the programmer after you that you don't use it at all.

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -23,58 +23,108 @@
...
      *
      * @param b the blockstate to be mined
-     * @return a byte indicating the index in the tools array that worked best
+     * @return A byte containing the index in the tools array that worked best
      */
-    public byte getBestSlot(IBlockState b) {
+    public byte getBestSlot(Block b) {
         byte best = 0;
-        double value = -1;
+        double value = Double.NEGATIVE_INFINITY;
```
leijurv: This one is okay as is since you do compare against it, and use the initial value

src/main/java/baritone/utils/ToolSet.java
leijurv: Ah sorry, should have been more clear. 
I'd prefer if you had a function in this class, like `fullCalculation` or something idk that was just

```
private void fullCalculation(Block block) {  
  double ret = getBestDestructionTime(block);
  if (Baritone.settings().considerPotionEffects.get()) {
    ret *= potionAmplifier();
  }
  return ret;
}
```

And then instead of these shenannigans in the constructor, getStrVsBlock could just be `return breakStrengthCache.computeIfAbsent(state.getBlock(), this::fullCalculation);`

PR #243, opened 2018-10-28
Add alias "forward" to "thisway"

SylveonBottle:
I can't for the life of me remember the command "thisway", I'm literally opening up an old youtube video on Baritone just to remember what command the recorder used.

I'll instinctual type `forward 100` (or `.b forward 100` in Impact's case) when I want Baritone to path forwards.

### DIFF
--- a/src/main/java/baritone/utils/ExampleBaritoneControl.java
+++ b/src/main/java/baritone/utils/ExampleBaritoneControl.java
@@ -299,7 +299,7 @@ public class ExampleBaritoneControl extends Behavior implements Helper {
             logDirect("Started mining blocks of type " + Arrays.toString(blockTypes));
             return true;
         }
-        if (msg.startsWith("thisway")) {
+        if (msg.startsWith("thisway") || msg.equals("forward")) {
             try {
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/utils/ExampleBaritoneControl.java
```diff
@@ -299,7 +299,7 @@ public boolean isInGoal(BlockPos pos) {
             logDirect("Started mining blocks of type " + Arrays.toString(blockTypes));
             return true;
         }
-        if (msg.startsWith("thisway")) {
+        if (msg.startsWith("thisway") || msg.equals("forward")) {
```
LeafHacker: I'd be inclined to start using regexes if you want to support multiple aliases:

```suggestion
        if (msg.matches("^(thisway|forward).*")) {
```

SylveonBottle:
Honestly I just copy/pasted a snippet of code without looking into it much, making sure it wouldn't break at the most. I'm just wanting "forward" as an alias, and probably won't look into it any further than that.

Mhowser:
bump

leijurv:
fine https://github.com/cabaletta/baritone/commit/29a9e6e1d1d27bf3bfdc715d756ef9a77e362c1b

PR #245, opened 2018-11-04
Pathing processes WIP

### LINKED ISSUE #241: Change .goto <block_type> to be persistent
Right now it just scans the world for that block type and sets the goal. However, it doesn't update this when a cached chunk becomes actually loaded. So, on your way to a cached portal location you would walk right by a portal that's newly created since the last time you were there, and it would ignore it. Perhaps this needs some sort of GotoBehavior, I'm not sure, but whenever a segment ends it should rescan the world for that type (like what MineBehavior does but less frequent).

leijurv:
fixes #241


leijurv:
Still TODO: ideally somehow each process could be notified if the path it requested failed to calculate? And if the requested pause / cancel has actually gone through and happened?

PR #267, opened 2018-11-22
add a secondary failure cutoff

### LINKED ISSUE #262: Pathing fails the first time
Not much more to explain, when you load a world and set a goal, `path` instantly fails to calculate once, then succeeds when you run it again. Quite strange.

leijurv:
Unintentionally fixes #262 I have no idea why.

Anyway this adds a secondary cutoff. 

If you tell it to path, it will start executing after 500ms if it's found a path, but can continue calculating up to 2000ms if there is no valid (long enough) path found. Same goes for planning ahead, with 4000ms and 5000ms. This helps it get started as fast as possible, without compromising on its ability to find paths in really tricky situations that require lots of backtracking.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -247,7 +247,9 @@ public class Settings {
     public Setting<Integer> movementTimeoutTicks = new Setting<>(100);
 
     /**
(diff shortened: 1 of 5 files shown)


SuperOP535:
epic style

PR #283, opened 2018-12-01
allow mapping type to be specified with a jvm arg

babbaj:


leijurv:
jvm arg is probably NoT the way to pick gradle behavior, this should probably be a different gradle task, or a gradle command line arg, or something like that

babbaj:
then you want the same thing but with a command line argument

leijurv:
well actually, i think we'd want a single gradle task that builds 6 jars, api/standalone/unoptimized and notch/searge

SuperOP535:
Closed for inactivity.

PR #293, opened 2018-12-06
Create SETTINGS.md

wilcooo:
A nicer way to see all settings (documentation)

### DIFF
--- a/SETTINGS.md
+++ b/SETTINGS.md
@@ -1,0 +1,535 @@
+### allowBreak
+Allow Baritone to break blocks
+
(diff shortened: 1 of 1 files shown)


leijurv:
Thanks for the PR! However, probably not going to merge this, because this would need to be updated alongside Settings.java every time. Fancy formatted Settings info can be found by running `gradlew javadoc` then opening `build/docs/javadoc/baritone/api/Settings.html`. Perhaps eventually those will be generated automatically on every commit and hosted somewhere.

PR #304, opened 2019-01-07
real gamers use bufferedreader

EvilSourcerer:
scanner is an outdated artifact of the 20th century.

### DIFF
--- a/src/api/java/baritone/api/utils/SettingsUtil.java
+++ b/src/api/java/baritone/api/utils/SettingsUtil.java
@@ -23,9 +23,14 @@ import net.minecraft.item.Item;
 import net.minecraft.util.ResourceLocation;
 
 import java.awt.*;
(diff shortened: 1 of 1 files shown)


leijurv:
So you're saying Scanner is outdated from the 20th century, and we should use BufferedReader instead?

What If I Told You That
```Another difference is Scanner is newer than BufferedReader, only introduced in Java 5, while BufferedReader is present in Java from JDK 1.1 version.```
http://www.java67.com/2016/06/5-difference-between-bufferedreader-and-scanner-in-java.html

EvilSourcerer:
this makes me lose faith in Oracle. They seriously implemented a slower method in the 21st century? Wow... I guess the best I can say is that its an obsolete artifact from the early 21st century.

ZeroMemes:
This PR is life saving, change my mind.

PR #321, opened 2019-01-30
Mega cancer rendering

leijurv:


leijurv:
Still gotta deal with how it doesn't clean up the GL behind it... all mobs render semitransparent lmao

Also the creation of the BSI in baritone is actually garbage XD

PR #337, opened 2019-02-14
Translation of usage.md

ViniDalvino:
The translation isn't 100% done. I would need help for completing it as I am confuse on how to translate end. I would also need help to fix the grammar error.


usereewrjrjeejr:
Ok

PR #410, opened 2019-05-01
Automatic CRLF -> LF normalization

SuperOP535:
LF is clearly superior


ZeroMemes:
Thank you Kanye very cool!

PR #412, opened 2019-05-03
render dank lit boxes for goalYLevel

EvilSourcerer:
Very dank and lit PR 🔥👌👌💪


leijurv:
Mf modified the git ignore for no reason and didn't add a setting javadoc but whatever sure

PR #430, opened 2019-05-27
made some parts way more readble

joerez:
this will make it more human friendly


0-x-2-2:
dont fall for joes lies

PR #438, opened 2019-06-03
Made the bitcoin address stand out as a way to donate. #436

jvyden420:
"clearly a bitcoin address" - 0x


ZeroMemes:
I was going to do this and Leijurv told me not to, intent wasn’t necessarily to make it obvious.

PR #527, opened 2019-07-12
EntityPigZombie check

StijnSimons:
Not finished; maybe add an option for EntityPigZombie?


StijnSimons:
Note; Baritone will just run away from spiders that are already attacking you.

StijnSimons:
Squashing and re-committing

PR #594, opened 2019-07-20
Adding a top command

StijnSimons:
Still work in progress; just wanted to make the PR.


StijnSimons:
Should be ready to go now

0-x-2-2:
this is pozzed

IdanHo:
You could optimise this by going from 255 downwards and break on an air block (instead of doing 255 checks everytime), also why does this commit have "followentity" added to the help page

StijnSimons:
Because it wasnnt added to my previous PR.

0-x-2-2:
this seems flawed. what about the nether roof or a 1x1 or a non solid block lol

Deauthorized:
adding a #surface alias would be cool as well

CDAGaming:
All review comments addressed in #1959, recommended for re-review

dominikaaaa:
I'm closing this in favor of #1959 as they seem to accomplish the same thing but this one is less effective.

Issue #152, opened 2018-09-11
Backdoor


c0nn3r:
Perhaps baritone can have a backdoor so users accounts on T2TB can be stolen? Might make for a fun FitMC reaction.

leijurv:
wow that is a great idea can you make a pull request please

0-x-2-2:
what did he mean by this

Haxified:
yes i support this feature pls make it real, I love fitmc!!1!!!

usereewrjrjeejr:
This is smart

usereewrjrjeejr:
I peed my pants

exwundee:
Can confirm this would be epic.

leijurv:
haxified is salivating rn ^

usereewrjrjeejr:
Hurry the fuck up and do it

LeafHacker:
~~if a backdoor were in baritone this issue would probably be left open as a red herring~~

usereewrjrjeejr:
I'm waiting....

Mhowser:
backdoor when?

usereewrjrjeejr:
https://github.com/cabaletta/baritone/pull/363

exwundee:
We'll get it one day...

usereewrjrjeejr:
Soon

fluffapparition:
Dude no stop it please we aren't having that

leijurv:
oh yes we are 😈

SBVector:
So basically i should not this or impact

leijurv:
@sbvector buy futureclient.net there is no backdoor if that matters so much to you smh

ZeroMemes:
future shill detected?!?!?!?

SBVector:
I mean it should matter i would prefer my account in my hands only.

ZeroMemes:
Future, Baritone and Impact are all safe pieces of software that do not place your account at risk.

fluffapparition:
Nobody is going to use software that gets their minecraft account stolen @leijurv

usereewrjrjeejr:
Wroooong

mrphax0r:
Why is it open? Its already added

0-x-2-2:
I am crying and shaking right now my favorite issue was closed #436

aUniqueUser:
> Nobody is going to use software that gets their minecraft account stolen @leijurv

People Use Aristois

cookiedragon234:
🤔

[...]

PR #677, opened 2019-07-27
improved goals.

joerez:
Security improvements.


StijnSimons:
wtf is this????

StijnSimons:
We’d rather not have a backdoor.

leijurv:
actually this fixes #152

StijnSimons:
Ah yes i see i approve

BryceAS202313:
> actually this fixes #152

This appears to be the only pr that actually fixes #152 rather than just printing to console.

Thunder33345:
How do we know which server they used the command in?
bad PR, need more security to check server ip

Issue #650, opened 2019-07-25
Baritone is unable to make a one block upwards diagonal jump.


BluSpectrum:
## Some information
Operating system: Windows 10
Java version: 8
Minecraft version: 1.14/1.13
Baritone version: 1.14/1.13
Forge mods (if used): None

## Description
Baritone fails to make an upwards diagonal jump with allowPlace off. It can make regular horizontal jumps though. (See picture for more details)
I assume this happens with 1.12 and any other versions available because it seems like a general pathfinding issue.

![image](https://user-images.githubusercontent.com/40967759/61851459-0e946380-ae85-11e9-9cd3-6e892114a2a6.png)

## How to reproduce
1. Reset Baritone settings
2. enable allowParkour, disable allowPlace and allowBreak
3. Use #goto, #click, etc. to set a destination that requires a diagonal upwards jump
4. Baritone will fail to make the jump without placing a block

c0nn3r:
I'll see what I can do @BluSpectrum! 👍

0-x-2-2:
Fixed with #701 being merged 😎

BluSpectrum:
Thank you. I really appreciate this.

PR #701, opened 2019-07-31
Add parkour ascend

c0nn3r:
I made a promise to BluSpectrum in issue #650 that I would do what many (such as @0-x-2-2) thought to be impossible - allow Baritone to do upwards parkour jump to ascend

Since a parkour ascend and a parkour can never be possible in the same cardinal direction from the same origin, I opted to make the existing `MovementParkour dynamicY`. This purely adds new movement options, no previously valid `MovementParkour` is now invalid in favor of an ascend. As a part of this constraint, no ascend parkour place exists. This is because it's ambiguous if a parkour place or an ascend parkour place should be prioritized. For the same reason that parkour place only runs over a gap of 3, parkour place does not ascend.

BluSpectrum, I hope this helps you find happiness. 🤙

### DIFF
--- a/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
@@ -43,18 +43,20 @@ public class MovementParkour extends Movement {
 
     private final EnumFacing direction;
     private final int dist;
+    private final boolean ascend;
 
-    private MovementParkour(IBaritone baritone, BetterBlockPos src, int dist, EnumFacing dir) {
-        super(baritone, src, src.offset(dir, dist), EMPTY, src.offset(dir, dist).down());
+    private MovementParkour(IBaritone baritone, BetterBlockPos src, int dist, EnumFacing dir, boolean ascend) {
+        super(baritone, src, src.offset(dir, dist).up(ascend ? 1 : 0), EMPTY, src.offset(dir, dist).down(ascend ? 0 : 1));
         this.direction = dir;
         this.dist = dist;
+        this.ascend = ascend;
     }
 
     public static MovementParkour cost(CalculationContext context, BetterBlockPos src, EnumFacing direction) {
         MutableMoveResult res = new MutableMoveResult();
         cost(context, src.x, src.y, src.z, direction, res);
         int dist = Math.abs(res.x - src.x) + Math.abs(res.z - src.z);
-        return new MovementParkour(context.getBaritone(), src, dist, direction);
+        return new MovementParkour(context.getBaritone(), src, dist, direction, res.y != src.y);
     }
 
     public static void cost(CalculationContext context, int x, int y, int z, EnumFacing dir, MutableMoveResult res) {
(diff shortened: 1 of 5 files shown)


0-x-2-2:
Wtf I never said this.....

babbaj:
🥇

0-x-2-2:
🥇

c0nn3r:
@0-x-2-2 if you didn’t think it was impossible why didn’t you do it .. 🤔

0-x-2-2:
> @0-x-2-2 if you didn’t think it was impossible why didn’t you do it .. 🤔

im being owned by this mans deadly trolling spree.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -103,19 +105,35 @@ public static void cost(CalculationContext context, int x, int y, int z, EnumFac
...
+            }
+            if (!MovementHelper.fullyPassable(context, destX, y + 2, destZ)) {
+                return;
+            }
+            if (!MovementHelper.fullyPassable(context, destX, y, destZ)) {
+                if (i <= 3 && context.allowParkourAscend && context.canSprint && MovementHelper.canWalkOn(context.bsi, destX, y, destZ) && checkOvershootSafety(context.bsi, destX + xDiff, y + 1, destZ + zDiff)) {
+                    res.x = destX;
+                    res.y = y + 1;
+                    res.z = destZ;
+                    res.cost = costFromJumpDistance(i) + context.jumpPenalty;
```
leijurv: costFromJumpDistance returns the cost of a standard jump. This needs to be changed because for example, when distance is 3, costFromJumpDistance(3) returns `WALK_ONE_BLOCK_COST * 3` while you really want it to be `SPRINT_ONE_BLOCK_COST * 3` since you sprint in the execution of this movement.

This line should instead be `res.cost = i * SPRINT_ONE_BLOCK_COST + context.jumpPenalty;`

0-x-2-2: wow this is a serious bug

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -231,7 +257,7 @@ public MovementState updateState(MovementState state) {
                     // go in the opposite order to check DOWN before all horizontals -- down is preferable because you don't have to look to the side while in midair, which could mess up the trajectory
                     state.setInput(Input.CLICK_RIGHT, true);
                 }
-                if (dist == 3) { // this is a 2 block gap, dest = src + direction * 3
+                if (dist == 3 && !ascend) { // this is a 2 block gap, dest = src + direction * 3
```
leijurv: Why?

c0nn3r: This is akin to a sprintable ascend. When you are doing a non ascend distance 3, you're not sprinting and this code is so that you don't jump too late. While sprinting, making the jump is not an issue so we should jump earlier to land the jump without hitting the edge of the block and losing our forward velocity

leijurv: I see

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -43,18 +43,20 @@
...
         this.dist = dist;
+        this.ascend = ascend;
     }
 
     public static MovementParkour cost(CalculationContext context, BetterBlockPos src, EnumFacing direction) {
         MutableMoveResult res = new MutableMoveResult();
         cost(context, src.x, src.y, src.z, direction, res);
         int dist = Math.abs(res.x - src.x) + Math.abs(res.z - src.z);
-        return new MovementParkour(context.getBaritone(), src, dist, direction);
+        return new MovementParkour(context.getBaritone(), src, dist, direction, res.y != src.y);
```
leijurv: Please make this `res.y > src.y`, if there is ever something like parkour descend this would cause subtle issues.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
leijurv: Can you add a comment explaining this?

c0nn3r:
Thanks for the helpful feedback @leijurv 😄 !

0-x-2-2:
hello i have contacted the higher up members at cabaletta and asked them to merge this. this will be merged within the hour!!!!!!!!!!!!!! thank you for the contribution.

c0nn3r:
[🎉 Yay!](https://www.youtube.com/watch?v=tDXVWDGqvco)

0-x-2-2:
So basically he is contributor

SylveonBottle:
How much would this help Baritone walk across bedrock floors, do you think?

leijurv:
Huh I guess it really would

SylveonBottle:
> Huh I guess it really would

Would it be too much hassle to record a quick snippet of doing that? 
From my experience baritone's always been a bit slow when navigating bedrock floor, falling into holes and jumping back out over and over. I'd like to imagine this new feature would make it a whole lot faster movement-wise, now. Though there isn't a pre-compiled version for me to test it rn

leijurv:
@SylveonBottle https://www.youtube.com/watch?v=bOyS-3FVHIw

KentoNishi:
@c0nn3r you've been featured in [this](https://www.youtube.com/watch?v=LPcPUwL6MjQ) video by FitMC! Congrats 👏

ThatOneJose95:
Sorry if this is the most stupidest thing you have ever heard here but I have never done anything with github compiling code or coding at all and I want to use this in 1.14. What do I need to do to put this "Merge" with the parkour ascend?

PR #707, opened 2019-08-01
use dimensional coordinates (extra parameter for #goal)

IronException:
a parameter option to automaticaly convert dimension coords. eg type overworld coords in nether without calculating it yourself :D. love terbin


0-x-2-2:
made everything worse and removed the copyright settings for intellij for something that should just be like 5 lines...

PR #712, opened 2019-08-02
dimensional coords extension

IronException:
Its an option to add a dimension and it will convert to coords (if you are in overworld and add nether it will multiply the coords by 8 eg). I made it as short as possible.

features:
 -only multiplies x and z values
-works for end, overworld and nether

because it doesnt make -doesnt multiply if ~ is inputed
-you dont have to input it but if you want just add it to the end of the goal or goto command

### DIFF
--- a/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
+++ b/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
@@ -44,10 +44,13 @@ import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.TextComponentString;
 import net.minecraft.util.text.TextFormatting;
 import net.minecraft.util.text.event.ClickEvent;
+import net.minecraft.world.DimensionType;
 import net.minecraft.world.chunk.Chunk;
 
 import java.nio.file.Path;
 import java.util.*;
+
+import static org.apache.commons.lang3.StringUtils.isNumeric;
 
 public class ExampleBaritoneControl implements Helper, AbstractGameEventListener {
     private static final String COMMAND_PREFIX = "#";
@@ -260,7 +263,7 @@ public class ExampleBaritoneControl implements Helper, AbstractGameEventListener
             try {
                 String[] coords = msg.substring("build".length()).trim().split(" ");
                 file = coords[0] + ".schematic";
-                origin = new BlockPos(parseOrDefault(coords[1], ctx.playerFeet().x), parseOrDefault(coords[2], ctx.playerFeet().y), parseOrDefault(coords[3], ctx.playerFeet().z));
+                origin = new BlockPos(parseOrDefault(coords[1], ctx.playerFeet().x, 1), parseOrDefault(coords[2], ctx.playerFeet().y, 1), parseOrDefault(coords[3], ctx.playerFeet().z, 1));
             } catch (Exception ex) {
                 file = msg.substring(5).trim() + ".schematic";
                 origin = ctx.playerFeet();
@@ -672,8 +675,8 @@ public class ExampleBaritoneControl implements Helper, AbstractGameEventListener
         return false;
     }
 
-    private int parseOrDefault(String str, int i) {
-        return str.equals("~") ? i : str.startsWith("~") ? Integer.parseInt(str.substring(1)) + i : Integer.parseInt(str);
+    private int parseOrDefault(String str, int i, double dimensionFactor) {
+        return str.equals("~") ? i : str.startsWith("~") ? (int) (Integer.parseInt(str.substring(1)) * dimensionFactor) + i : (int) (Integer.parseInt(str) * dimensionFactor);
     }
 
     private void log(List<ItemStack> stacks) {
@@ -688,18 +691,23 @@ public class ExampleBaritoneControl implements Helper, AbstractGameEventListener
         Goal goal;
         try {
             BetterBlockPos playerFeet = ctx.playerFeet();
-            switch (params.length) {
+
+            int length = params.length - 1; // length has to be smaller when a dimension parameter is added
+            if (params.length < 1 || (isNumeric(params[params.length - 1]) || params[params.length - 1].startsWith("~"))) {
+                length = params.length;
+            }
+            switch (length) {
                 case 0:
                     goal = new GoalBlock(playerFeet);
                     break;
                 case 1:
-                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y));
+                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y, 1));
                     break;
                 case 2:
-                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x), parseOrDefault(params[1], playerFeet.z));
+                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x, calculateDimensionFactor(params[params.length - 1])), parseOrDefault(params[1], playerFeet.z, calculateDimensionFactor(params[params.length - 1])));
                     break;
                 case 3:
-                    goal = new GoalBlock(new BlockPos(parseOrDefault(params[0], playerFeet.x), parseOrDefault(params[1], playerFeet.y), parseOrDefault(params[2], playerFeet.z)));
+                    goal = new GoalBlock(new BlockPos(parseOrDefault(params[0], playerFeet.x, calculateDimensionFactor(params[params.length - 1])), parseOrDefault(params[1], playerFeet.y, 1), parseOrDefault(params[2], playerFeet.z, calculateDimensionFactor(params[params.length - 1]))));
                     break;
                 default:
                     logDirect("unable to understand lol");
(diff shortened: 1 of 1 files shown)


src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -688,18 +692,20 @@ private Goal parseGoal(String[] params) {
         Goal goal;
         try {
             BetterBlockPos playerFeet = ctx.playerFeet();
-            switch (params.length) {
+
+            // tests wether a dimension input exists because the switch has to be handled differently
+            switch (params.length < 1 || (isNumeric(params[params.length - 1]) || params[params.length - 1].startsWith("~")) ? params.length : params.length - 1) {
```
leijurv: disgusting

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -688,18 +692,20 @@ private Goal parseGoal(String[] params) {
...
                 case 0:
                     goal = new GoalBlock(playerFeet);
                     break;
                 case 1:
-                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y));
+                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y, 1));
                     break;
                 case 2:
-                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x), parseOrDefault(params[1], playerFeet.z));
+                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x, Math.pow(8, ctx.world().provider.getDimensionType().getId() - getDimensionByName(params[params.length - 1]).getId())), parseOrDefault(params[1], playerFeet.z, Math.pow(8, ctx.world().provider.getDimensionType().getId() - getDimensionByName(params[params.length - 1]).getId())));
```
leijurv: move to a helper function

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -711,4 +717,11 @@ private Goal parseGoal(String[] params) {
         }
         return goal;
     }
+
+    public DimensionType getDimensionByName(String nameIn){ String name = nameIn.toLowerCase();
+        if("end".contains(name)) return DimensionType.THE_END;
```
leijurv: does not follow baritone style -- we do not allow `if`s without `{`

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -672,8 +675,9 @@ public boolean isInGoal(int x, int y, int z) {
         return false;
     }
 
-    private int parseOrDefault(String str, int i) {
-        return str.equals("~") ? i : str.startsWith("~") ? Integer.parseInt(str.substring(1)) + i : Integer.parseInt(str);
+    // dimensionFactor is so the coords are converted
+    private int parseOrDefault(String str, int i, double dimensionFactor) {
+        return str.equals("~") ? i : str.startsWith("~") ? Integer.parseInt(str.substring(1)) + i : (int) (Integer.parseInt(str) * dimensionFactor);
```
leijurv: why doesn't the dimension factor apply? if i'm currently in the nether and i do `goal ~100 ~ ~100 overworld` shouldn't it add 100 to my overworld coordinates? i feel like `i` could be `(int) (i * dimensionFactor)`

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
leijurv: split this out into a length before switch instead of a gross ternary.

```
int length;
if (params.length < 1 || (isNumeric(params[params.length - 1]) || params[params.length - 1].startsWith("~")) {
   length = params.length;
}else{
   length  = params.length - 1;
}
switch(length){
```

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: well, when you just do ~ I think you wouldnt want to go somewhere further out then your coordinate. i is the coordinate where you are right now so I think that should stay. If you add 100 now to your overworld coordinates it would be nether coords + 100 / 8 to go there. So the string that is getting parsed should be multiplied in both times. Does that make sense?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: oh alright

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: So I add a function thar calculates how much to multiply to the helper function with DimensionType as parameters.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: sure boss, I also like to have it split up instead of 1 line

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -1,714 +1,734 @@
-/*
```
leijurv: I'm not sure what you did. Perhaps you replaced the LF endings with CRLF? Remove all the `\r` please.  Git thinks you rewrote this entire file.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -711,4 +719,14 @@ private Goal parseGoal(String[] params) {
         }
         return goal;
     }
+
+    public DimensionType getDimensionByName(String nameIn){ String name = nameIn.toLowerCase();
```
leijurv: what

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -711,4 +719,14 @@ private Goal parseGoal(String[] params) {
...
+        if("the_end".contains(name)) { return DimensionType.THE_END; }
```
leijurv: newline after { and before } come on lol

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -711,4 +719,14 @@ private Goal parseGoal(String[] params) {
...
+        if("the_overworld".contains(name) || "surface".contains(name)) { return DimensionType.OVERWORLD; }
+        if("the_nether".contains(name) || "hell".contains(name)) { return DimensionType.NETHER; }
+        return ctx.world().provider.getDimensionType();
+    }
+
+    public double calculateDimensionFactor(DimensionType from, DimensionType to){ return Math.pow(8, from.getId() - to.getId()); }
```
leijurv: just match the existing formatting

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -711,4 +719,14 @@ private Goal parseGoal(String[] params) {
         }
         return goal;
     }
+
+    public DimensionType getDimensionByName(String nameIn){ String name = nameIn.toLowerCase();
+        if("the_end".contains(name)) { return DimensionType.THE_END; }
+        if("the_overworld".contains(name) || "surface".contains(name)) { return DimensionType.OVERWORLD; }
+        if("the_nether".contains(name) || "hell".contains(name)) { return DimensionType.NETHER; }
```
leijurv: space between `if` and `(`

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -688,18 +691,23 @@ private Goal parseGoal(String[] params) {
...
                 case 0:
                     goal = new GoalBlock(playerFeet);
                     break;
                 case 1:
-                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y));
+                    goal = new GoalYLevel(parseOrDefault(params[0], playerFeet.y, 1));
                     break;
                 case 2:
-                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x), parseOrDefault(params[1], playerFeet.z));
+                    goal = new GoalXZ(parseOrDefault(params[0], playerFeet.x, calculateDimensionFactor(ctx.world().provider.getDimensionType(), getDimensionByName(params[params.length - 1]))), parseOrDefault(params[1], playerFeet.z, calculateDimensionFactor(ctx.world().provider.getDimensionType(), getDimensionByName(params[params.length - 1]))));
```
leijurv: if `from` is always `ctx.world().provider.getDimensionType()` then shouldn't `calculateDimensionFactor` just take in the `to`, and it can do `ctx.world().provider.getDimensionType()` within the function?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -688,18 +691,23 @@ private Goal parseGoal(String[] params) {
         Goal goal;
         try {
             BetterBlockPos playerFeet = ctx.playerFeet();
-            switch (params.length) {
+
+            int length = params.length - 1; // length has to be smaller when a dimension parameter is added
```
leijurv: nice! 👍

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
leijurv: dont define an entire function on a single line lol

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: its so you doesnt have to care about capital letters. The new String is because a check wouldnt pass otherwise

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
IronException: yea true. I thought other functions might use the function too but I think all helper functions are just used in that file where they are defined. So I moght as well make them private

leijurv:
Code looks all good!

Issue #721, opened 2019-08-03
#click slight improvements


Bayleaf154:
It would be nice to have the click command as a way to set goals to simplify the cleararea command some what.

leijurv:
????????????????????????????????????????????????????? it already is ??????????????????????????????????? just click and drag ???????????????????????????????????????????????

PR #722, opened 2019-08-03
[Added] Middle mouse click option to select goal

Fefect:
Small improvement from #721 
As an alternative from dragging with left mouse button (incase the position cannot be selected easily in one go).
Goal selection with middlemouse button while in click mode


StijnSimons:
I’m quite sure you can just click a block and it’ll path there.

leijurv:
i'm not seeing the benefit here?

you can left click a single block right now, this just adds an option to middle click to the same effect...?

SIGSTACKFAULT:
Really, there should be a keybind to set the goal to where your're looking _at any time._

jonnyawsom3:
> Really, there should be a keybind to set the goal to where your're looking _at any time._

i presume you mean pathing goal and not digging like they mean

leijurv:
yeah cringe not really useful imo

Issue #474, opened 2019-07-06
A more customizable tunnel feature


0xx06:
We (mostly I) need a more customizable tunnel feature. This includes:

- tunneling in parametrizable dimensions (.b tunnel with*height); 
- diagonal tunneling

I heard that there are logistical problems to extend the current tunnel feature for these options, but maybe it could become a separate module.

Thunder33345:
also let's not miss floor block option, (for example always ensure floor to be obsidian)
see #319 for width X height and also about filling sides(say you dont want it to be enclosed, also an easier way to deal with lava)
see #339 for backfilling
see #480 for tunnel and grab blocks if near
lets say this can be used as a compilation issue related to tunnel

dominikaaaa:
Bumping this because it's pretty old and deserves some more attention. The only part left is to add diagonal support. 

I *might* work on this in the future, don't count on it tho

PossiblePanda:
Bumping! We need this

Issue #319, opened 2019-01-28
Use builder (or rather, breaker) to add a 2x3 highway tunneling mode


leijurv:
Could be useful idk

Could just make a big 2x3x100000 empty schematic lol

but then how to deal with lava hmmMMMM

Thunder33345:
make a custom one?
like to fill above or not
like to fill sides or not
to fill bottom or not

pretty sure something simple like digging side ways to bypass/skip if filling is "disabled"

mikeNG:
1x2 with autofill is another similar useful idea

SylveonBottle:
I recall an old minebot thing where it'd pretty much tunnel for you. It'd place blocks on the bottom to bridge over chasms, and place blocks 'n stuff on either side to stop lava from pouring in. 

Dev would have to put up a system in place to place blocks inside of lava blocks to remove them.

Bendito999:
i personally really like the idea of automatic torch placement too i'm always pausing and unpausing my mining stuff with a hot key so that i can quick place a torch so that teh creepers don't get me, that'd be cool

Haxified:
any updates @leijurv

S-B99:
This is now a thing.

S-B99:
See #971 for torch placement

PR #740, opened 2019-08-05
[New fork, same content] Custom tunneling

Fefect:
Allows for custom tunnelling specified by given parameters.
The player location is the middle of the tunnel if possible (3 = uneven, possible: 4 = even, not possible, otherwise it'll stray to the right one block)
Example: https://dl.dropboxusercontent.com/s/yxya9u6nlyipzcy/2019-08-03_20-42-46.mp4

The current tunneling feature only supports a 2x1 area, this can support and width, height and depth in any of the 4 directions (north,east,south,west).

New PR because my whole fork was scuffed for some reason. (Also less statpadding, just for you BadDev).

Solves #319 and #474

### DIFF
--- a/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
+++ b/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
@@ -38,6 +38,7 @@ import net.minecraft.entity.Entity;
 import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
+import net.minecraft.util.EnumFacing;
 import net.minecraft.util.ReportedException;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.text.ITextComponent;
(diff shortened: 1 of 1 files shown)


src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -368,6 +369,74 @@ public boolean isInGoal(int x, int y, int z) {
             logDirect("tunneling");
             return true;
         }
+        if (msg.startsWith("tunnelcustom")) {
+            String suffix = msg.substring("tunnelcustom".length());
+            int a;
```
StijnSimons: It would def help giving these integers an actual name, not just a b c.

IronException:
Do I get it right that you can basically tell it to dig a tunnel with a specific size?
If so you could also use #cleararea.
I still support a way to do the cleararea to infinity so I like that pr

NotStarRift:
How would I go about adding this feature to my current 1.14.4 version of Baritone?

S-B99:
Closed by #1332 instead.

Issue #580, opened 2019-07-19
Bot can't walk through an open iron door.


Master-64:
## Some information
Operating system: Windows 10 - 64 Bit
Java version: 1.8.0-b132
Minecraft version: 1.12.2
Baritone version: 1.3.3
Forge mods (if used): None

## Exception, error or logs
No logs to report on this.

## How to reproduce
Let's say that we tell the bot to walk 5 blocks in front of him (to his goal) but there is a wall blocking said goal, with an open iron door in the middle of the wall. He can walk around the wall, or he could walk through an iron door that's opened that would save tons of time. So what does this bot do in this scenario? He walks AROUND the whole structure.

## Final checklist
- [ ] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [ ] I have included steps to reproduce the issue.
- [ ] I have not used any OwO's or UwU's in this issue.

leijurv:
thank you @nacgarg very cool

0-x-2-2:
im crying and shaking right now.

MINEBOT WOULD NEVER DO THIS.

Deauthorized:
lol

PR #741, opened 2019-08-05
[Fixed] open iron door not being passable

Fefect:
Fixes the fact that open iron doors cannot be passable.
Solves issue #580 
Proof: https://dl.dropboxusercontent.com/s/9fo3e6y0xmy2dnk/2019-08-05_19-09-40.mp4


leijurv:
Open vs not open isn't good enough, we also need to know from which axis we are approaching a door sadly.

If a door is placed along the X axis, it is passable along the X axis if it's open and not passable if it's closed, but it's passable along the Z axis if it's **closed** and not passable if it's **open**.

This isn't solvable just inside that method because it doesn't have the information of which direction we would be approaching (and potentially passing through) the door in.

PR #752, opened 2019-08-08
Disable lighting on path renderer

HowardStark:
TWO LINES OF CODE


leijurv:
sir, the technology doesn't exist

here is the technology. i'm simply asking you to make it work with optifine shaders.

it's not possible.

MICKEY WAS ABLE TO DO THIS IN A CAVE, WITH TWO LINES OF CODE!

PR #798, opened 2019-08-16
Little optimization

rdmir3:
[Avoids iterators](https://stackoverflow.com/a/33232565), which will likely help [GC overhead](https://stackoverflow.com/questions/256859#comment28239652_256861). The next step is to optimize [list iteration](https://stackoverflow.com/a/257009). Please click the arrow next to "Merge Pull Request" to choose "Squash commits" before merging.

### DIFF
--- a/src/main/java/baritone/cache/WaypointCollection.java
+++ b/src/main/java/baritone/cache/WaypointCollection.java
@@ -56,7 +56,8 @@ public class WaypointCollection implements IWaypointCollection {
     }
 
     private void load() {
-        for (Waypoint.Tag tag : Waypoint.Tag.values()) {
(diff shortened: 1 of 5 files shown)


leijurv:
Show the before and after bytecode. I don't believe that this is even an optimization, and if so, I don't believe that proguard or the JIT aren't already doing it.

0-x-2-2:
📸

ZeroMemes:
A legitimate optimization similar to this was done by me here https://github.com/cabaletta/baritone/commit/a98677dbdacc60bec126786de77d07bac7441bfd. The context in which it is applied here is not applicable.

PR #808, opened 2019-08-18
Stop parseGoal from removing the last argument (when it shouldn't)

### LINKED ISSUE #782: cant get it to use -XYZ
## What do you need help with?
 i cant get it to use a negative coordinates in commands

reb0rt:
This fixes #782. Using #goto -100 -100 for example would send you to y=-100 rather than x=-100 z=-100. 

isNumeric returns false if the input is anything other than a positive integer. This meant that parseGoal wasn't correctly parsing goals that ended in 'impure' numbers (negatives, decimals, etc.). Replacing isNumeric with isCreatable fixes this

### DIFF
--- a/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
+++ b/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
@@ -50,7 +50,7 @@ import net.minecraft.world.chunk.Chunk;
 import java.nio.file.Path;
 import java.util.*;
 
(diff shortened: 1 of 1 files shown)


rebort201:
For context, this means typing #goal -10 -10 would normally set the goal at y=-10 rather than x=-10 z=-10

leijurv:
oh shit thats a better solution than the garbage i pushed

leijurv:
i'll force push undo my last commit then merge this LOL

leijurv:
thanks!

PR #845, opened 2019-08-24
multi-line processing and distance limit arguments to farm

millennIumAMbiguity:
*added multi-line/multi-commands processing.
*added a distance limit arguments to farm.

Signed-off-by: millennIumAMbiguity <37588844+millennIumAMbiguity@users.noreply.github.com>

demo: `#goto -314 -113 #mine dirt 49 #farm 7`
demo

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -14,6 +14,10 @@ To disable direct chat control (with no prefix), turn off the `chatControl` sett
 
 `help` for (rudimentary) help. You can see what it says [here](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/utils/ExampleBaritoneControl.java#L47).
 
+It is possible to automatically execute commands after one another. example: `#mine oak_log 64 #mine stone 64 #goto x y z`. this try to harvest 64 oak logs and 64 cobblestone/stone and then head to "x y z". perfect for harvesting materials for your house while you go and eat a snack.
+not all commands will automatically execute the next one, for example: `thisway` and `goal`. if you ever get stuck or want to do the next operation, use `next`.
+all upcoming commands will be canceled when typing any command except for `next` and `add`.
+
 To toggle a boolean setting, just say its name in chat (for example saying `allowBreak` toggles whether Baritone will consider breaking blocks). For a numeric setting, say its name then the new value (like `primaryTimeoutMS 250`). It's case insensitive. To reset a setting to its default value, say `acceptableThrowawayItems reset`. To reset all settings, say `reset`. To see all settings that have been modified from their default values, say `modified`.
 
 Some common examples:
@@ -31,11 +35,13 @@ Some common examples:
 - `build` to build a schematic. `build blah` will load `schematics/blah.schematic` and build it with the origin being your player feet. `build blah x y z` to set the origin. Any of those can be relative to your player (`~ 69 ~-420` would build at x=player x, y=69, z=player z-420).
 - `schematica` to build the schematic that is currently open in schematica
 - `tunnel` to dig just straight ahead and make a tunnel
-- `farm` to automatically harvest, replant, or bone meal crops
+- `farm range x y z` or `farm range x z` or `farm range y` or `farm range` or `farm` to automatically harvest, replant, or bone meal crops inside of its range. if range is 0 then the range is limitless. default is limitless from the player's feats position.
 - `axis` to go to an axis or diagonal axis at y=120 (`axisHeight` is a configurable setting, defaults to 120).
 - `explore x z` to explore the world from the origin of x,z. Leave out x and z to default to player feet. This will continually path towards the closest chunk to the origin that it's never seen before. `explorefilter filter.json` with optional invert can be used to load in a list of chunks to load.
 - `invert` to invert the current goal and path. This gets as far away from it as possible, instead of as close as possible. For example, do `goal` then `invert` to run as far as possible from where you're standing at the start.
 - `version` to get the version of Baritone you're running
+- `next` to skip to the next command in the command queue.
+- `add` add one ore more comm,ands to the command queue. example: `#add #farm 16`.
 - `damn` daniel
 
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -14,9 +14,9 @@ To disable direct chat control (with no prefix), turn off the `chatControl` sett
 
 `help` for (rudimentary) help. You can see what it says [here](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/utils/ExampleBaritoneControl.java#L47).
 
-It is possible to automatically execute commands after one another. example: `#mine oak_log 64 #mine stone 64 #goto x y z`. this try to harvest 64 oak logs and 64 cobblestone/stone and then head to "x y z". perfect for harvesting materials for your house while you go and eat a snack.
-not all commands will automatically execute the next one, for example: `thisway` and `goal`. if you ever get stuck or want to do the next operation, use `next`.
-all upcoming commands will be canceled when typing any command except for `next` and `add`.
+It is possible to automatically execute commands after one another. Example: `#mine oak_log 64 #mine stone 64 #goto x y z`. This tries to harvest 64 oak logs and 64 cobblestone/stone and then head to "x y z". Perfect for harvesting materials for your house while you go and eat a snack.
+Not all commands will automatically execute the next one, for example: `thisway` and `goal`. If you ever get stuck or want to do the next operation, use `next`.
+All upcoming commands will be canceled when typing any command except for `next` and `add`.
 
 To toggle a boolean setting, just say its name in chat (for example saying `allowBreak` toggles whether Baritone will consider breaking blocks). For a numeric setting, say its name then the new value (like `primaryTimeoutMS 250`). It's case insensitive. To reset a setting to its default value, say `acceptableThrowawayItems reset`. To reset all settings, say `reset`. To see all settings that have been modified from their default values, say `modified`.
 
@@ -35,13 +35,14 @@ Some common examples:
 - `build` to build a schematic. `build blah` will load `schematics/blah.schematic` and build it with the origin being your player feet. `build blah x y z` to set the origin. Any of those can be relative to your player (`~ 69 ~-420` would build at x=player x, y=69, z=player z-420).
 - `schematica` to build the schematic that is currently open in schematica
 - `tunnel` to dig just straight ahead and make a tunnel
-- `farm range x y z` or `farm range x z` or `farm range y` or `farm range` or `farm` to automatically harvest, replant, or bone meal crops inside of its range. if range is 0 then the range is limitless. default is limitless from the player's feats position.
+- `farm range x y z` or `farm range x z` or `farm range y` or `farm range` or `farm` to automatically harvest, replant, or bone meal crops inside of its range. If range is 0 then the range is limitless. Default is limitless from the player's feats position.
 - `axis` to go to an axis or diagonal axis at y=120 (`axisHeight` is a configurable setting, defaults to 120).
 - `explore x z` to explore the world from the origin of x,z. Leave out x and z to default to player feet. This will continually path towards the closest chunk to the origin that it's never seen before. `explorefilter filter.json` with optional invert can be used to load in a list of chunks to load.
 - `invert` to invert the current goal and path. This gets as far away from it as possible, instead of as close as possible. For example, do `goal` then `invert` to run as far as possible from where you're standing at the start.
 - `version` to get the version of Baritone you're running
+- `commandlist` to show all upcoming commands.
 - `next` to skip to the next command in the command queue.
-- `add` add one ore more comm,ands to the command queue. example: `#add #farm 16`.
+- `add` add one ore more commands to the command queue. Example: `#add #farm 16`.
 - `damn` daniel
 
 
--- a/src/api/java/baritone/api/process/IFarmProcess.java
+++ b/src/api/java/baritone/api/process/IFarmProcess.java
@@ -18,5 +18,9 @@
 package baritone.api.process;
 
 public interface IFarmProcess extends IBaritoneProcess {
-    void farm();
+    void farm(int range, int posX, int posY, int posZ);
+
+    default void farm(){
+        farm(0, 0, 0, 0);
+    }
 }
--- a/src/api/java/baritone/api/process/IFarmProcess.java
+++ b/src/api/java/baritone/api/process/IFarmProcess.java
@@ -17,10 +17,8 @@
 
 package baritone.api.process;
 
-public interface IFarmProcess extends IBaritoneProcess {
-    void farm(int range, int posX, int posY, int posZ);
+import net.minecraft.util.math.BlockPos;
 
-    default void farm(){
-        farm(0, 0, 0, 0);
-    }
+public interface IFarmProcess extends IBaritoneProcess {
+    void farm(int range, BlockPos pos);
 }
--- a/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
+++ b/src/api/java/baritone/api/utils/ExampleBaritoneControl.java
@@ -59,6 +59,8 @@ public class ExampleBaritoneControl implements Helper, AbstractGameEventListener
     public final IPlayerContext ctx;
 
     public ExampleBaritoneControl(IBaritone baritone) {
+        if (thisExampleBaritoneControl == null)
+            thisExampleBaritoneControl = this;
         this.baritone = baritone;
         this.ctx = baritone.getPlayerContext();
         baritone.getGameEventHandler().registerEventListener(this);
@@ -68,7 +70,15 @@ public class ExampleBaritoneControl implements Helper, AbstractGameEventListener
     public void onSendChatMessage(ChatEvent event) {
         String msg = event.getMessage();
         if (BaritoneAPI.getSettings().prefixControl.value && msg.startsWith(COMMAND_PREFIX)) {
-            if (!runCommand(msg.substring(COMMAND_PREFIX.length()))) {
+
+            String[] commandsList = msg.split(COMMAND_PREFIX);
+            commandsToBeExecutedOld = commandsToBeExecuted;
+            commandsToBeExecuted = "";
+            for (int i = 2; i < commandsList.length; i++) {
+                commandsToBeExecuted += COMMAND_PREFIX+ commandsList[i];
+            }
+
+            if (!runCommand(commandsList[1])) {
                 logDirect("Invalid command");
             }
             event.cancel(); // always cancel if using prefixControl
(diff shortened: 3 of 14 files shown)


src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -59,6 +59,8 @@
     public final IPlayerContext ctx;
 
     public ExampleBaritoneControl(IBaritone baritone) {
+        if (thisExampleBaritoneControl == null)
```
leijurv: no, we d oNot do this, each baritone instance is 100% segmented from the other. no static variables **anywhere**. make a field in Baritone

src/api/java/baritone/api/process/IFarmProcess.java
```diff
@@ -18,5 +18,9 @@
 package baritone.api.process;
 
 public interface IFarmProcess extends IBaritoneProcess {
-    void farm();
+    void farm(int range, int posX, int posY, int posZ);
```
leijurv: use a blockpos. or even better an Optional<BlockPos>

USAGE.md
```diff
@@ -31,11 +35,13 @@ Some common examples:
...
-- `farm` to automatically harvest, replant, or bone meal crops
+- `farm range x y z` or `farm range x z` or `farm range y` or `farm range` or `farm` to automatically harvest, replant, or bone meal crops inside of its range. if range is 0 then the range is limitless. default is limitless from the player's feats position.
 - `axis` to go to an axis or diagonal axis at y=120 (`axisHeight` is a configurable setting, defaults to 120).
 - `explore x z` to explore the world from the origin of x,z. Leave out x and z to default to player feet. This will continually path towards the closest chunk to the origin that it's never seen before. `explorefilter filter.json` with optional invert can be used to load in a list of chunks to load.
 - `invert` to invert the current goal and path. This gets as far away from it as possible, instead of as close as possible. For example, do `goal` then `invert` to run as far as possible from where you're standing at the start.
 - `version` to get the version of Baritone you're running
+- `next` to skip to the next command in the command queue.
+- `add` add one ore more comm,ands to the command queue. example: `#add #farm 16`.
```
leijurv: comm,ands

USAGE.md
```diff
@@ -14,6 +14,10 @@ To disable direct chat control (with no prefix), turn off the `chatControl` sett
 
 `help` for (rudimentary) help. You can see what it says [here](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/utils/ExampleBaritoneControl.java#L47).
 
+It is possible to automatically execute commands after one another. example: `#mine oak_log 64 #mine stone 64 #goto x y z`. this try to harvest 64 oak logs and 64 cobblestone/stone and then head to "x y z". perfect for harvesting materials for your house while you go and eat a snack.
+not all commands will automatically execute the next one, for example: `thisway` and `goal`. if you ever get stuck or want to do the next operation, use `next`.
+all upcoming commands will be canceled when typing any command except for `next` and `add`.
```
leijurv: capitalize

USAGE.md
```diff
@@ -14,6 +14,10 @@ To disable direct chat control (with no prefix), turn off the `chatControl` sett
 
 `help` for (rudimentary) help. You can see what it says [here](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/utils/ExampleBaritoneControl.java#L47).
 
+It is possible to automatically execute commands after one another. example: `#mine oak_log 64 #mine stone 64 #goto x y z`. this try to harvest 64 oak logs and 64 cobblestone/stone and then head to "x y z". perfect for harvesting materials for your house while you go and eat a snack.
```
leijurv: grammar and capitalization

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -365,8 +375,52 @@ public boolean isInGoal(int x, int y, int z) {
...
-            baritone.getFarmProcess().farm();
+        if (msg.startsWith("farm")) {
+            String rest = msg.substring(4).trim();
+            String[] params = rest.split(" ");
+            int[] paramsInt = new int[params.length];
+            try {
+                for (int i = 0; i < paramsInt.length; i++) {
+                    paramsInt[i] = Integer.parseInt(params[i]);
+                }
+            } catch (NumberFormatException | ArrayIndexOutOfBoundsException | NullPointerException ex) {
```
leijurv: how could arrayindexoutofbounsdexception ever be caused...? or nullpointerexception?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -365,8 +375,52 @@ public boolean isInGoal(int x, int y, int z) {
...
         }
-        if (msg.equals("farm")) {
-            baritone.getFarmProcess().farm();
+        if (msg.startsWith("farm")) {
+            String rest = msg.substring(4).trim();
+            String[] params = rest.split(" ");
+            int[] paramsInt = new int[params.length];
+            try {
+                for (int i = 0; i < paramsInt.length; i++) {
+                    paramsInt[i] = Integer.parseInt(params[i]);
```
leijurv: use the existing parseordefault meme that allows for ~ syntax

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -365,8 +375,52 @@ public boolean isInGoal(int x, int y, int z) {
...
+                case 3:
+                    args[1] = paramsInt[1];
+                    args[2] = ctx.playerFeet().y;
+                    args[3] = paramsInt[2];
+                    break;
+                case 4:
+                    args = paramsInt;
+                    break;
+                default:
+                    logDirect("to many arguments");
```
leijurv: ?????

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -365,8 +375,52 @@ public boolean isInGoal(int x, int y, int z) {
...
+                    return true;
+            }
+            baritone.getFarmProcess().farm(args[0], args[1], args[2], args[3]);
```
leijurv: cring

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +716,47 @@ public boolean isInGoal(int x, int y, int z) {
         }
         if (msg.equals("damn")) {
             logDirect("daniel");
+            return true;
```
leijurv: no

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +716,47 @@ public boolean isInGoal(int x, int y, int z) {
...
+        }
+        if (msg.startsWith("add")){
+            String newCommands = msg.substring(3).trim();
+            commandsToBeExecuted = commandsToBeExecutedOld+ newCommands;
```
leijurv: spacing?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +716,47 @@ public boolean isInGoal(int x, int y, int z) {
...
+            return true;
+        }
+        if (msg.equals("next")){
+            commandsToBeExecuted = commandsToBeExecutedOld;
```
leijurv: what is even going on with commandstobeexecutedold

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +716,47 @@ public boolean isInGoal(int x, int y, int z) {
...
+            logDirect(commandsToBeExecuted);
+            return true;
         }
         return false;
     }
 
+    private static ExampleBaritoneControl thisExampleBaritoneControl;
+    private static String commandsToBeExecuted = "";
+    private static String commandsToBeExecutedOld = "";
+    public static void nextCommand(){ // execute nex command in commandsToBeExecuted
```
leijurv: ?????????

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +716,47 @@ public boolean isInGoal(int x, int y, int z) {
...
+        }
+        if (msg.equals("commandlist")){
+            commandsToBeExecuted = commandsToBeExecutedOld;
+            logDirect(commandsToBeExecuted);
+            return true;
         }
         return false;
     }
 
+    private static ExampleBaritoneControl thisExampleBaritoneControl;
```
leijurv: fields go at the top. this field needs to be killed and the other two need to be non static

src/main/java/baritone/process/BackfillProcess.java
```diff
@@ -124,6 +124,7 @@ public void onLostControl() {
         if (blocksToReplace != null && !blocksToReplace.isEmpty()) {
             blocksToReplace.clear();
         }
+        super.onLostControl();
```
leijurv: this is a reasonable way to do this. have you checked if cancel will cancel one or all?

src/main/java/baritone/process/FarmProcess.java
```diff
@@ -96,7 +99,9 @@ public boolean isActive() {
     }
 
     @Override
-    public void farm() {
+    public void farm(int range, int posX, int posY, int posZ) {
+        center = new BlockPos(posX, posY, posZ);
```
leijurv: see you immediately turn it into a blockpos, why not pass it as a blockpos?

src/main/java/baritone/process/FarmProcess.java
```diff
@@ -181,6 +186,12 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
         List<BlockPos> bonemealable = new ArrayList<>();
         List<BlockPos> openSoulsand = new ArrayList<>();
         for (BlockPos pos : locations) {
+
+            //check if outside of range
+            if (range != 0 && pos.getDistance(center.getX(), center.getY(), center.getZ()) > range) {
```
leijurv: 👍

src/main/java/baritone/process/MineProcess.java
```diff
@@ -85,7 +85,7 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
             if (curr >= desiredQuantity) {
                 logDirect("Have " + curr + " " + item.getItemStackDisplayName(new ItemStack(item, 1)));
                 cancel();
-                return null;
+                return new PathingCommand(null, PathingCommandType.REQUEST_PAUSE);
```
leijurv: why?

src/main/java/baritone/utils/BaritoneProcessHelper.java
```diff
@@ -33,6 +34,11 @@ public BaritoneProcessHelper(Baritone baritone) {
         baritone.getPathingControlManager().registerProcess(this);
     }
 
+    @Override
+    public void onLostControl(){
+        ExampleBaritoneControl.nextCommand();
```
leijurv: you have a reference to the Baritone here, so it shouldn't be an issue to make examplebaritonecontrol a field

however, i don't think that's reeeeeeeally the way to go, since examplebaritonecontrol is supposed to be an example of how to use the baritone api

tell you what, add a new event to the gameeventlistener like onProcessCompleted(IBaritoneProcess completed), and then this will call `baritone.getGameEventListener.onProcessCompleted(this)`, and examplebaritonecontrol's constructor adds itself as a gameeventlistener and listens for that event,and uses that to advance the commands.

so scratch everything i said above about making examplebaritonecontrol a field of baritone,don't do that, do this instead lol

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: when you enter any commands ti will clear the upcoming commands unless the command entered is next or add. commandsToBeExecutedOld is used to restore the commandsToBeExecuted

src/main/java/baritone/process/MineProcess.java
millennIumAMbiguity: because else it trows and crashes the game. this was annoying because I thaught the crashing was my falt but turns out that MineProcess.java have that one difference compared to all other processes.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: why?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: yes?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: "no, we do not do this,"

src/main/java/baritone/process/MineProcess.java
leijurv: i mean the implied question is why didn't this used to crash, and why does it only crash with your other changes applied

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
leijurv: intended easter egg. typing `damn` in chat should still send `damn` it just has baritone also do `[Baritone] daniel`

src/main/java/baritone/process/MineProcess.java
millennIumAMbiguity: it crashes when its called twice in one frame.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: I'm not going to test that right now but when I tested it worked like this:
`#daim`
`[Baritone] daniel`
`[Baritone] Invalid command`

or something like that

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: I'm not sure about how you want me to do this.

leijurv: Sorry, I changed my mind about this, see the last comment. I also said `read the last comment first since i changed my mind like half way through lol`. Make this class listen for a ProcessCompletionEvent

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -58,7 +58,13 @@
     public final IBaritone baritone;
     public final IPlayerContext ctx;
 
+    private static ExampleBaritoneControl thisExampleBaritoneControl;
```
0-x-2-2: :elmo:

0-x-2-2: :elmo:

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +719,43 @@ public boolean isInGoal(int x, int y, int z) {
         }
         if (msg.equals("damn")) {
             logDirect("daniel");
+            return true;
```
0-x-2-2: why did you mark this as resolved it was not even removed 🤔

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
0-x-2-2: bad

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -365,8 +379,51 @@ public boolean isInGoal(int x, int y, int z) {
...
-            baritone.getFarmProcess().farm();
+        if (msg.startsWith("farm")) {
+            String rest = msg.substring(4).trim();
+            String[] params = rest.split(" ");
+            int[] args = new int[]{-1, 0, 0, 0};
+            if (params.length > 0) {
+                args[0] = Integer.parseInt(params[0]);
+            }
+            try {
+                switch (params.length) {
```
0-x-2-2: this is horrifying.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
0-x-2-2: this is intended

src/api/java/baritone/api/utils/Helper.java
```diff
@@ -69,4 +69,9 @@ default void logDirect(String message) {
         component.appendSibling(new TextComponentString(" " + message));
         Minecraft.getMinecraft().addScheduledTask(() -> BaritoneAPI.getSettings().logger.value.accept(component));
     }
+    default void logDirectTextComponentString(TextComponentString message) {
```
0-x-2-2: is this really required...

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
```diff
@@ -662,10 +719,43 @@ public boolean isInGoal(int x, int y, int z) {
...
+        String[] commandsList = commandsToBeExecuted.split(COMMAND_PREFIX);
+        if (commandsList.length > 1) {
+            String msg0 = commandsList[1];
+            StringBuilder sb = new StringBuilder();
+            for (int i = 2; i < commandsList.length; i++) {
+                sb.append(COMMAND_PREFIX);
+                sb.append(commandsList[i]);
+            }
+            commandsToBeExecuted = sb.toString();
+            thisExampleBaritoneControl.logDirectTextComponentString(new TextComponentString(String.format("%s Executing: %s" + COMMAND_PREFIX + msg0, TextFormatting.YELLOW, TextFormatting.WHITE)));
```
0-x-2-2: does not look like `logDirectTextComponentString` is required at all lol. remove `logDirectTextComponentString` and replace it with `logDirect` it should work perfectly fine.

src/main/java/baritone/utils/BaritoneProcessHelper.java
0-x-2-2: > you have a reference to the Baritone here, so it shouldn't be an issue to make examplebaritonecontrol a field
> 
> however, i don't think that's reeeeeeeally the way to go, since examplebaritonecontrol is supposed to be an example of how to use the baritone api
> 
> tell you what, add a new event to the gameeventlistener like onProcessCompleted(IBaritoneProcess completed), and then this will call `baritone.getGameEventListener.onProcessCompleted(this)`, and examplebaritonecontrol's constructor adds itself as a gameeventlistener and listens for that event,and uses that to advance the commands.
> 
> so scratch everything i said above about making examplebaritonecontrol a field of baritone,don't do that, do this instead lol

👍

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: Without `return true;` or with `return false;` the command will return: 
`[Baritone] daniel`
`[Baritone] Invalid command`

With `return true;` the command will return: 
`[Baritone] daniel`

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: Should I have `args[0] = Integer.parseInt(params[0]);` in every case from 1 to 4 instead?

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
leijurv: For the last time, this is **intended behavior**, don't add this `return true;`. This will not be merged until the `return true;` is removed.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
leijurv: no you're just really overthinking it and this is wild complicated for no reason

imo there should be a farm range setting. to change the farm range, change that setting.

the farm center should be your playerfeet, or you pass x y z as args. farm should take 0 arguments or 3.

millennIumAMbiguity: that's stupid. while using this for myself I never had any use for setting the coordinates. But I constantly need a different range of settings. 

I just realized that one problem with the coordinates function is very easy to fix. If you are too far away from the target center, it will not work. to fix this is to simply use the `goto`.
so the command `#farm 10 x y z` will be translated to `#goto x y z #farm 10`.

leijurv: truly horrifying

leijurv: then make those separate settings. this is too much complexity in farm lol.

goto already has the logic for parsing this.

i'm fine if you want to do `farm radius` = farm at that radius and `farm` = farm with no radius. in that case, remove the farm radius setting.

src/api/java/baritone/api/utils/Helper.java
millennIumAMbiguity: I use it to have different and multiple colors.
I use yellow in logic messages. see the demo.

and no, the default Minecraft color system adds a character.

src/api/java/baritone/api/utils/ExampleBaritoneControl.java
millennIumAMbiguity: soooo...
`farm` that works just like the original and
`farmrange` that have a range setting.

What we/I can do... is:
`farmrange i` farm within the range of `i` from player feet.
`farmrange  x1 y1 z1 x2 y2 z2` or `farmrange  x1 z1 x2 z2` to specify an area.
`farmrange i shape` farm within the range of `i` from player feet with a specific shape like a rectangle.
`farmthis` auto-set aria of the current position and farm it.

leijurv:
super cringe

feel free to remake the "distance limit arguments to farm" as its own pr tho

PR #867, opened 2019-08-28
Add replantCrops setting

Orinion:
add replantCrops setting, simmilar to replantNetherWart for the other crops.
Usefull to disable when f.E. raiding villiges

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -662,7 +662,7 @@ public final class Settings {
     public final Setting<Boolean> replantCrops = new Setting<>(true);
 
     /**
-     * Replant nether wart while farming
+     * Replant nether wart while farming. This setting only has an effect when replantCrops is also enabled
      */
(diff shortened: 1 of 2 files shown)


leijurv:
Nice! Can you make two small changes:

We use `{ }` around all blocks in Baritone even if they're only one line, can you add curlies around your two `if`s

When replantCrops is false, this should make it not replant nether wart.  (i.e. nether wart should only be replanted if both replantCrops is true, and replantNetherWart is true)

Orinion:
sure!
And should i update the docs etc. too?

leijurv:
Yes actually, add something to replantNetherWart's documentation saying like `This setting only has an effect when replantCrops is also enabled`

leijurv:
Can you remove the unrelated change to waypoints? (separate pr)

Also for 1.13 and 1.14 we merge in from master, instead of cherry picking separate PRs to each. Thank you though, but only the one to 1.12 will be merged, then we'll merge master overall into 1.13, then that into 1.14.

0-x-2-2:
cringe pr bruh

Issue #869, opened 2019-08-28
Cannot "goto" saved locations


hexxone:
## Some information
Operating system:
Java version: 1.8.0_221
Minecraft version: 1.14.4
Baritone version: (the one included in Impact 4.8.0-20190827-nightly)
Forge mods (if used): none

## Exception, error or logs
`§5[§dBaritone§5]§7 Saved user defined position BlockPos{x=-1229, y=90, z=-303} under name 'home'. Say 'goto home' to set goal, say 'list user' to list custom waypoints.`

`§5[§dBaritone§5]§7 None saved for tag HOME`

## How to reproduce
1. use #save home
2. use #goto home

Orinion:
Its a conflict with the home tag. The tags are 'home', 'user', 'death' and 'bed', the waypoint gets saved in 'user' but when using goto baritone looks in the tag 'home'. Same happens if you do "save bed" etc.

Hexxonite:
Alright, thanks for clarification and fix :)

PR #879, opened 2019-08-30
Check if waypoint name is Tag

Orinion:
before, saving a waypint with a tag name would not result in an error, but you would not be able to run "goto" on it. see #869


leijurv:
Looks good! Can you put the `{` on the same line as the `for`

PR #890, opened 2019-09-01
Coordinate censoring (fixes #872)

### LINKED ISSUE #872: Add a setting to disable coords in client side of the chat
## Describe your suggestion
For example when typing in coords, in the chat you see the coords you put in. For recording purposes, its easy to block whats being directly typed. However, after typing the coords, baritone tells you exactly where its going, which there should be a setting to disable that so it doesn't stack up in the chat history, so when it does making it extremely hard to cover the coords. If there is already feature like this please inform me. Otherwise please add this in.

## Settings
Baritone put coords in chat, True to see where its leading you, False if baritone doesn't put the coords in the chat.

## Context
To make censoring coords more easier for recording on servers that leaking coordinates may be problematic. This way we don't have to blur out half the screen every time we open the chat or go to sleep.

## Final checklist

LoganDark:
#872


LoganDark:
Censors all types of goals:

image

BetterBlockPos censors its own coordinates when this is enabled as well, and I have converted any instance of logging a BlockPos to chat with BetterBlockPos so that it can be censored.

leijurv:
Looks good, function name is a bit unwieldy long though, can you rename to `maybeCensor`?

Zamazenta:
<message deleted>

Zamazenta:
<m>

LoganDark:
@Zamazenta hey, mind adding a comment in your comment?

image

Zamazenta:
@LoganDark sorry i can't do that

LoganDark:
possiblyCensorASingleCoordinateThatMayBeAPartOfALongerOneButIsPassedToThisMethodIndividuallyDependingOnWhetherOrNotTheSettingToCensorCoordinatesIsEnabledAndIfItIsReturnACensoredStringOrElseReturnTheArgumentThatWasPassedToTheMethodAsTheFirstIntegerArgument(int a)

Zamazenta:
>

PR #894, opened 2019-09-01
Setting to prefer silk touch, fixes #883

### LINKED ISSUE #883: Disable autotool / silk touch glass blocks
I'd like a feature to disable the autotool function when mining /clearing an area. Mainly i'd need to get the glass blocks with a silk touch pickaxe when clearing /mining. This could be done with a simple setting like #autotool true/false or even something customizable.

LoganDark:
For all tools with equal speed (i.e. pickaxe with silk touch vs. your hand on glowstone), prefers silk touch tools over everything else. The fastest tool will always be chosen, however when choosing between tools of equal speed, Baritone will prefer silk touch ones.


LoganDark:
Oh wait, I think I introduced a small logic error when editing the code after testing, give me a bit

LoganDark:
Tested and ready to merge

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -90,30 +90,39 @@ private int getMaterialCost(ItemStack itemStack) {
...
-                materialCost = getMaterialCost(itemStack);
-            } else if (v == value) {
-                int c = getMaterialCost(itemStack);
-                if (c < materialCost) {
-                    value = v;
+                lowestCost = getMaterialCost(itemStack);
+                bestSilkTouch = silkTouch;
+            } else if (speed == highestSpeed) {
+                int cost = getMaterialCost(itemStack);
+                if (cost < lowestCost && (silkTouch || !bestSilkTouch) ||
```
leijurv: Can you add one extra parentheses around the first clause of this plz? While you and I may realize immediately that `a && b || c` becomes `(a && b) || c` it's not always apparent to readers =)

PR #909, opened 2019-09-04
Mining go Home after Mining and check inventory while Mining

derBobo:
Edited the MineProcess to have the ability to check the inventory for free space while mining , and also added the feature to go home, when the minigProcess gets canceled.


derBobo:
adding changes to master

PR #926, opened 2019-09-11
1.14.4 Switch to Litematica

FTC55:
This pull request replaces the old schematica/schematic system in favor of the new 1.13+ compatible litematica/litematic one. ~~Everything related to the old system has been disabled but left in the source in the event of future fixes that will allow schematics to properly work again.~~
EDIT: As indicated by leijurv, everything from schematica / the old system has been torn out.


FTC55:
Travis CI fails due to javadoc which is totally unrelated to my changes. What should I do?

FTC55:
On top of that, I used some code from https://github.com/maruohon/litematica/tree/fabric_1.14.4 but I'm not familiar with licenses and crediting so It would be nice if someone competent could help in this regard.

derBobo:
Travis doesn`t fail beacuase of the javadocs, it failse because of the java.lang.IllegalStateException at line 5828 of the Travis log

leijurv:
Fantastic, thank you so much!

I have a few questions: why against 1.14.4 not 1.13.2? Since it is needed there too. We merge from the lowest version upward, so changes are generally made to master (1.12.2) then merged into 1.13.2 then 1.14.4. Since this is for 1.13.2+ I think it's better if it's pulled into 1.13.2.

All the old schematic loading code and all the schematica stuff should be torn out. This has already happened in 1.14.4 but if you're doing 1.13.2 you may have to replicate it I guess? (is easy though) 

For the license stuff, I believe the best way is to put everything that's pasted into its own package with only that code, then put in a LICENSE file into just that folder that's copied from wherever it's from. What license is it specifically? We can pull in MIT, apache, or lgpl into baritone without issue. 

Don't worry about Travis or codacy on 1.14.4 btw.

FTC55:
@leijurv I am delighted i could help this much. I don't think there would be any problems replicating this in 1.13.2 however there are some files from litematica that i've considerably shortened since most of the methods weren't needed. Is there anything different with licensing when the files aren't exact copies but instead modifications? Also, the repository the code came from is licensed as GPL-3.0

leijurv:
Damn. That's a real shame. That means we can't legally paste it into baritone, gpl-3.0 is literally cancer. I can't merge this pr without violating either/both of those licenses. You'd have to reimplement the same logic using different code, or implement it not from code but from their documentation, like what I did for .schematic. =/

FTC55:
This sucks. I was starting to get worried too when i read "Same License" in the github entry for GPL-3.0 but i thought it could be somehow worked around. At this point I'm not entirely sure I'm skilled enough to rework the entire code but most importantly I wouldn't know how to go about actually reimplementing without risking to produce something too similar to what i'm currently using.

FTC55:
@leijurv I started trying to rewrite some code but I would like to know if I should take everything down until everything is rewritten to avoid repercussions due to my fork being public and possibly violating the license.

leijurv:
I don't think you have anything to worry about in that regard, I believe the issue would only come in once you tried to distribute it compiled or something like that. No need to take it down I don't think

src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStateContainer.java
```diff
@@ -0,0 +1,129 @@
+package baritone.utils.schematic.litematica;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.block.Blocks;
+import net.minecraft.nbt.ListNBT;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Vec3i;
+
+public class LitematicaBlockStateContainer implements ILitematicaBlockStatePaletteResizer
```
ZeroMemes: This is also taken from the decompiled Minecraft game source. Once again, the game class `net.minecraft.world.chunk.BlockStateContainer` is not final, so you should extend it and then implement the additional functionality desired.

src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStatePaletteHashMap.java
```diff
@@ -0,0 +1,101 @@
+package baritone.utils.schematic.litematica;
+
+import javax.annotation.Nullable;
+import net.minecraft.block.BlockState;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.nbt.ListNBT;
+import net.minecraft.util.ObjectIntIdentityMap;
+import net.minecraft.nbt.NBTUtil;
+
+public class LitematicaBlockStatePaletteHashMap implements ILitematicaBlockStatePalette
```
ZeroMemes: aNoThEr class taken from the decompiled Minecraft game source. Same course of action should be taken as the others.

src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStatePaletteLinear.java
```diff
@@ -0,0 +1,112 @@
+package baritone.utils.schematic.litematica;
+
+import javax.annotation.Nullable;
+import net.minecraft.block.BlockState;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.nbt.ListNBT;
+import net.minecraft.nbt.NBTUtil;
+
+public class LitematicaBlockStatePaletteLinear implements ILitematicaBlockStatePalette
```
ZeroMemes: Same shit

src/main/java/baritone/utils/schematic/litematica/ILitematicaBlockStatePalette.java
```diff
@@ -0,0 +1,27 @@
+package baritone.utils.schematic.litematica;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.nbt.ListNBT;
+
+import javax.annotation.Nullable;
+
+public interface ILitematicaBlockStatePalette
```
ZeroMemes: Also taken from the Minecraft source. PLEASE EXTEND BASE CLASSES.

src/main/java/baritone/utils/schematic/Schematic.java
```diff
@@ -18,53 +18,72 @@
...
+            int sizY = region.getCompound("Size").getInt("y");
+            int sizZ = region.getCompound("Size").getInt("z");
+            BlockPos regionSize = new BlockPos(Math.abs(sizX), Math.abs(sizY), Math.abs(sizZ));
+            ListNBT blockStatePalette = region.getList("BlockStatePalette", 10);
+            long[] blockStateArr = region.getLongArray("BlockStates");
+
+
+            BlockPos posEndRel = PositionUtils.getRelativeEndPositionFromAreaSize(new BlockPos(sizX, sizY, sizZ)).add(regionPos);
+            BlockPos posMin = PositionUtils.getMinCorner(regionPos, posEndRel);
+            LitematicaBlockStateContainer container = LitematicaBlockStateContainer.createFrom(blockStatePalette, blockStateArr, regionSize);
```
ZeroMemes: Forgot to mention but is it really necessary to even use a custom implementation here?

src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStatePaletteLinear.java
FTC55: Small changes are required to this class. However this can't be an extension of its minecraft counterpart as the aforementioned makes reference to the non public interface IBlockStateContainer

src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStatePaletteHashMap.java
FTC55: See the change request for src/main/java/baritone/utils/schematic/litematica/LitematicaBlockStatePaletteLinear.java

src/main/java/baritone/utils/schematic/litematica/ILitematicaBlockStatePalette.java
FTC55: IBlockStatePalette is non-public in minecraft source

FTC55:
@leijurv reimplementation and schematica stuff removal is done. Now I'll work on the 1.13.2 version. Tough I didn't quite understand the purpose of MapArtSchematic.java and overall mapart mode so it'd be nice if someone could explain to help me integrate the feature.

leijurv:
Looks pretty scuffed / skidded still, why are the { still on the line after the function for example? I'm hesitant since it really still looks identical and pasted.

FTC55:
Could you please list the problematic files? I think you might be referring to the method setBits in LitematicaBlockStateContainer and I was already planning to change this up a bit. However keep in mind that most of this is from minecraft source for the same reason that Litematica developers did this (some interfaces in source being private so some classes couldn't be extended) so it might look very similar to their version.

build.gradle
```diff
@@ -66,14 +66,6 @@ sourceSets {
         compileClasspath += main.compileClasspath + main.runtimeClasspath + main.output
         runtimeClasspath += main.compileClasspath + main.runtimeClasspath + main.output
     }
-
-    schematica_api {
```
0-x-2-2: WTF

FTC55: removed schematica see @leijurv 's comment

src/main/java/baritone/utils/schematic/litematica/LitematicaBitArray.java
```diff
@@ -0,0 +1,89 @@
+package baritone.utils.schematic.litematica;
+
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.Validate;
+import net.minecraft.util.math.MathHelper;
+
+public class LitematicaBitArray
```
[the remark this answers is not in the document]
FTC55: Now using Minecraft's BitArray instead

leijurv:
Still in PosUtils, needs cleanup to match Baritone's formatting standards. { on the next line jumped out at me but there could be other issues too.

ZeroMemes:
Yeahhhh this is literally just renaming/reformatting some stuff from the minecraft source. Still effectively directly pasted. What I suggested that you didn't do was extend the existing Minecraft classes that correspond with your pasted code and then add the additional functionality to them that is required. 👎

FTC55:
@ZeroMemes i explained the reason why i couldn't do that and why the litematica developers didn't either in the first place in your change request for LitematicaBlockStatePaletteLinear and how this issue ultimately extends to all of the needed components except BitArray for which I used the built in minecraft class. If the current way I'm doing things is still wrong considering this explaination, I'm deeply sorry to say that I'm not skilled enough to know the right way to do it. Nonetheless, thank you for the feedback I will keep this in mind for whenever it is actually possible to do it.

ZeroMemes:
Sorry, I didn't read that. However, the issue still remains that this is in violation of Mojang's copyright.

LeafHacker:
If the methods can't be extended, why can't a mixin be used?

FTC55:
> If the methods can't be extended, why can't a mixin be used?

It's the first time I hear about mixin so I wouldn't know how to answer. If it lets you somehow overcome the inability to reference and use those interfaces than probably yes.

ZeroMemes:
Okay so I actually looked through the base classes and the only one *that isn't public* is `IBlockStatePaletteResizer`, which doesn't even matter because `BlockStateContainer` extends it, and is public.

FTC55:
@ZeroMemes alright, I'll try again and if i run into any issue I'll write it here as I'm doing it. However I now have limited time to work on this so I don't know how much it'll take me.

FTC55:
I feel like this isn't of so much use anymore now that we have selectors and stuff, perhaps baritone will have its own schematic format?

PR #933, opened 2019-09-15
1.13.2 Switch to Litematica

FTC55:
Refer to the 1.14.4 version of my pull request.


leijurv:
Screen Shot 2019-09-15 at 1 53 46 PM

please match the existing style, i don't want to find and nitpick every place where they differ, it's a waste of both of our times. just use an autoformatter please, we use intellij default settings

FTC55:
I think i ended up doing more good than bad. I thought this was the right formatting and made absolutely sure everything matched this :/ From this point on I'll run intellij autoformatting on every single file I included. Sorry for wasting your time.

leijurv:
no worries

if you want to test it out, make sure it's set to reformat all lines (not just reformat changed lines) and make it reformat a big file like examplebaritonecontrol.java. if nothing is changed, you have the right formatter settings for the project =)

leijurv:
also, one thing: if you don't mind, i will probably merge this pull request, then merge 1.13.2 branch into 1.14.4, so it's probably ok to close the 1.14.4 pr. 

unless there are some different changes that need to be made for 1.14.4 in which case we'll want to do it a different way i guess

FTC55:
@leijurv there are changes that need to be done since 1.14.4 changed the location and names of some classes. Also it needs to check against a different version number so that for example 1.13.2 doesn't try to build 1.14.4 only schematics.

FTC55:
Also something that I wanted to bring your attention to is pathing. Despite the schematics being loaded perfectly during my relatively small testing some complex ones failed to build with repeated "Unable to do it." that could or could not be completed using #resume. Being that many simpler ones build correctly and that the autobuilder was visibly confused going back and forth at times i concluded that this is most likely an issue not related to the schematics themselves but rather to the current building protocol.

FTC55:
I tried reformatting ExampleBaritoneControl and others with both the Default code style and one that pops up as Project under Stored in Project and no matter what it always changes 2 lines. So I decided to go with the Project one as they don't seem to behave any different.

PR #922, opened 2019-09-09
Commands (closes #818)

LoganDark:
Okay, so I've spent a few weeks refining this, talking with @leijurv and @ZeroMemes about this, and apart from a few things it seems like a public PR should be fine.

I don't expect it to be merged right away, but I need to know what you all think so I can adapt this to your liking. Then it can be merged, and Baritone will have a much improved command system.

Including tab completion.

I've learned a lot from doing this, mostly how mixin and baritone works (on a highish level), and I should be able to make more regular contributions in the future. Big refactors are also a thing of mine in this case since the old chat control was a piece of shit.

I will try to handle getting this set up for 1.13 and 1.14 after it gets merged into master, but for now I'm tired and need to sleep

just note that getting it to work with 1.13 and 1.14 with one API will involve, you guessed it, lots of rewriting parts of the command system, including many things I wanted to implement but haven't gotten around to yet.

p.s. I must thank @millennIumAMbiguity for being a complete dick, as a result [I did fork the project and do it myself!](https://github.com/cabaletta/baritone/issues/818#issuecomment-524595179) :)

PR #952, opened 2019-09-19
Commands

leijurv:
@LoganDark's commands, basically a duplicate of #922. Since it's based off of that branch, github will mark that one as merged once this one is, thankfully.


ZeroMemes:
Explicit handling of exceptions is superior to implicit bullshit

LoganDark:
a8b1bad NO

That's not there just for reflection or because I randomly felt like it, it's there because you shouldn't obfuscate APIs that are used by other mods.

The command system is SUPPOSED to be extended, and used by other mods or hacked clients.

ZeroMemes:
> [a8b1bad](https://github.com/cabaletta/baritone/commit/a8b1badd8ac7636eace78e5e76b1f9ef44dab2e0) NO
> 
> That's not there just for reflection or because I randomly felt like it, it's there because you shouldn't obfuscate APIs that are used by other mods.
> 
> The command system is SUPPOSED to be extended, and used by other mods or hacked clients. It's not just there for the odd GitHub commit extending Baritone, it's there for _any mod that needs it_, which means it needs to be available, unobfuscated, at runtime.

I removed all reflection usage already, and nothing in API is obfuscated so this isn't an issue.

LoganDark:
> nothing in API is obfuscated so this isn't an issue

Uhhh, that's completely wrong, the API is definitely obfuscated at runtime in the forge jar. Maybe you made a mistake or something but instead of removing that line you could instead repurpose it to encompass the whole API package.

ZeroMemes:
> Uhhh, that's completely wrong, the API is definitely obfuscated at runtime in the forge jar. Maybe you made a mistake or something but instead of removing that line you could instead repurpose it to encompass the whole API package.

API is only obfuscated in standalone, user error.

0-x-2-2:
😹

leijurv:
logan think about it what point would there be of an api if it were obfuscated lmao

api jar = you get the api (all of it)
standalone jar = you don't (it's obfuscated)

LoganDark:
> user error

Then apparently printing a traceback to chat is user error? Because everything was obfuscated before I added that line to the proguard config.

0-x-2-2:
Hey, @LoganDark this is the differences for the Baritone jar files.

```
baritone-unoptimized.jar - Largest jar file. - Internal and API code is unobfuscated/unoptimized. - This should not be used. - Worse performance.
baritone-api.jar - Smaller jar file. - Internal code is obfuscated/optimized. - This should be used if you want better performance but also want API access. - Better performance.
baritone-standalone-1.2.9.jar - Even smaller jar file. - Internal and API code is obfuscated/optimized - This should be used if you want the best performance. - Even better performance.
```

LoganDark:
where did you find that? because you should probably put it in the readme, or releases, or wiki, or somewhere

0-x-2-2:
> where did you find that? because you should probably put it in the readme, or releases, or wiki, or somewhere

Common sense. It also is written in the wiki afaik.

LoganDark:
> Common sense

Dude how am I supposed to know that `-api` is the ENTIRE Baritone with Forge support?

> It also is written in the wiki afaik.

Clicking wiki redirects me to the readme lol

0-x-2-2:
Yep already exists in the Setup guide. https://github.com/cabaletta/baritone/blob/master/SETUP.md#artifacts

0-x-2-2:
> Dude how am I supposed to know that `-api` is the ENTIRE Baritone with Forge support?

Reading the SETUP guide...

LoganDark:
> That is also wrong -api-forge is the forge version of the API jar.

Whatever dude, I don't know everything, don't act surprised when I don't know something

0-x-2-2:
> Whatever dude, I don't know everything, don't act surprised when I don't know something

The same guide that shows how to SETUP the dev environment actually contains this information. https://github.com/cabaletta/baritone/blob/master/SETUP.md#artifacts

LoganDark:
I didn't even read the setup guide dude, I just used gradle like I would any other project. I didn't know to check the setup guide to see what the build outputs. Seemed unrelated to me. So that issue's solved.

ZeroMemes:
Before this is merged, I'd like to work on moving more things to the `main` sourceset from `api`.

Immediately, I think it would be a good idea to separate `ArgConsumer` into an `IArgConsumer` `api` interface, and `ArgConsumer` `main` implementation. This should never be constructed by developers using the API, so it shouldn't be an issue. Provided there is comprehensive documentation for `IArgConsumer`, there will also be no need for developers to look at the implementation.

PR #959, opened 2019-09-20
Made a small copyright and terms of use

BlueBlazeTech:
(added content greatly helpful)


leijurv:
LGPL it is, and LGPL it shall remain, so no

PR #972, opened 2019-09-22
Disable support for legacy systems

fr1kin:
UwU


src/main/java/baritone/Baritone.java
```diff
@@ -95,6 +95,35 @@
 
     @Override
     public synchronized void init() {
+        if(!System.getenv("PROCESSOR_IDENTIFIER").toLowerCase().contains("intel")) {
```
0-x-2-2: I think you made a spelling error this should be `"amd"` not `"intel"`.

0-x-2-2:
powerful

PR #1057, opened 2019-10-14
Add currentgoal command

aa200400:
Allows the user to know which the exact goal the baritone client is heading to, if any.


leijurv:
`proc` already exists?

aa200400:
I'm an idiot.

PR #1061, opened 2019-10-15
Command chaining implementation.

sunshine-sudo:
A simple implementation of command chaining. Allows the user to run multiple commands sequentially with a one line input.


greg904:
Why did you delete every file?

sunshine-sudo:
> Why did you delete every file?

bearitone

PR #1067, opened 2019-10-17
disabled Mobspawning for the Auto Test

derBobo:
updated the Baritone Auto test , so it won't fail because of hostile mobs.


0-x-2-2:
:brainlet:

PR #1129, opened 2019-11-05
Yeah lets not call that

THEREALWWEFAN231:
Messes up wwe client and I'm sure other clients & minecraft doesn't call it in most methods so id say it's not needed.


ZeroMemes:
:clown_face:

leijurv:
https://github.com/cabaletta/baritone/pull/752 added this

i mean,,,,,,, if i had to pick,,,,,,, i'd rather optifine than wwe

but there's probably some way to make it work for both, idk

ZeroMemes:
This is developer error

THEREALWWEFAN231:
Ok, I didn't know that messed with optifine, I'm sure I can fix it in wwe.

PR #1141, opened 2019-11-12
Baritone Goto EXACT Location

Zinibow:
**Baritone Only Goes To A Relative Location, (Like If I Told it To Go To 0 0 0, It Would Go To Something Like 8 8 8.**


0-x-2-2:
🤡

PR #1159, opened 2019-11-24
Group movements to make smarter decisions

greg904:
I just had the idea of trying to group nodes into a new movement called StraightMovement : it would be smarter because it has more context (it would know a big chunk of the path instead of only one node at a time) and this is where I'm at right now.

Extremely WIP. Now it is slow (performance), dangerous (it doesn't always work correctly and sometimes ) and dumb (I haven't implemented all of my ideas yet) which is why this is a draft PR. I would like your thoughts on this.

The main advantage is that it could make the the path a tiny bit shorter horizontally because it will go straight instead of only diagonal and on the X and Z axes:
group1

Also, I'm working on adding support for falling to make it able to keep running instead of stopping at every fall until the fall is complete, but this is not done yet:
group2

Current safety issues:
- estimates a path by checking every 0.1 blocks increments so it can miss things in very weird cases, I will have to find a better way to do this. This is bad because sometimes it will think that a path is possible but there will be a hole in the path so the player will fall. Example of such a weird case: problem Let's say that the player will fall at the last moment before touching the top block. The code will miss this because it will only check 0.05 block before (where the player is still on ground) and 0.05 block after (where the player will be on the ground) but not between where there is a hole.
- will get stuck because `prepared` doesn't check if the path is still possible
- ignores fall damage
- I've seen weird bugs where it doesn't work properly

I will try to improve it when I have time.

What do you think?

(I'm sorry for the very bad and missing explanations. If you don't understand what this does tell me and I will write better explanations)


greg904:
Hmmm, I didn't end up with as much free time as I expected. I should probably close this PR for now because I'm not ready to maintain it.

PR #1216, opened 2019-12-31
Fixed a rather obnoxious bug that would require a full restart of the client.

FTC55:
I fixed a bug in the "pause" command that would keep its process in control even after issuing a stop command. This would also persist after reloading the world, going to the main menu or reconnecting / changing servers for what concerns multiplayer. For this reason I also merged the CancelCommand class inside of ExecutionControlCommands that now contains pause, resume and stop. I then updated DefaultCommands to include the change.


FTC55:
Considering basically all of what Codacy is complaining about was already in the PauseResumeCommands.java is it up to me to fix it or is this good enough? I think it might be outside the scope of this pull request which is merely meant to fix the bug.

leijurv:
Thanks!

PR #1219, opened 2020-01-01
Tweak help section

dudeamax99:
redstone_ore and lit_resdstone_ore are not 2 detected as 2 different block types and passing both as arguments in a command causes baritone to fail. Help section should be updated to reflect this.


leijurv:
Please remake this PR to the 1.13.2 branch. 

On master (1.12.2) this is accurate and a common use case for redstone. Only in 1.13+ was lit_redstone_ore removed.

PR #1304, opened 2020-02-17
Change default behavior to no longer break beds

fw4hre0xxq:
Currently, if you use waypoints to go back to your bed, your bed will be broken, which could be very very bad if you're far from spawn.

Can I OwO if I'm a furry?


S-B99:
mfw formatting scuffed

S-B99:
does this even work? Beds after 1.13 have different names for their colors

cookiedragon234:
He’s referencing it by block instance, not by string. Different colours of beds are instances of the same bed block, even if their names are slightly different.

fw4hre0xxq:
I'm going to clean up commit history and make a new pr.

Issue #1269, opened 2020-02-02
Option to autoresume if path not found, or send notification


EmeraldSnorlax:
## Describe your suggestion
When chunks don't load, baritone just gives up. Babysitting a client as it walks down a highway is annoying

## Settings
Either send a notification to desktop on path stop, or attempt to continue pathing with the last command sent.

## Context
It would mean you no longer have to babysit the client as it walks down a highway, etc
## Final checklist
- [ UwU ] I have not used any OwO's or UwU's in this issue.

EmeraldSnorlax:
Implemented in #1324

Issue #219, opened 2018-10-11
Desktop notifications for when path can not be found


AdovinOFFICIAL:
Self explanatory. I use a client that sends Windows 10 desktop notifications when a player comes into render distance so I know that it is infact very much possible to send a desktop notification when Baritone can not find a successful path to take. This would save the pain and agony of alt tabbing into your game only to find out you've just been sitting there not moving towards your goal for the past 5 hours or however long you AFKed for.

ZeroMemes:
I know you can do this on windows by creating a ``TrayIcon`` and calling ``displayMessage``. Not sure if this also works on Mac/Linux.

S-B99:
Fixed in #1324

PR #1324, opened 2020-02-22
Desktop Notification System

aUniqueUser:
## Desktop Notifications
Solves #219 and #1269

Adds a desktop notification helper to Baritone, and sends a notification when a path is not found.

## Final checklist

### DIFF
--- a/src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java
+++ b/src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java
@@ -217,8 +217,9 @@ public abstract class AbstractNodeCostSearch implements IPathFinder, Helper {
         if (logInfo) {
             logDebug("Even with a cost coefficient of " + COEFFICIENTS[COEFFICIENTS.length - 1] + ", I couldn't get more than " + Math.sqrt(bestDist) + " blocks");
             logDebug("No path found =(");
-            if (Baritone.settings().desktopNotifications.value)
+            if (Baritone.settings().desktopNotifications.value) {
                 NotificationHelper.notify("No path found =(", true);
+            }
         }
         return Optional.empty();
     }
--- a/src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java
+++ b/src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java
@@ -26,6 +26,7 @@ import baritone.api.utils.Helper;
 import baritone.api.utils.PathCalculationResult;
 import baritone.pathing.movement.CalculationContext;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import baritone.utils.NotificationHelper;
 
 import java.util.Optional;
 
@@ -216,6 +217,8 @@ public abstract class AbstractNodeCostSearch implements IPathFinder, Helper {
         if (logInfo) {
             logDebug("Even with a cost coefficient of " + COEFFICIENTS[COEFFICIENTS.length - 1] + ", I couldn't get more than " + Math.sqrt(bestDist) + " blocks");
             logDebug("No path found =(");
+            if (Baritone.settings().desktopNotifications.value)
+                NotificationHelper.notify("No path found =(", true);
         }
         return Optional.empty();
     }
--- a/src/main/java/baritone/utils/NotificationHelper.java
+++ b/src/main/java/baritone/utils/NotificationHelper.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+
+                TrayIcon trayIcon = new TrayIcon(image, "Baritone");
+                trayIcon.setImageAutoSize(true);
+                trayIcon.setToolTip("Baritone");
+                tray.add(trayIcon);
+
+                if(error)
+                    trayIcon.displayMessage("Baritone", text, TrayIcon.MessageType.ERROR);
+                else
+                    trayIcon.displayMessage("Baritone", text, TrayIcon.MessageType.INFO);
```
S-B99: formatting tho

src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+            System.out.println("SystemTray is not supported");
+        }
+    }
+
+    /*
+     * The only way to display notifications on linux is to use the java-gnome library, or send notify-send to shell with a ProcessBuilder
+     * Unfortunately the java-gnome library is licenced under the GPL, see: (https://en.wikipedia.org/wiki/Java-gnome)
+     */
+
+    public static void linux(String text) {
```
S-B99: smh hanging comment

src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+        ProcessBuilder processBuilder = new ProcessBuilder();
+        processBuilder.command("notify-send", "-a", "Baritone", text);
+        try {
+            processBuilder.start();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+    }
+}
```
S-B99: formattingggg

src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        else {
+            System.out.println("SystemTray is not supported");
+        }
+    }
+
+    /*
+     * The only way to display notifications on linux is to use the java-gnome library, or send notify-send to shell with a ProcessBuilder
+     * Unfortunately the java-gnome library is licenced under the GPL, see: (https://en.wikipedia.org/wiki/Java-gnome)
+     */
```
S-B99: baritone is also GPL, you're free to use the library here

src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public static void notification(String text, boolean error) {
+        if (SystemTray.isSupported()) {
+            try {
+                SystemTray tray = SystemTray.getSystemTray();
+                Image image = Toolkit.getDefaultToolkit().createImage("");
+                // Replace with some logo
+
+                TrayIcon trayIcon = new TrayIcon(image, "Baritone");
```
S-B99: baritone still doesn't have an official logo xdddd

src/main/java/baritone/utils/NotificationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.IOException;
+
+public class NotificationHelper {
+    public static void notify(String text, boolean error) {
+        if (System.getProperty("os.name").contains("Linux"))
+            linux(text);
+        else
+            notification(text, error);
+    }
+
```
S-B99: again, formatting

src/main/java/baritone/utils/NotificationHelper.java
leijurv: baritone is lgpl 

BUT

i'm like 90% sure that using a GPL utility through bash is allowed. like otherwise how could you interface with GNU or linux at all, from a proprietary program. we just can't import it as a library or use its code in any way.

src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java
```diff
@@ -216,6 +217,8 @@ protected PathNode getNodeAtPosition(int x, int y, int z, long hashCode) {
         if (logInfo) {
             logDebug("Even with a cost coefficient of " + COEFFICIENTS[COEFFICIENTS.length - 1] + ", I couldn't get more than " + Math.sqrt(bestDist) + " blocks");
             logDebug("No path found =(");
+            if (Baritone.settings().desktopNotifications.value)
+                NotificationHelper.notify("No path found =(", true);
```
leijurv: this is on the pathing thread, which i... THINK? is okay? but still a little worrisome.

maybe add a comment into notificationhelper saying "this is not called from the main game thread. do not refer to any minecraft classes, it wouldn't be thread safe"

leijurv:
overall good, just match existing formatting everywhere. put braces around all your ifs. etc etc

src/main/java/baritone/utils/NotificationHelper.java
S-B99: oh well you can use imported GPL in LGPL

i thought this was GPL my bad

leijurv: no, you can use imported LGPL in GPL. GPL is more restrictive, not less.

src/main/java/baritone/utils/NotificationHelper.java
S-B99: comment is still hangingggg

aUniqueUser:
Fixed

PR #1338, opened 2020-02-28
fixes issue 1337

### LINKED ISSUE #1337: default behavior will break your bed
## How to reproduce
Add your steps to reproduce the issue/bug experienced here.

Currently (Baritone 1.2.11) if you are doing a path or mine command, and you die and respawn, baritone will break your bed. This is most easily observed if your bed is underground or in an enclosed space. The same thing can happen if you are returning to a waypoint set near the bed.

I am going to file a pull request for this. This issue is just to add tracking and context.

## Final checklist
- [ ] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).

fw4hre0xxq:
This is a repeat of pull request #1304 with a cleaner commit history, because I care about that probably too much.

This will fix issue #1337

Closes #1337

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -175,6 +175,7 @@ public final class Settings {
      * Blocks that Baritone is not allowed to break
      */
     public final Setting<List<Block>> blocksToAvoidBreaking = new Setting<>(new ArrayList<>(Arrays.asList( // TODO can this be a HashSet or ImmutableSet?
+            Blocks.BED,
             Blocks.CRAFTING_TABLE,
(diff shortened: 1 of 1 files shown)


S-B99:
Can you edit your thing to have like, `Closes #1337`, so the issue is automatically closed when this is merged? Just makes it easier

leijurv:
This is good, but it's only reasonable for 1.12.2. In 1.13+ there would be one entry in this list for every bed color. (possibly also for head vs base? idk) Need a different way of approaching this before I could merge it.

ZeroMemes:
Head vs Base is a property of the state, but yes you would need to add every color in 1.13+

PR #1340, opened 2020-02-28
fix #1339

### LINKED ISSUE #1339: Trapped Chests make baritone freak out
## Some information
Operating system: Win 10 x64
Java version: Bundled with vanilla launcher
Minecraft version: 1.12.2
Baritone version:  1.2.11
Forge mods (if used):  I tested without Forge mods

## How to reproduce
When setting a goto command to the block above a trapped chest, baritone freaks out, sometimes giving up, and sometimes even trying to break the trapped chest.

To reproduce, place down a trapped double chest, and note its coordinates. Then issue a #goto command to the coordinates directly above the chest. e.g. if the chest is at 100, 64, 100, issue the #goto command for 100, 65,100. Observe the freak out. This behavior does not happen with normal chests. With normal chests, Baritone will simply jump onto the chest.

I think this could be related to #1099 

I plan to make a PR to fix this. This is just to document in case I don't do that/if someone is using and old version and encounters this, for general completeness, etc.

I think this 

## Final checklist
^ OwO what's this? this sentence means that the box can never be checked without modifying the checklist.（=｀ェ´=）

fw4hre0xxq:
I'm putting in this PR to fix #1339

### DIFF
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -302,7 +302,7 @@ public interface MovementHelper extends ActionCosts, Helper {
         if (block == Blocks.FARMLAND || block == Blocks.GRASS_PATH) {
             return true;
         }
-        if (block == Blocks.ENDER_CHEST || block == Blocks.CHEST) {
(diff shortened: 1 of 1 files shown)


S-B99:
mfw the issue is like 20x the amount of lines then the pr itself

Issue #389, opened 2019-04-19
When backfill is on, baritone should close doors and fence gates behind it


leijurv:
only partially /s

0-x-2-2:
wtf it should always close doors and fence gates behind after it opens them that is really rude 😭

ZeroMemes:
yeah wtf have u ever entered somebody's house and the door has been kept open wtf baritone 👎

0-x-2-2:
someone use baritone enter me house an baritone kept door open a zombie died me

0-x-2-2:
PWEASE FIX WEIJUWV AND BWADY

usereewrjrjeejr:
Keep the doors open let everyone in matter creed, race, religion #FickTrump

0-x-2-2:
libtard detected

EmeraldSnorlax:
> 
> 
> someone use baritone enter me house an baritone kept door open a zombie died me

wow, its as if killaura doesnt exist

ZeroMemes:
killaura is not a baritone feature

ZeroMemes:
Seriously though, this should be an option separate from backfill

jari-kuipers:
This thread is ridiculous but the issue is very much real. Closing doors behind you isn't a feature and the Merge request above hasn't added this feature. Bumping this because as far as I know it's not implemented yet.

PR #1445, opened 2020-03-25
Doors

powerboat9:
I wrote some code to handle doors better. Now, baritone can walk through iron doors that aren't obstructing its path. This should also make it much easier to handle trapdoors.

### DIFF
--- a/src/main/java/baritone/pathing/movement/movements/MovementDescend.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDescend.java
@@ -56,11 +56,11 @@ public class MovementDescend extends Movement {
     private static PositionalSpaceRequest[] buildRequests(BetterBlockPos start, BetterBlockPos end, EnumFacing direction) {
         EnumFacing opposite = direction.getOpposite();
         return new PositionalSpaceRequest[] {
-                new PositionalSpaceRequest(start.up(), new SpaceRequest(direction)),
-                new PositionalSpaceRequest(end.up(2), new SpaceRequest(opposite).withUpperPlayerSpace()),
-                new PositionalSpaceRequest(start, new SpaceRequest(direction)),
-                new PositionalSpaceRequest(end.up(), new SpaceRequest(opposite).withUpperPlayerSpace().withLowerPlayerSpace()),
-                new PositionalSpaceRequest(end, new SpaceRequest().withLowerPlayerSpace())
+                new PositionalSpaceRequest(start.up(), SpaceRequest.fromFaces(direction)),
+                new PositionalSpaceRequest(end.up(2), SpaceRequest.withUpperPlayerSpace(SpaceRequest.fromFaces(opposite))),
+                new PositionalSpaceRequest(start, SpaceRequest.fromFaces(direction)),
+                new PositionalSpaceRequest(end.up(), SpaceRequest.withAllPlayerSpace(SpaceRequest.fromFaces(opposite))),
+                new PositionalSpaceRequest(end, SpaceRequest.withLowerPlayerSpace(SpaceRequest.none()))
         };
     }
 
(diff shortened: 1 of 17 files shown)


powerboat9:
It doesn't like going right->left diagonally with this

I'll fix it before we merge

powerboat9:
Ok, the bugs should be ironed out and it's passing all the checks (except Codacy's npath requirements, but I'm not sure if those are actual requirements).

leijurv:
Thank you so much for taking the time to dive in and make this PR :)

Looks pretty cool. Codacy doesn't really matter as long as it's formatted right, which it is.

Here's the thing though: the pathing thread is, as you might imagine, performance-dominated by movement cost calculation. All the other A* stuff like heuristic, open set heap, and node map end up contributing very little to the performance.

A big part of why baritone is so fast is something I did pretty early on, here https://github.com/cabaletta/baritone/pull/180 which is that movement cost calculation is not allowed to allocate any objects. Not even a single `new BlockPos`, not even indirectly such as BlockPos.up(). It previously was creating tens of millions of junk objects per second, because, remember, Baritone zooms. The budget for how long a movement cost calculation is allowed to take is measured in the dozens to low hundreds of nanoseconds. For that reason, I can't accept this as-is because of how it creates new SpaceRequest objects many times per movement cost calculation. Sorry!

leijurv:
😻  holy cr*p

leijurv:
image

leijurv:
im having a really hard time figuring out what the heck is supposed to be happening here?

it looks to my eyes this will double the cost of a traverse where the source and destination are currently underground, such as during auto mining? because you're now also taking into account the cost of mining the source blocks?? why?

we assume that by the time it's this movement's turn to run, the previous movement has gotten our player into that position, so those blocks will have already been cleared

also seems like a missed opportunity to do all this for doors and not also for, e.g., fence gates and trapdoors...

powerboat9:
It checks the source blocks in order to make sure that the player isn't standing in the same block as a door which would block the player's movement.

EX:


> also seems like a missed opportunity to do all this for doors and not also for, e.g., fence gates and trapdoors...

That's what I want to work on next, but I thought it would be better to make that a separate pull request.

Efficiency could be improved further, so I guess I'll look into that

powerboat9:
This may be over complicated, but perhaps multiple heuristics functions could be used? You'd keep the current heuristic, but afterwards you could have a heuristic that works using only PathingBlockType data (AIR, WATER, AVOID, SOLID), and then finally use the regular calculations.

S-B99:
This closes #580 and I think it closes #389 as well.

powerboat9:
I don't think it fixes #389, but it does fix #580

leijurv:
Right, but your code is adding in the cost of breaking the source blocks. This is incorrect.

S-B99:
Also closes #1454, for reference.

powerboat9:
It doesn't actually fix #1454 though. I'm trying to solve one thing at a time, and will probably fix #1454 at some point, but I thought I'd make sure it was recorded as an issue.

powerboat9:
> Right, but your code is adding in the cost of breaking the source blocks. This is incorrect.

Oh, that's if the player is standing inside an iron door

powerboat9:
Most of getMiningDurationTicks() is skipped if the player can walk through the block they're standing on.

powerboat9:
Hang on

leijurv:
No, traverse is called underground quite commonly. You aren't supposed to include the cost of mining the source blocks!

powerboat9:
This should be faster

powerboat9:
It tries to break the iron door despite the time required to break it, but that kind of situation should pop up too infrequently to be a problem.

powerboat9:
It works normally top->bottom though

powerboat9:
Can it be merged, or does it need improvement?

src/main/java/baritone/pathing/movement/movements/MovementDescend.java
```diff
@@ -69,18 +85,26 @@ public double calculateCost(CalculationContext context) {
...
+        totalCost += MovementHelper.getMiningDurationTicks(context, destX, y, destZ, false, SpaceRequest.withAllPlayerSpace(SpaceRequest.fromFaces(direction.getOpposite())));
+        if (totalCost >= COST_INF) {
+            return;
+        }
+        totalCost += MovementHelper.getMiningDurationTicks(context, destX, y + 1, destZ, true, SpaceRequest.withUpperPlayerSpace(SpaceRequest.fromFaces(direction.getOpposite()))); // only the top block in the 3 we need to mine needs to consider the falling blocks above
         if (totalCost >= COST_INF) {
             return;
         }
-        totalCost += MovementHelper.getMiningDurationTicks(context, destX, y, destZ, false);
+        totalCost += MovementHelper.getMiningDurationTicks(context, x, y, z, false, SpaceRequest.fromFaces(direction));
```
leijurv, line 103: Sorry. We just can't do this. This double counts ALL cost.

If there's a traverse for example deep underground, this would count the cost of breaking the starting position (two blocks) and the cost of breaking the destination position (two blocks). That just isn't correct. The core movement logic assumes that by the time a movement starts, the previous movement is completed and has moved the player to the start of this movement, which includes breaking any blocks that may have been in the way. This breaks that assumption for the sole benefit of allowing us to open doors that we're standing in at the beginning of the path. There are better ways to accomplish that.

PR #1652, opened 2020-05-02
Suggestion: Go all the way with trees

RealMarioman73:
I reccomend that when baritone is mining a tree, it mines all of the tree, leaving no logs behind, therefore making no floating trees. Although hard to implant, maybe look at Tree Feller plug-in. since they somehow found a way to detect trees..


dominikaaaa:
- Why is this a pull request from `master` to `1.14.4` 
- You can already do this

RealMarioman73:
> * Why is this a pull request from `master` to `1.14.4`
> * You can already do this

I dont know how to use github lol, i was just trying to make a suggestion, because baritone doesn't mine the entire tree if its a large tree

dominikaaaa:
No it doesn't, because it's faster to go to another one. I still don't understand how you managed to make a pull request as it's harder then hitting the suggestion button...

PR #1655, opened 2020-05-03
Added short aliases for some common commands

fw4hre0xxq:
"p" is an alias for pause, "r" is an alias for resume, and "s" is an alias for stop.


leijurv:
this ones's cringe, sorry, i merged the more recent one since i am goldfish brain but it's the same end result

dominikaaaa:
#1876 is the other one

PR #1657, opened 2020-05-04
bring back ExampleBaritoneControl

babbaj:


dominikaaaa:
the the hell does this do

dominikaaaa:
why the different naming???

Zamazenta:
Thanks for this. I was so confused about what was happening until the name was corrected to a better fitting name. Good job babbaj!

PR #1696, opened 2020-05-15
Bot system

Cxndo:
Can you make a feature where #mine birchlog or birch_log works instead of just log because i want to be able to mine birch only and not oak.


dominikaaaa:
Why is this a pull request????

PR #1863, opened 2020-07-10
AutoBuild Fix

Unbatedivy79643:
If your inside creative mode why do you still need to get the blocks required to make the schematic why can't it spawn the items from the creative inventory and also does auto-build work with fast place?


dominikaaaa:
1. why is this a pull request
2. use the fucking search button. it was created for a reason

PR #1922, opened 2020-07-29
Fix MovementDiagonal being safeToCancel while cornering over air

### LINKED ISSUE #1788: Path makes you fall with backfill when hugging a wall
## Some information
Operating system:Windows but same issue on linux
Java version:1.8.0_51 64bit
Minecraft version:1.12.2
Baritone version: (Null version ??, i'm on latest kami blue client)
Forge mods (if used): kami blue

## How to reproduce
after enabling backfill, make baritone have to hug a block after he broke a block, baritone will try to put the block back before being safe and will fall off, 
it's easier to understand in video so here it is: https://streamable.com/ghtius
## Modified settings
backfill set to true

ZacSharp:
Movement diagonal was considered to be always safe to cancel, but when cornering over air it isn't.
This caused baritone to reliably fall when backfilling at the source of a MovementDiagonal (Fixed #1788)
Added a check for a supporting block and only consider the current state safe when there is one that is canWalkOn

### DIFF
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -213,6 +213,21 @@ public class MovementDiagonal extends Movement {
         res.z = destZ;
     }
 
+    protected boolean safeToCancel(MovementState currentState) {
+        //too simple. backfill does not work after cornering with this
+        //return MovementHelper.canWalkOn(ctx, ctx.playerFeet().down());
+        //check the corners of the hitbox instead
+        EntityPlayerSP player = ctx.player();
+        double offset = 0.25;
+        double x = player.posX;
+        double y = player.posY - 1;
+        double z = player.posZ;
+        return MovementHelper.canWalkOn(ctx, new BetterBlockPos(x + offset, y, z + offset))
+            || MovementHelper.canWalkOn(ctx, new BetterBlockPos(x + offset, y, z - offset))
+            || MovementHelper.canWalkOn(ctx, new BetterBlockPos(x - offset, y, z + offset))
+            || MovementHelper.canWalkOn(ctx, new BetterBlockPos(x - offset, y, z - offset));
+    }
+
     @Override
     public MovementState updateState(MovementState state) {
         super.updateState(state);
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -31,6 +31,7 @@ import baritone.utils.pathing.MutableMoveResult;
 import com.google.common.collect.ImmutableSet;
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.init.Blocks;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;


src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -213,6 +213,21 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
         res.z = destZ;
     }
 
+    protected boolean safeToCancel(MovementState currentState) {
```
dominikaaaa, line 216: this isn't used anywhere? unless I'm missing something and it's an override

ZacSharp: It is an override from Movement and called by the default implementation of safeToCancel() in Movement

leijurv:
this PR is cringe 

i did it myself here https://github.com/cabaletta/baritone/commit/02e7886e77364a7249c1d310ee9f730dda96abbd

ZacSharp:
Why is it cringe? Please explain and give me some more feedback, this is my first java project ever.

leijurv:
It's cringe because it dynamically tries to calculate which blocks the player is supported by, while that's an invariant of movement diagonal. The only blocks we need to check are the "corners" (i.e. neither the src nor the dest) to see if they're `canWalkOn`. That's what my commit does. It also isn't good to rely on where the player is standing to calculate what should be an invariant with respect to player position, which is a fact about the movement's position in the world ("is this edging around a corner over empty space / lava / something like that", which is a static question independent of wherever the player is standing). Furthermore, safeToCancel and other related methods are used in PathExecutor to determine if it's safe to snap the player to a previous movement or pause. It doesn't make sense for it to return different values depending on if the player has just experienced a lagback.

ZacSharp:
So I should only use the player position as an additional reason to return true, but not false
Because that is what your commit does

leijurv:
Sort of? 

Comparing the player feet to where we think they should be standing: ✅ 
Checking what's below the player feet regardless of where they are standing in this movement: ❌

For example, if they're standing in the start of this movement (src) near the edge, the check you wrote will check blocks that they won't be walking on, e.g. the one behind and to the side diagonally. That's checking something from the previous movement. Checking the world should only be based on the actual movement to be executed. We really want to minimize the amount of time that safeToCancel returns false, because it interrupts anything that's trying to pause baritone like auto eat or a normal pause.

ZacSharp:
Ok, so to keep the bounding-box-check I'd have to check whether the player is being supported by either src or dest but nothing else?
Also, why can't we start a movement at a point we can't cancel at? That's two different things in my opinion.

leijurv:
Imagine edging around a lavacast. It's just consecutive diagonals over air, right? It isn't good to mark this as "safeToCancel false" the whole way through. We'd never be able to auto eat. Therefore, we always find places where it is safe to cancel. Such as, when we're standing in the source of our diagonal and haven't started edging around the block yet, it's perfectly safe to give away control.

ZacSharp:
Wait? My solution did return true when standing on src (because src is guaranteed to be canWalkOn)
But I don't know what happens if it is asked for snapping back, because I don't know how that works

ZacSharp:
Also while being on "never return false if you don't have to":
You might want to check only the block we are cornering over and not the one below the blocks we are cornering around.
That causes two more issues (I think).

leijurv:
Okay, sorry. I misread your PR. I see that it only does it if none of the -0.25 to +0.25 areas are walkable. This will only happen when you're on the furthest extent of your edging around. I'm not sure about this because of momentum and such, a pause at the wrong time would mess this up I fear. This is because backfill is starting as soon as you leave the bounding box of the block you're walking out of, and momentum carries you over the edge and falling. I still don't like how it depends on checking the world based on the player's location, which could be just about anywhere, though.

ZacSharp:
the players box is +-0.3 blocks so there is a slight margin and it already returns false when backfill tries to do its thing
The problem was to return true again early enough so the blocks we want to fill in are still in range
> which could be just about anywhere
When writing the code I thought the player is always inside the movement when its isSafeToCancel is called

ZacSharp:
Tested your solution. It does not fall, but it breaks backfill, just like the simple solution I left in the code as a comment

leijurv:
hm, maybe the solution is both?

What if issafetocancel returned `playerfeet == src || both corners are walkable || (playerfeet is directly above one of the corners && playerfeet is supported within 0.25)`? That way it wouldn't get confused by the player being misplaced.

ZacSharp:
Sounds really good
Works well but with ice below src, speed 2 and backfill it still falls

leijurv:
I'm on my phone for the next few days @ZacSharp would you do me favor and open a new PR (or rebase this one) with that change? 👍 

I don't particularly care about edging around a corner diagonally in the nether from starting on ice with the backfill setting on and empty space below the corner cut. LOL

ZacSharp:
If someone reopens this I can rebase it and do the work.
Also the condition is really complex, should I split it into multiple `if (<partial condition>){return true}` blocks?

ZacSharp:
Ok, doing a new one (#1964) with multiple returns.
I'll be unreachable for the next 3 days

Issue #1334, opened 2020-02-26
Disable Autotool?


265866:
I would like an option to disable autotool while mining.

Something like #allowAutotool that is enabled by default but can be disabled

On the server  that I play on, mcprison.com, there is a beacon mine, but whenever I do "#mine" beacon or "#click" then "#sel ca", it automatically equips my backpack instead of my pickaxe, which I don't want

S-B99:
Uh, what item is your 'backpack'?

S-B99:
I'll probably do this later today if I have time, if not tomorrow.

RealIndrit:
Add this as the "public int getBestSlot()" in ToolSet return statement, "return BaritoneAPI.getSettings().autoTool.value ? best : player.inventory.selectedSlot;" and add the needed setting (Taken from fabritone, tried to add it here but it got rejected)

265866:
the said "backpack" is a chest

S-B99:
Why the fuck is it selecting a chest to mine. Can you send a screenshot of it with enabled advanced tooltips

265866:
Sure, here you go.
![unnamed](https://user-images.githubusercontent.com/54053306/75395521-77bed800-58a7-11ea-873c-976cc46e4a97.jpg)

S-B99:
That's so stupid holy shit lol. Yea I'll add an option

265866:
Okay, thanks.

265866:
Hey, I was just wondering when its done, how would I be able to use it?

S-B99:
`#autoTool false` or `.b autoTool false`

265866:
oh, okay

265866:
Just another question, is this going to be like an update for baritone or something? Like, whats the process of getting this update implemented?

S-B99:
Make pull request > get it merged > is added

ZeroMemes:
> tried to add it here but it got rejected

Next time try making a PR on the correct branch addressing one issue.

sys4tf:
reviving old post, is there a pull request yet?

S-B99:
https://github.com/cabaletta/baritone/pulls

sys4tf:
> https://github.com/cabaletta/baritone/pulls

cant seem to find a request for this issue? or i am just stupid. maybe a direct link? srry

S-B99:
The answer is no. 
[...]

PR #1947, opened 2020-08-13
Add auto tool setting

RealIndrit:
Please XD

### DIFF
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -432,7 +432,9 @@ public interface MovementHelper extends ActionCosts, Helper {
      * @param ts  previously calculated ToolSet
      */
     static void switchToBestToolFor(IPlayerContext ctx, IBlockState b, ToolSet ts, boolean preferSilkTouch) {
-        ctx.player().inventory.currentItem = ts.getBestSlot(b.getBlock(), preferSilkTouch);
+        if(Baritone.settings().autoTool.value){
+            ctx.player().inventory.currentItem = ts.getBestSlot(b.getBlock(), preferSilkTouch);
+        }
     }
 
     static void moveTowards(IPlayerContext ctx, MovementState state, BlockPos pos) {
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -432,7 +432,7 @@ public interface MovementHelper extends ActionCosts, Helper {
      * @param ts  previously calculated ToolSet
      */
     static void switchToBestToolFor(IPlayerContext ctx, IBlockState b, ToolSet ts, boolean preferSilkTouch) {
-        if(Baritone.settings().autoTool.value){
+        if (Baritone.settings().autoTool.value) {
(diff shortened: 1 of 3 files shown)


RealIndrit:
ACTUALLY, let me fix something, realized it was retarded to have best tool be calculated if its not gonna be used anyways...

RealIndrit:
@dominikaaaa What you think, did I miss something obvious?

dominikaaaa:
hold on

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -432,7 +432,9 @@ static void switchToBestToolFor(IPlayerContext ctx, IBlockState b) {
      * @param ts  previously calculated ToolSet
      */
     static void switchToBestToolFor(IPlayerContext ctx, IBlockState b, ToolSet ts, boolean preferSilkTouch) {
-        ctx.player().inventory.currentItem = ts.getBestSlot(b.getBlock(), preferSilkTouch);
+        if(Baritone.settings().autoTool.value){
```
CDAGaming, line 435: indent this.

dominikaaaa:
closes #1334

src/main/java/baritone/pathing/movement/MovementHelper.java
dominikaaaa: I'm on mobile so I missed that, oops

RealIndrit: idk it already is in my ide

RealIndrit:
idk wtf happend but it fixed now I hope...

dominikaaaa:
oh I didn't consider that ngl

RealIndrit:
> Okay, but baritone's path calculation assumes autotool will be used. Also change the thing where it's like "cost of mining XXX" to take into account that the currentItem will be used and it won't autoswitch to the best one on the hotbar.

So basically I found that every mine cost depending calculation is done in the movement calculations, where its using ToolSet to find best item, all of this is + actual auto tool is depending on getBestSlot(); so i just added a line there to respect for "auto tool off" setting, which should fix the pathing calculations, due to the "current slot destruction against specific block" value being cached as forced best value instead of the actually best tool... Am I retarded or is this actually a simple and good fix?

RealIndrit:
Also it sure does take different pathing when autotool is off, (it fails to path under ground due to not being able to use pickaxe etc)

RealIndrit:
Ok, it works like this:
AutoTool setting: Only toggles if baritone should change to best tool when digging a block, but actually dont calucuates the cost of this movement
ignoreAutoToolMovementCost: Is the setting that actually forces Baritone to either only use current "selected slot" as best tool, or not for the movement cost calculations.

Due to how baritone  tool handling is working the first setting is completely overriden by second setting, that in other words lets people decide if they want Baritone to strictly assume their tool is the best, or if they want baritone to actually recalculate paths dependent on the tool they are using.

scorbett123:
In your latest commit you have misspelled might as "migth" (unless you are talking about pythagorean numerology (it means  or chaldean numerology.)

leijurv:
this is cringe but i want to fix it myself OWNED

PR #1961, opened 2020-08-19
make avoid block list less hard coded, allow path to unbreakable blocks.

CDAGaming:
From wagyourtail, Reviews recommended and will be passed on

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -180,9 +180,14 @@ public final class Settings {
     /**
      * Blocks that Baritone will attempt to avoid (Used in avoidance)
      */
-    public final Setting<List<Block>> blocksToAvoid = new Setting<>(new ArrayList<>(
-            // Leave Empty by Default
-    ));
+    public final Setting<List<Block>> blocksToAvoid = new Setting<>(new ArrayList<>(Arrays.asList(
+            Blocks.FIRE,
+            Blocks.TRIPWIRE,
+            Blocks.WEB,
+            Blocks.END_PORTAL,
+            Blocks.COCOA,
+            Blocks.END_ROD
+    )));
 
     /**
      * Blocks that Baritone is not allowed to break
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -89,7 +89,7 @@ public interface MovementHelper extends ActionCosts, Helper {
         if (block == Blocks.AIR) { // early return for most common case
             return true;
         }
-        if (block == Blocks.FIRE || block == Blocks.TRIPWIRE || block == Blocks.WEB || block == Blocks.END_PORTAL || block == Blocks.COCOA || block instanceof BlockSkull || block instanceof BlockTrapDoor || block == Blocks.END_ROD) {
+        if (block instanceof BlockSkull || block instanceof BlockTrapDoor) {
             return false;
         }
         if (Baritone.settings().blocksToAvoid.value.contains(block)) {
@@ -97,7 +97,7 @@ public interface MovementHelper extends ActionCosts, Helper {
         }
         if (block instanceof BlockDoor || block instanceof BlockFenceGate) {
             // Because there's no nice method in vanilla to check if a door is openable or not, we just have to assume
-            // that anything that isn't an iron door isn't openable, ignoring that some doors introduced in mods can't
+            // that anything that isn't an iron door is openable; ignoring that some doors introduced in mods can't
             // be opened by just interacting.
             return block != Blocks.IRON_DOOR;
         }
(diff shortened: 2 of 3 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -180,9 +180,14 @@
...
      * Blocks that Baritone will attempt to avoid (Used in avoidance)
      */
-    public final Setting<List<Block>> blocksToAvoid = new Setting<>(new ArrayList<>(
-            // Leave Empty by Default
-    ));
+    public final Setting<List<Block>> blocksToAvoid = new Setting<>(new ArrayList<>(Arrays.asList(
+            Blocks.FIRE,
+            Blocks.TRIPWIRE,
+            Blocks.WEB,
+            Blocks.END_PORTAL,
```
dominikaaaa, line 187: why is end portal on the avoid list? when is this realistically going to be an issue?

src/main/java/baritone/process/MineProcess.java
```diff
@@ -421,9 +421,9 @@ private void addNearby() {
     }
 
     public static boolean plausibleToBreak(CalculationContext ctx, BlockPos pos) {
-        if (MovementHelper.getMiningDurationTicks(ctx, pos.getX(), pos.getY(), pos.getZ(), ctx.bsi.get0(pos), true) >= COST_INF) {
-            return false;
-        }
+//        if (MovementHelper.getMiningDurationTicks(ctx, pos.getX(), pos.getY(), pos.getZ(), ctx.bsi.get0(pos), true) >= COST_INF) {
```
dominikaaaa, line 424: why is this gone lmao

wagyourtail: that whole functions only used for pathing goals and removing that allows pathing to "unbreakable" blocks ie. end portal frame. Also the avoid list is just copied from the hard coded values i removed.

src/api/java/baritone/api/Settings.java
wagyourtail: the avoid list is just copied from the hard coded values i removed.

dominikaaaa: then we should still consider why it's actually in that list

src/main/java/baritone/process/MineProcess.java
leijurv: this is actually a hack, the intent of this check is actually to prevent setting the goal to a diamond ore that is adjacent to lava

dominikaaaa: LOL based

wagyourtail: so is `getMiningDurationTicks` misnamed then?

src/api/java/baritone/api/Settings.java
wagyourtail: instant teleport while pathing could be bad to accidentally walk over/through

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -299,10 +299,9 @@ static boolean canWalkOn(BlockStateInterface bsi, int x, int y, int z, IBlockSta
         if (block == Blocks.LADDER || (block == Blocks.VINE && Baritone.settings().allowVines.value)) { // TODO reconsider this
             return true;
         }
-        if (block == Blocks.FARMLAND || block == Blocks.GRASS_PATH) {
-            return true;
-        }
-        if (block == Blocks.ENDER_CHEST || block == Blocks.CHEST || block == Blocks.TRAPPED_CHEST) {
+        if (block == Blocks.FARMLAND || block == Blocks.GRASS_PATH ||
+            block == Blocks.ENDER_CHEST || block == Blocks.CHEST || block == Blocks.TRAPPED_CHEST ||
+            block == Blocks.END_PORTAL_FRAME || block == Blocks.ENCHANTING_TABLE) {
```
leijurv, line 304: Why end portal frame and enchanting table?

src/main/java/baritone/process/MineProcess.java
```diff
@@ -421,7 +421,7 @@ private void addNearby() {
     }
 
     public static boolean plausibleToBreak(CalculationContext ctx, BlockPos pos) {
-        if (MovementHelper.getMiningDurationTicks(ctx, pos.getX(), pos.getY(), pos.getZ(), ctx.bsi.get0(pos), true) >= COST_INF) {
+        if (MovementHelper.avoidBreaking(ctx.bsi, pos.getX(), pos.getY(), pos.getZ(),  ctx.bsi.get0(pos))) {
```
leijurv, line 424: Why change this?

src/main/java/baritone/pathing/movement/MovementHelper.java
dominikaaaa: I asked that already, it's in the original list somewhere??

src/main/java/baritone/process/MineProcess.java
CDAGaming: See the prior review convo where @wagyourtail explained why.

leijurv: I did not see any such explanation? `that whole function is only used for pathing goals and removing that allows pathing to "unbreakable" blocks ie. end portal frame.` This isn't a reason. If the mining duration is above the maximum cost, it is not plausible to break. What bug does this fix / why make this change?

wagyourtail: #goto end_portal_frame

leijurv: Your PR also makes end portal frame "can walk through" which is incorrect. That could cause untold issues and stuck paths.

wagyourtail: canWalkThrough on closer inspection appeared to me to be a function that should be named "canWalkOn" as that appears to be what it does else why are chests and Grass_path true

src/main/java/baritone/pathing/movement/MovementHelper.java
wagyourtail: blocks that are non-full appear to need to be in `canWalkOn` for baritone to be able to walk on them. If I'm mistaken about the purpose of this function, this change isn't actually necessary for the other stuff to work.

wagyourtail:
I have to revisit this, this is bad

PR #1963, opened 2020-08-19
Adds a FullBright Command

scorbett123:
This adds a command that turns the gamma up so you can see during mining, and get back from mining as baritone doesn't place torches.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1110,6 +1110,11 @@ public final class Settings {
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
     /**
+     *
+     */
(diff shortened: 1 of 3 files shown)


RealIndrit:
This doesn’t fix any existing problem and is only useless code tbh, just use waypoints and path to it to get back home after mining... You don’t need to go home yourself... Also gamma won’t fix torch problem, because the problem is not the light, it’s the monster spawning... I am pretty sure this won’t be merged due to the reasons I just listed, if you want light, look into making it place torches instead, and if you use an utility client with this, just enable full bright in it externally

Tragafantas:
Or just use #tunnel, I think it places torches

dominikaaaa:
Placing torches is a different suggestion, and I think this is out of scope for Baritone.

PR #1970, opened 2020-08-23
Fix a Rendering Conflict with some other Mods

CDAGaming:
In testing within Fabritone, it was found that these two lines (Which were previously used for an unknown optifine workaround) are no longer needed, and had no known side effects as of writing.

This can be backported as needed, but details are unknown how far this can go down

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@
  */
 
 group 'baritone'
-version '1.6.2'
+version '1.6.1'
 
 buildscript {
     repositories {


leijurv:
@ZeroMemes do you remember why this is here?

dominikaaaa:
> do you remember why this is here 

# :joy_cat: 

top tier GL meme

CDAGaming:
> cda has regressed to his old ways of `[Fix]` despite having agreed to the contrary

Hey, it's hard to get used to, I'm slowly adopting it though xD
But yeah, those 2 statements for the lighting did nothing for the rendering, so I really wanna know when/why it was ever a thing, as potentially this is unneeded code in multiple branches.

leijurv:
See #752 and #1129

dominikaaaa:
please can we finally remove this it fucks up external baritone + impact compatibility 🥺🥺🥺

CDAGaming:
Yea, and tbf in 1.16, it has no discernable effect over any portion of optifine (Both Forge and with optifabric uneffected by this change). It might have been a thing in the past, but yea this is unused with no purpose in 1.16.x at least unless someone mentions otherwise.

CDAGaming:
Bumped. Surprised leijurv made a release without this...

PR #1978, opened 2020-08-26
✨ Add improved ETA and a command to access it

### LINKED ISSUE #16: Smarter ETA for full path
Currently the API can give you the ticks remaining in the current path segment fairly accurately, by summing up the freshly recalculated costs of the remaining movements. 

This could be easily extended to the planned next path.

However, for longer distance paths this doesn't work, as the segments aren't calculated yet. How can we give an ETA of the entire thing, from start to goal, even if it isn't calculated yet?

I'm considering something like: `goalHeuristic(currentPos) * ticksElapsedSoFar / (goalHeuristic(startPos) - goalHeuristic(currentPos))`

### LINKED ISSUE #1480: ETA Command
## Describe your suggestion
When using Baritone, I quite often find myself like a obnoxious 7 year old trying to ask baritone _"ARE WE THERE YET???"_
To fix this, a simple `#eta` command would be very useful. It would show the remaining time in seconds, or if you explicitly set it (see below), it would show in ticks.

## Settings
There is not much to add to this. Maybe an option like `showEtaInTicks` would be useful, but otherwise there wouldn't be much.

## Context
My suggestion would improve baritone because (at least I think) there is an API method for finding the ETA, but it currently does not have a command.

## Final checklist

ZacSharp:
Closes #16 (full-path eta)
Closes #1480 (`#eta`)

### DIFF
--- a/src/api/java/baritone/api/behavior/IPathingBehavior.java
+++ b/src/api/java/baritone/api/behavior/IPathingBehavior.java
@@ -59,6 +59,15 @@ public interface IPathingBehavior extends IBehavior {
     }
 
     /**
+     * Returns the estimated remaining ticks to the current goal.
+     * Given that the return type is an optional, {@link Optional#empty()}
+     * will be returned in the case that there is no current goal.
+     *
+     * @return The estimated remaining ticks to the current goal.
+     */
+    Optional<Double> estimatedTicksToGoal();
+
+    /**
      * @return The current pathing goal
      */
     Goal getGoal();
--- a/src/api/java/baritone/api/pathing/goals/GoalComposite.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalComposite.java
@@ -61,7 +61,7 @@ public class GoalComposite implements Goal {
     public double heuristic() {
         double min = Double.MAX_VALUE;
         for (Goal g : goals) {
-            //just take the highest value that is guaranteed to be inside the goal
+            // just take the highest value that is guaranteed to be inside the goal
             min = Math.min(min, g.heuristic());
         }
         return min;
--- a/src/api/java/baritone/api/pathing/goals/GoalComposite.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalComposite.java
@@ -58,6 +58,16 @@ public class GoalComposite implements Goal {
     }
 
     @Override
+    public double heuristic() {
+        double min = Double.MAX_VALUE;
+        for (Goal g: goals) {
+            //just take the highest value that is guaranteed to be inside the goal
+            min = Math.min(min, g.heuristic());
+        }
+        return min;
+    }
+
+    @Override
     public String toString() {
         return "GoalComposite" + Arrays.toString(goals);
     }
--- a/src/api/java/baritone/api/pathing/goals/GoalComposite.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalComposite.java
@@ -60,7 +60,7 @@ public class GoalComposite implements Goal {
     @Override
     public double heuristic() {
         double min = Double.MAX_VALUE;
-        for (Goal g: goals) {
+        for (Goal g : goals) {
             //just take the highest value that is guaranteed to be inside the goal
             min = Math.min(min, g.heuristic());
         }
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -71,7 +71,13 @@ public class GoalNear implements Goal, IGoalRenderPos {
                 }
             }
         }
-        return Collections.max(maybeAlwaysInside);
+        double maxInside = Double.NEGATIVE_INFINITY;
+        for (double inside : maybeAlwaysInside) {
+            if (inside < minOutside) {
+                maxInside = Math.max(maxInside, inside);
+            }
+        }
+        return maxInside;
     }
 
     @Override
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -56,14 +56,14 @@ public class GoalNear implements Goal, IGoalRenderPos {
 
     @Override
     public double heuristic() {//TODO less hacky solution
-        int range = (int)Math.ceil(Math.sqrt(rangeSq));
+        int range = (int) Math.ceil(Math.sqrt(rangeSq));
         HashSet<Double> maybeAlwaysInside = new HashSet<>();
         HashSet<Double> sometimesOutside = new HashSet<>();
         for (int dx = -range; dx <= range; dx++) {
             for (int dy = -range; dy <= range; dy++) {
                 for (int dz = -range; dz <= range; dz++) {
-                    double h = heuristic(x+dx, y+dy, z+dz);
-                    if (!sometimesOutside.contains(h) && isInGoal(x+dx, y+dy, z+dz)) {
+                    double h = heuristic(x + dx, y + dy, z + dz);
+                    if (!sometimesOutside.contains(h) && isInGoal(x + dx, y + dy, z + dz)) {
                         maybeAlwaysInside.add(h);
                     } else {
                         maybeAlwaysInside.remove(h);
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -56,7 +56,7 @@ public class GoalNear implements Goal, IGoalRenderPos {
 
     @Override
     public double heuristic() {//TODO less hacky solution
-        int range = (int)Math.ceil(Math.abs(Math.sqrt(rangeSq)));
+        int range = (int)Math.ceil(Math.sqrt(rangeSq));
         HashSet<Double> maybeAlwaysInside = new HashSet<>();
         HashSet<Double> sometimesOutside = new HashSet<>();
         for (int dx = -range; dx <= range; dx++) {
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -55,7 +55,7 @@ public class GoalNear implements Goal, IGoalRenderPos {
     }
 
     @Override
-    public double heuristic() {//TODO less hacky solution
+    public double heuristic() {// TODO less hacky solution
         int range = (int) Math.ceil(Math.sqrt(rangeSq));
         HashSet<Double> maybeAlwaysInside = new HashSet<>();
         HashSet<Double> sometimesOutside = new HashSet<>();
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -21,6 +21,9 @@ import baritone.api.utils.SettingsUtil;
 import baritone.api.utils.interfaces.IGoalRenderPos;
 import net.minecraft.util.math.BlockPos;
 
+import java.util.Collections;
+import java.util.HashSet;
+
 public class GoalNear implements Goal, IGoalRenderPos {
 
     private final int x;
@@ -49,6 +52,27 @@ public class GoalNear implements Goal, IGoalRenderPos {
         int yDiff = y - this.y;
         int zDiff = z - this.z;
         return GoalBlock.calculate(xDiff, yDiff, zDiff);
+    }
+
+    @Override
+    public double heuristic() {//TODO less hacky solution
+        int range = (int)Math.ceil(Math.abs(Math.sqrt(rangeSq)));
+        HashSet<Double> maybeAlwaysInside = new HashSet<>();
+        HashSet<Double> sometimesOutside = new HashSet<>();
+        for (int dx = -range; dx <= range; dx++) {
+            for (int dy = -range; dy <= range; dy++) {
+                for (int dz = -range; dz <= range; dz++) {
+                    double h = heuristic(x+dx, y+dy, z+dz);
+                    if (!sometimesOutside.contains(h) && isInGoal(x+dx, y+dy, z+dz)) {
+                        maybeAlwaysInside.add(h);
+                    } else {
+                        maybeAlwaysInside.remove(h);
+                        sometimesOutside.add(h);
+                    }
+                }
+            }
+        }
+        return Collections.max(maybeAlwaysInside);
     }
 
     @Override
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -56,7 +56,7 @@ public class GoalNear implements Goal, IGoalRenderPos {
     @Override
     public double heuristic() {// TODO less hacky solution
         int range = (int) Math.ceil(Math.sqrt(rangeSq));
-        DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet();
+        DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet(); // see pull request #1978
         double minOutside = Double.POSITIVE_INFINITY;
         for (int dx = -range; dx <= range; dx++) {
             for (int dy = -range; dy <= range; dy++) {
@@ -72,7 +72,7 @@ public class GoalNear implements Goal, IGoalRenderPos {
         }
         double maxInside = Double.NEGATIVE_INFINITY;
         DoubleIterator it = maybeAlwaysInside.iterator();
-        while(it.hasNext()) {
+        while (it.hasNext()) {
             double inside = it.nextDouble();
             if (inside < minOutside) {
                 maxInside = Math.max(maxInside, inside);
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -58,16 +58,15 @@ public class GoalNear implements Goal, IGoalRenderPos {
     public double heuristic() {// TODO less hacky solution
         int range = (int) Math.ceil(Math.sqrt(rangeSq));
         HashSet<Double> maybeAlwaysInside = new HashSet<>();
-        HashSet<Double> sometimesOutside = new HashSet<>();
+        double minOutside = Double.POSITIVE_INFINITY;
         for (int dx = -range; dx <= range; dx++) {
             for (int dy = -range; dy <= range; dy++) {
                 for (int dz = -range; dz <= range; dz++) {
                     double h = heuristic(x + dx, y + dy, z + dz);
-                    if (!sometimesOutside.contains(h) && isInGoal(x + dx, y + dy, z + dz)) {
+                    if (h < minOutside && isInGoal(x + dx, y + dy, z + dz)) {
                         maybeAlwaysInside.add(h);
                     } else {
-                        maybeAlwaysInside.remove(h);
-                        sometimesOutside.add(h);
+                        minOutside = Math.min(minOutside, h);
                     }
                 }
             }
--- a/src/api/java/baritone/api/pathing/goals/GoalNear.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalNear.java
@@ -19,10 +19,9 @@ package baritone.api.pathing.goals;
 
 import baritone.api.utils.SettingsUtil;
 import baritone.api.utils.interfaces.IGoalRenderPos;
+import it.unimi.dsi.fastutil.doubles.DoubleOpenHashSet;
+import it.unimi.dsi.fastutil.doubles.DoubleIterator;
 import net.minecraft.util.math.BlockPos;
-
-import java.util.Collections;
-import java.util.HashSet;
 
 public class GoalNear implements Goal, IGoalRenderPos {
 
@@ -57,7 +56,7 @@ public class GoalNear implements Goal, IGoalRenderPos {
     @Override
     public double heuristic() {// TODO less hacky solution
         int range = (int) Math.ceil(Math.sqrt(rangeSq));
-        HashSet<Double> maybeAlwaysInside = new HashSet<>();
+        DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet();
         double minOutside = Double.POSITIVE_INFINITY;
         for (int dx = -range; dx <= range; dx++) {
             for (int dy = -range; dy <= range; dy++) {
@@ -72,7 +71,9 @@ public class GoalNear implements Goal, IGoalRenderPos {
             }
         }
         double maxInside = Double.NEGATIVE_INFINITY;
-        for (double inside : maybeAlwaysInside) {
+        DoubleIterator it = maybeAlwaysInside.iterator();
+        while(it.hasNext()) {
+            double inside = it.nextDouble();
             if (inside < minOutside) {
                 maxInside = Math.max(maxInside, inside);
             }
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -21,6 +21,9 @@ import baritone.api.utils.SettingsUtil;
 import net.minecraft.util.math.BlockPos;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.NoSuchElementException;
 
 /**
  * Useful for automated combat (retreating specifically)
@@ -78,6 +81,45 @@ public class GoalRunAway implements Goal {
             min = min * 0.6 + GoalYLevel.calculate(maintainY, y) * 1.5;
         }
         return min;
+    }
+
+    @Override
+    public double heuristic() {//TODO less hacky solution
+        int distance = (int)Math.ceil(Math.abs(Math.sqrt(distanceSq)));
+        int minX = from[0].getX() - distance;
+        int minY = from[0].getY() - distance;
+        int minZ = from[0].getZ() - distance;
+        int maxX = from[0].getX() + distance;
+        int maxY = from[0].getY() + distance;
+        int maxZ = from[0].getZ() + distance;
+        for (BlockPos p : from) {
+            minX = Math.min(minX, p.getX() - distance);
+            minY = Math.min(minY, p.getY() - distance);
+            minZ = Math.min(minZ, p.getZ() - distance);
+            maxX = Math.max(minX, p.getX() + distance);
+            maxY = Math.max(minY, p.getY() + distance);
+            maxZ = Math.max(minZ, p.getZ() + distance);
+        }
+        HashSet<Double> maybeAlwaysInside = new HashSet<>();
+        HashSet<Double> sometimesOutside = new HashSet<>();
+        for (int x = minX; x <= maxX; x++) {
+            for (int y = minX; y <= maxX; y++) {
+                for (int z = minX; z <= maxX; z++) {
+                    double h = heuristic(x, y, z);
+                    if (!sometimesOutside.contains(h) && isInGoal(x, y, z)) {
+                        maybeAlwaysInside.add(h);
+                    } else {
+                        maybeAlwaysInside.remove(h);
+                        sometimesOutside.add(h);
+                    }
+                }
+            }
+        }
+        try {
+            return Collections.max(maybeAlwaysInside);
+        } catch (NoSuchElementException e) {
+            return Double.NEGATIVE_INFINITY;
+        }
     }
 
     @Override
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -85,12 +85,12 @@ public class GoalRunAway implements Goal {
     @Override
     public double heuristic() {// TODO less hacky solution
         int distance = (int) Math.ceil(Math.sqrt(distanceSq));
-        int minX = from[0].getX() - distance;
-        int minY = from[0].getY() - distance;
-        int minZ = from[0].getZ() - distance;
-        int maxX = from[0].getX() + distance;
-        int maxY = from[0].getY() + distance;
-        int maxZ = from[0].getZ() + distance;
+        int minX = Integer.MAX_VALUE;
+        int minY = Integer.MAX_VALUE;
+        int minZ = Integer.MAX_VALUE;
+        int maxX = Integer.MIN_VALUE;
+        int maxY = Integer.MIN_VALUE;
+        int maxZ = Integer.MIN_VALUE;
         for (BlockPos p : from) {
             minX = Math.min(minX, p.getX() - distance);
             minY = Math.min(minY, p.getY() - distance);
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -85,7 +85,7 @@ public class GoalRunAway implements Goal {
 
     @Override
     public double heuristic() {//TODO less hacky solution
-        int distance = (int)Math.ceil(Math.abs(Math.sqrt(distanceSq)));
+        int distance = (int)Math.ceil(Math.sqrt(distanceSq));
         int minX = from[0].getX() - distance;
         int minY = from[0].getY() - distance;
         int minZ = from[0].getZ() - distance;
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -103,8 +103,8 @@ public class GoalRunAway implements Goal {
         HashSet<Double> maybeAlwaysInside = new HashSet<>();
         HashSet<Double> sometimesOutside = new HashSet<>();
         for (int x = minX; x <= maxX; x++) {
-            for (int y = minX; y <= maxX; y++) {
-                for (int z = minX; z <= maxX; z++) {
+            for (int y = minY; y <= maxY; y++) {
+                for (int z = minZ; z <= maxZ; z++) {
                     double h = heuristic(x, y, z);
                     if (!sometimesOutside.contains(h) && isInGoal(x, y, z)) {
                         maybeAlwaysInside.add(h);
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -68,7 +68,7 @@ public class GoalRunAway implements Goal {
     }
 
     @Override
-    public double heuristic(int x, int y, int z) {//mostly copied from GoalBlock
+    public double heuristic(int x, int y, int z) {// mostly copied from GoalBlock
         double min = Double.MAX_VALUE;
         for (BlockPos p : from) {
             double h = GoalXZ.calculate(p.getX() - x, p.getZ() - z);
@@ -84,7 +84,7 @@ public class GoalRunAway implements Goal {
     }
 
     @Override
-    public double heuristic() {//TODO less hacky solution
+    public double heuristic() {// TODO less hacky solution
         int distance = (int) Math.ceil(Math.sqrt(distanceSq));
         int minX = from[0].getX() - distance;
         int minY = from[0].getY() - distance;
--- a/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalRunAway.java
@@ -99,7 +99,7 @@ public class GoalRunAway implements Goal {
             maxY = Math.max(minY, p.getY() + distance);
             maxZ = Math.max(minZ, p.getZ() + distance);
         }
-        DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet();
+        DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet(); // see pull request #1978
         double minOutside = Double.POSITIVE_INFINITY;
         for (int x = minX; x <= maxX; x++) {
             for (int y = minY; y <= maxY; y++) {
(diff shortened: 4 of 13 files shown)


src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -372,6 +377,16 @@ public CalculationContext secretInternalGetCalculationContext() {
...
 
+    public Optional<Double> estimatedTicksToGoal(){
+        if (goal == null){
+            return Optional.empty();
+        }
+        BetterBlockPos currentPos = ctx.playerFeet();
+        double current = goal.heuristic(currentPos.x, currentPos.y, currentPos.z);
+        double start = goal.heuristic(startPosition.x, startPosition.y, startPosition.z);
+        return Optional.of(current * ticksElapsedSoFar / (start - current));
+    }
```
dominikaaaa, line 388: formatting

ZacSharp: Can you be more specific on what I should change please?

dominikaaaa:
```diff
+    public Optional<Double> estimatedTicksToGoal() {
-    public Optional<Double> estimatedTicksToGoal(){
+        if (goal == null) {
-        if (goal == null){
=            return Optional.empty();
=        }
+
=        BetterBlockPos currentPos = ctx.playerFeet();
=        double current = goal.heuristic(currentPos.x, currentPos.y, currentPos.z);
=        double start = goal.heuristic(startPosition.x, startPosition.y, startPosition.z);
=        return Optional.of(current * ticksElapsedSoFar / (start - current));
=    }
```

ZacSharp: Well if that's an issue I'll fix it ofc

src/main/java/baritone/command/defaults/EtaCommand.java
```diff
@@ -0,0 +1,78 @@
...
+import baritone.api.command.Command;
+import baritone.api.command.exception.CommandException;
+import baritone.api.command.exception.CommandInvalidStateException;
+import baritone.api.command.argument.IArgConsumer;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+public class EtaCommand extends Command {
```
leijurv, line 33: ETA instead of Eta please

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -372,6 +377,16 @@ public CalculationContext secretInternalGetCalculationContext() {
...
     }
 
+    public Optional<Double> estimatedTicksToGoal() {
+        if (goal == null) {
+            return Optional.empty();
+        }
+        BetterBlockPos currentPos = ctx.playerFeet();
+        double current = goal.heuristic(currentPos.x, currentPos.y, currentPos.z);
+        double start = goal.heuristic(startPosition.x, startPosition.y, startPosition.z);
+        return Optional.of(current * ticksElapsedSoFar / (start - current));
```
leijurv, line 387: Can this cause an issue by dividing by zero? If I start a path then immediately do ETA, `start - current` would be zero. What happens then?

src/main/java/baritone/command/defaults/EtaCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public Stream<String> tabComplete(String label, IArgConsumer args) {
+        return Stream.empty();
+    }
+
+    @Override
+    public String getShortDesc() {
+        return "View the current eta";
```
leijurv, line 63: Capital ETA everywhere here

src/main/java/baritone/command/defaults/EtaCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public List<String> getLongDesc() {
+        return Arrays.asList(
+                "The eta command provides information about the estimated time until the next segment.",
+                "and the goal",
+                "",
+                "Be aware that the eta to your goal is really unprecise",
+                "",
+                "Usage:",
+                "> proc - View eta, if present"
```
leijurv, line 75: proc? Is this a copy/paste mistake?

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -468,6 +483,8 @@ private void findPathInNewThread(final BlockPos start, final boolean talkAboutIt
                         if (executor.get().getPath().positions().contains(expectedSegmentStart)) {
                             queuePathEvent(PathEvent.CALC_FINISHED_NOW_EXECUTING);
                             current = executor.get();
+                            ticksElapsedSoFar = 0;
```
leijurv, line 486: Perhaps these should be moved to a function like "resetEstimate", and maybe it should also be called when the goal changes? I'm not sure though, that isn't really necessary / might be too much work.

src/api/java/baritone/api/behavior/IPathingBehavior.java
```diff
@@ -58,6 +58,15 @@
...
     }
 
+    /**
+     * Returns the estimated remaining ticks to the current goal.
+     * Given that the return type is an optional, {@link Optional#empty()}
+     * will be returned in the case that there is no current goal.
+     *
+     * @return The estimated remaining ticks to the current goal.
+     */
+    Optional<Double> estimatedTicksToGoal();
```
leijurv, line 68: Normally I'd suggest `OptionalDouble` but this is probably fine just for consistency / API compatibility.

src/main/java/baritone/command/defaults/EtaCommand.java
ZacSharp: Yup, it is.

src/main/java/baritone/behavior/PathingBehavior.java
dominikaaaa: user error 😂

src/api/java/baritone/api/behavior/IPathingBehavior.java
ZacSharp: If it was an alias for `Optional<Double>` i'd take it but I think all optional return values should be a `Optional` instance
(null is a different thing)

src/main/java/baritone/behavior/PathingBehavior.java
ZacSharp: No, then should return `Optional.empty()` because there is not ETA

ZacSharp: Also this might happen with clients calling this every tick (e.g. for a hud)

ZacSharp: Should it return `Optional.empty()` whenever the path has just started or only if `start - current == 0`?

ZacSharp: Can Baritone be following a path while calculating a primary segment? Because while thinking about this I noticed that I expected `estimatedTicksToGoal` to only be called when `startPosition` and `ticksElapsedSoFar` are both set for the same path, but the former is set before pathing and the latter after pathing.

src/main/java/baritone/behavior/PathingBehavior.java
leijurv: lmk if you think this is reasonable, or too much work for not much benefit

dominikaaaa: It would be nice if it was called again if the goal changes

ZacSharp: Yes but only if we start a full new path for that goal (e.g. not when the `GoalBlock` to `GoalXZ` simplification ends or `MineProcess` removes an ore). Would be nice if you tell me where to put the reset(s) to achieve that behaviour.

leijurv: yea i have no idea lol

ZacSharp:
This behaves really weird with `MineProcess`. I got a ETA of -61

dominikaaaa:
> I got a ETA of -61

fix:
```java
if (eta.toString.startsWith("-") {
    eta = eta * -1
}
```

RealIndrit:
Nice troll, above here we see kami clients way of solving problems 😎

leijurv:
absolutely fucking based comment from indrit

ZacSharp:
The problem is the player being further away from our goal than `StartPosition`, can we do anything about this (maybe using `Math.abs(start-current)` instead of `start-current`)?
Also `#tunnel` has a decreasing negative ETA, is that how it should be?

ZacSharp:
> maybe using `Math.abs(start-current)` instead of `start-current`

lol, meant that as a joke but it is actually correct :smile:.
The correct formula is `Math.abs(current-goal)*ticksElapsedSoFar/Math.abs(start-current)` (goal is the heuristic at the goal)
Is it worth it to add `heuristic()` with no args to `Goal` so eta works with goals that do not end with 0?

dominikaaaa:
> with goals that do not end in 0

yes.

ZacSharp:
Ok, doing it. If you want to help me, tell me what to return for `GoalComposite`, `GoalNear` and `GoalRunAway`, the rest is trivial

dominikaaaa:
`GoalRunAway` should be "No ETA, Running Away". Not sure what the other two are

ZacSharp:
> `GoalRunAway` should be "No ETA, Running Away". Not sure what the other two are

Actually not because first the command to access the ETA does not know about Baritone running away (it just receives a `Optional.empty()`) and second `GoalRunAway` is actually reachable (all positions to run away from are far away enough) so we should return an estimate time until that is the case

ZacSharp:
Ok, I'm stuck on finding a formula for the highest possible heuristic that can't occur for positions where `isInGoal` returns `false` for `GoalNear` and `GoalRunAway`. I've got a quick and dirty solution and if nobody (including me) comes up with something better in the next week I'll push it. Please help

src/main/java/baritone/behavior/PathingBehavior.java
ZacSharp: Is it an issue that this happens in a separate thread?

src/api/java/baritone/api/pathing/goals/GoalComposite.java
```diff
@@ -57,6 +57,16 @@ public double heuristic(int x, int y, int z) {
         return min;
     }
 
+    @Override
+    public double heuristic() {
+        double min = Double.MAX_VALUE;
+        for (Goal g: goals) {
```
leijurv, line 63: Plz run through an autoformatter to baritone style

src/api/java/baritone/api/pathing/goals/GoalNear.java
```diff
@@ -51,6 +54,27 @@ public double heuristic(int x, int y, int z) {
         return GoalBlock.calculate(xDiff, yDiff, zDiff);
     }
 
+    @Override
+    public double heuristic() {//TODO less hacky solution
```
leijurv, line 58: what the fuck is this

src/api/java/baritone/api/pathing/goals/GoalNear.java
```diff
@@ -51,6 +54,27 @@ public double heuristic(int x, int y, int z) {
...
+        int range = (int)Math.ceil(Math.abs(Math.sqrt(rangeSq)));
```
leijurv, line 59: Why are you abs-ing a sqrt???

src/api/java/baritone/api/pathing/goals/GoalNear.java
```diff
@@ -51,6 +54,27 @@ public double heuristic(int x, int y, int z) {
...
+        HashSet<Double> maybeAlwaysInside = new HashSet<>();
```
leijurv, line 60: what the FUCK

src/api/java/baritone/api/pathing/goals/GoalRunAway.java
```diff
@@ -80,6 +83,45 @@ public double heuristic(int x, int y, int z) {//mostly copied from GoalBlock
...
+                    } else {
+                        maybeAlwaysInside.remove(h);
+                        sometimesOutside.add(h);
+                    }
+                }
+            }
+        }
+        try {
+            return Collections.max(maybeAlwaysInside);
+        } catch (NoSuchElementException e) {
```
leijurv, line 120: WHAT

CDAGaming: THE

leijurv:
The hacky goals don't need hacky ETAs. `#tunnel` has no ETA, don't worry about it. Same for RunAway. There is no actual goal, they just go forever.

src/api/java/baritone/api/pathing/goals/GoalRunAway.java
l1ving: FUCK

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: It finds the highest heuristic that can not occur outside the goal.

leijurv: you want an ordered set then, where you can do queries based on what's greater than or less than a certain key??

leijurv: a heap, specifically??

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: true point
guess I didn't notice that the value I get is already guaranteed to be positive

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: Since I dunno what a heap is: maybe

src/api/java/baritone/api/pathing/goals/GoalComposite.java
ZacSharp: Any recommendations for a linux command line formatter that can conform to baritone style?
And if not, maybe some info on what "baritone style" is?

leijurv: uh it's basically the intellij autoformatter? iirc theres very few changes and you won't run into any of them. you can test if you have it right by doing reformat on all the code and checking if `git` says any files have been changed

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: I'll use a `TreeSet` because it has a better performance for add, remove and contains than `PriorityQueue` which is the only heap implementation in java (No i won't do my own for this)

ZacSharp: Wait, when comparing the performance for the most often used operations (add, contains, remove), `HasSet` (O(1), O(1), O(1)) is significantly better than `TreeSet` (O(log(n)), O(log(n)), O(log(n))) and `PriorityQueue` (O(log(n)), O(n), O(n)). I don't think the single access to the biggest element in the end changes much about that ordering.

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: What's the problem here?

src/api/java/baritone/api/pathing/goals/GoalRunAway.java
ZacSharp: What's the problem here?

src/api/java/baritone/api/pathing/goals/GoalNear.java
leijurv: I'm not saying that you could improve on the big O complexity of a HashSet, I'm implying that the actual behavior of a Set is almost certainly not what you want here and I can't imagine how it could be. `It finds the highest heuristic that can not occur outside the goal.` You're comparing by == on these double values. That is a horrible smell and I've literally never seen a case where a hash keyed on a float/double was actually correct. "the highest heuristic that cannot occur outside the goal" you should be doing queries based on `>` instead of `==`. Because of floating point inaccuracy you can't guarantee that the heuristic calculated at dx+1 dz-1 will be equal for example. At the very minimum you need an epsilon value, and even then I wouldn't trust it.

What exactly are you trying to accomplish? Can you explain more in depth what it means to get heuristics that can't occur outside the goal??

src/api/java/baritone/api/pathing/goals/GoalRunAway.java
0-x-2-2: LOL>

src/api/java/baritone/api/pathing/goals/GoalRunAway.java
```diff
@@ -80,6 +83,45 @@ public double heuristic(int x, int y, int z) {//mostly copied from GoalBlock
         return min;
     }
 
+    @Override
+    public double heuristic() {//TODO less hacky solution
```
CDAGaming, line 87: Space after //

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -372,6 +377,40 @@ public CalculationContext secretInternalGetCalculationContext() {
...
+        if (goal.isInGoal(ctx.playerFeet())) {
+            resetEstimatedTicksToGoal();
+            return Optional.of(0.0);
+        }
+        if (ticksElapsedSoFar == 0) {
+            return Optional.empty();
+        }
+        double current = goal.heuristic(currentPos.x, currentPos.y, currentPos.z);
+        double start = goal.heuristic(startPosition.x, startPosition.y, startPosition.z);
+        if (current == start) {//can't check above because current and start can be equal even if currentPos and startPosition are not
```
CDAGaming, line 394: space after //

CDAGaming (changes requested):
Looks good on formatting, fix the other comments and this'll look good

src/api/java/baritone/api/pathing/goals/GoalNear.java
ZacSharp: The full-path ETA needs the heuristic at the goal (even if that is hypothetical like for `#tunnel`) and if there are multiple possible ones I want to take the highest in order to get the lowest ETA. At the same time I need to make sure that the chosen heuristic can not occur outside the goal, so the ETA can not reach 0 before the goal is reached.
So "h can not occur outside the goal" basically means that for all combinations of x, y, z `heurisict(x,y,z) != h || isInGoal(x,y,z)` is true.
And yes, as of writing this I am thinking about whether that condition actually fits the purpose I explained above.

ZacSharp: And shouldn't floating point accuracy be deterministic so we can just ignore it here?

ZacSharp:
That commit message

l1ving:
I highly recommend using whatcommit https://github.com/l1ving/what-commit/ to generate commit messages for you

ZacSharp:
At least the grammar is better than mine.
I won't judge the content though.

src/api/java/baritone/api/pathing/goals/GoalNear.java
leijurv: "ETA cannot reach zero" again, you're using ≠ when you want >

A negative ETA is just as bad as a zero ETA? Why do you only care about eliminating 0 ETA and not - ETA?

"h cannot occur outside the goal" should actually be "h or lower cannot occur outside the goal".

`heurisict(x,y,z) != h || isInGoal(x,y,z)` should instead be `heurisict(x,y,z) > h || isInGoal(x,y,z)`

ZacSharp: That's what the last sentence was about. The current solution allows for e.g. 5 being returned although e.g. 3 is a possible heuristic outside the goal. In addition to that it fails when there is no in-goal-only heuristic.
I guess I will change this to start with `h = Double.MAX` and then do `if (!isInGoal(x, y, z)) h = Math.min(h, heuristic(x, y, z)` for each position inside the box. That way it always has a value that is guaranteed to be lower than all not-in-goal heuristics.

ZacSharp: We could also see it as "The heuristic says you could already be there. Why are you wasting time?" because when the heuristic is x at a certain point inside the we should not need to get closer to the goal than x.
I'll still change it because the new solution looks nicer and has better performance which is actually an issue for `#follow` with high followRange (100 or so, not really likely to happen).

l1ving:
Is there a new status on this? Are you working on it? I'd really appreciate such a feature

ZacSharp:
No new status here. I'd like to finish this as well and the only thing that I know of that needs to be resolved is the ordered set / HashSet thing and maybe changes you requested in your highly informative "WHAT THE FUCK" comments. I'm currently not working on it because I don't know what to do.
I know leiurv wants me to use an ordered set, but I don't understand why it should be better there (from my perspective it's simply less efficient and has messier code).

leijurv:
No matter how much you complain about the "what the fuck" it doesn't change the fact that a `HashSet<Double>` is **never** correct and I would **never** merge one. Floating point inaccuracy means that such a set is always indicative of bad code that will fail with the slightest numerical instability.

And, as I tried to explain here https://github.com/cabaletta/baritone/pull/1978#discussion_r505811985 the HashSet is built on top of a faulty assumption about heuristics being equal. The reason why you have a HashSet is (honestly I can't even remember I'm just rereading what I wrote) something about heuristics being `!=` to each other, while really you want `>`.

ZacSharp:
The thing about `==` and `<` is fixed and if I didn't miss anything, I don't rely on values being equal anymore. I only put in items in and don't remove them again, so no chance the second value is slightly off (it simply doesn't exist). The HashSet might have been based on the assumption of heuristics being equal, but that isn't the case any more and I'm leaving it a HashSet because that's the most efficient Collection I could find.

And while writing this I noticed that I need to change it to return the biggest entry below minOutside, not the biggest entry. That means I probably need more that the one biggest value, so I'm not sure if the better efficiency of an OrderedSet at the end might outweight the efficiency of a HashSet in the first part.
I remember trying to use some sort of ordered set and that I got some pretty messy and dysfunctional code when trying to use the headSet method to reduce the set size inside the triple loop.

If the "don't use HasSet<Double>" rule is that strict, I could simply use an OrderedSet as if it was a HashSet, but that doesn't help anything.

leijurv:
> The thing about == and < is fixed and if I didn't miss anything, I don't rely on values being equal anymore. 
okay, let me take a look

leijurv:
Thanks, please ignore my last message that I wrote last night. I should have reread your diff beforehand.

It looks like you're just using the HashSet to build up a list of doubles, then find the maximum.

```
HashSet<Double> maybeAlwaysInside = new HashSet<>();
...
maybeAlwaysInside.add(h);
...
return Collections.max(maybeAlwaysInside);
```

and

```
HashSet<Double> maybeAlwaysInside = new HashSet<>();
...
maybeAlwaysInside.add(h);
...
try {
    return Collections.max(maybeAlwaysInside);
} catch (NoSuchElementException e) {
    return Double.NEGATIVE_INFINITY;
}
```

So, this usage of a collection is fine from a correctness point of view. My previous complaint about this resulting in incorrect behavior is no longer correct. Sorry for repeating it.

However, it still is inefficient and, more than anything, **very** confusing to the reader. Especially when it's right next to `minOutside` which is done the "normal way". Why would finding the minimum use a normal `double` and the maximum use a `Set<Double>`?

So why don't we just replace like this:

`HashSet<Double> maybeAlwaysInside = new HashSet<>();` => `double maxHeuristic = Double.NEGATIVE_INFINITY;`

`maybeAlwaysInside.add(h);` => `maxHeuristic = Math.max(maxHeuristic, h);`

```
try {
    return Collections.max(maybeAlwaysInside);
} catch (NoSuchElementException e) {
    return Double.NEGATIVE_INFINITY;
}
```

=> `return maxHeuristic;`

ZacSharp:
That's what the second part of my comment mentions. I did a little mistake there when removing the calls to `maybeAlwaysInside.remove`. It doesn't just need to return the biggest entry, but the biggest entry below `minOutside`.
I'll test again if I manage to use a TreeSet and if I do, if that is more efficient.

leijurv:
Wait, so you're trying to return the.... second to lowest entry? Or what

ZacSharp:
No, the second lowest entry (and any other fixed offset) would be possible without a collection.
The return value has to be below minOutside, which can be lowered during the iteration, so I need to keep track of all possible lower values, in case the values preceding them become too big.

I did the testing with a tree set (dunno how I managed to fuck it up that badly a few months ago)
radius\method | HashSet (as committed) | TreeSet | TreeSet with constant shrinking
----|----|----|----
3 (default followRadius)|2,0,0,0,0,0,0 ms|0,0,0,0,0,0,0 ms|7,1,1,0,0,1,0 ms
520 (biggest render distance)|20s|50s|3min

Times in ms are measured using `System.currentTimeMillis()` and times in s or min are counted using the ticking of a clock as a guide (did it before putting in the measuring code and didn't repeat it afterwards).

Since the approaches are almost equal with the default radius and HashSet is significantly faster for big radii I chose to keep the HashSet.

leijurv:
> The return value has to be below minOutside, which can be lowered during the iteration, so I need to keep track of all possible lower values, in case the values preceding them become too big.

Okay, I finally understand what you're trying to do.

You want to find the highest `heuristic` with `isInGoal` is true, that is still lower than the lowest `heuristic` with `isInGoal` false.

Before any talk of what data structure to use, why even use any data structure? Because you're walking through basically your entire possibility space twice (once to build up your list, and a second time to go through the list). I wonder if it's faster to just do the entire thing twice, because then you wouldn't need to hash every double into a set? Something like this:

```
double minOutside = Double.POSITIVE_INFINITY;
for (int x = minX; x <= maxX; x++) {
    for (int y = minY; y <= maxY; y++) {
        for (int z = minZ; z <= maxZ; z++) {
            if (!isInGoal(x, y, z)) {
                minOutside = Math.min(minOutside, heuristic(x, y, z));
            }
        }
    }
}
double maxInside = Double.NEGATIVE_INFINITY;
for (int x = minX; x <= maxX; x++) {
    for (int y = minY; y <= maxY; y++) {
        for (int z = minZ; z <= maxZ; z++) {
            if (isInGoal(x, y, z)) {
                double h = heuristic(x, y, z);
                if (h < minOutside) {
                    maxInside = Math.max(maxInside, h);
                }
            }
        }
    }
}
return maxInside;
```

I think there's a possibility that this ^ will be faster than what you currently have, and the reason is that while it does go through the space twice, so does the approach with `Collections.max` (more or less), and this approach actually calls `heuristic` the same amount of times as before (notice that in the first pass, it only does `heuristic` where `isInGoal` is true, and on the second pass it only does it when `isInGoal` is false).

But anyway, assuming that we still want to only do one pass through xyz space and build up a collection:

Using a set could improve performance because your final pass-through would hit less entries. However, I don't think this is a good idea overall because you're boxing every double into a Double without need.

There are more options to try:

`DoubleArrayList` seems pretty much what you want. It will allow for quite fast insertions and iteration because it's all primitives, no Doubles in sight.

```
DoubleArrayList maybeAlwaysInside = new DoubleArrayList();
blah blah your loop {
     blah blah
     maybeAlwaysInside.add(h);
}
maybeAlwaysInside.trim();
for (double inside : maybeAlwaysInside.elements()) {
    blah blah
}
```

(why use elements instead of a direct for each? to avoid allocating a Double)

There are other data structures you can use to speed this up. For example, `DoubleRBTreeSet` or `DoubleAVLTreeSet`. This will allow O(log n) fetching of the highest value lower than `minOutside`. I'm not sure if you did this already with your TreeSet implementation, but these variants from fastutil offer interesting functions like `.headSet(someVal)` / `.tailSet(someVal)`. For example, maybe something like `.headSet(minOutside).lastDouble()`? I can't guarantee it'll be faster though, and it's actually possible it'll be slower. The reason is that the insertion step becomes O(log n) each, for a total of O(n log n) for the initial loop. That might not be enough to overcome the improvement in the second stage from an O(n) loop to a O(log n) tree lookup. Another interesting idea that's only possible with red-black trees (or similar trees) like this is maybe something like:

```
the first loop {
    if (h < minOutside) {
        minOutside = h;
        rbTreeSet.tailSet(h).clear(); // remove everything from the tree that's greater than h
    }
}
return rbTreeSet.lastDouble(); // we can just return the last double because everything above minOutside has already been cleared
```

I'm even less certain that that would be performant though - the worst case pattern would result in a lot of useless `.clear()` calls.


And if all else fails (is slower than `HashSet<Double>`), try `DoubleOpenHashSet` (it'll behave the exact same as your `HashSet<Double>` but with no boxing alloc). 

Something like this maybe:

```
DoubleOpenHashSet maybeAlwaysInside = new DoubleOpenHashSet();
blah blah your loop {
     blah blah
     maybeAlwaysInside.add(h);
}
DoubleIterator it = maybeAlwaysInside.iterator();
while (it.hasNext()) {
    double inside = it.nextDouble();
    blah blah
}
```

(yet again, that weird iterator is to avoid allocating a Double. `Iterable<Double>` can be foreach'd normally, and is fine when the underlying collection is boxed, but it allocates when the underlying collection is of primitives, so in that case you want to use a `DoubleIterator` and manually traverse it)


I'm also curious of a simple `ArrayList<Double>` is faster or slower than the `HashSet<Double>`. (but that won't be the final solution: ArrayList<Double> is just a slower version of DoubleArrayList from fastutil, because it boxes into Double)

ZacSharp:
<details><summary>I did some testing</summary>and <b>if</b> I ever do something like this again I'll fully automate it, not partially.<table><thead><tr><th></th><th>HashSet</th><th>TreeSet</th><th>TreeSet (reassign)</th><th>TreeSet (clear)</th><th>ArrayList</th><th>ArrayList (ensureCapacity)</th><th>Double Iteration</th><th>DoubleArrayList</th><th>DoubleArrayList (ensureCapacity)</th><th>DoubleOpenHashSet</th><th>DoubleOpenHashSet (initial size)</th><th>HashSet (initial size)</th><th>DoubleRBTreeSet</th><th>DoubleRBTreeSet (reassign)</th><th>DoubleRBTreeSet (clear)</th><th>DoubleAVLTreeSet</th><th>DoubleAVLTreeSet (reassign)</th><th>DoubleAVLTreeSet (clear)</th><th>TreeSet (conditional reassign)</th><th>TreeSet (conditional clear)</th><th>DoubleRBTreeSet (conditional reassign)</th><th>DoubleRBTreeSet (conditional clear)</th><th>DoubleAVLTreeSet (conditional reassign)</th><th>DoubleAVLTreeSet (conditional clear)</th></tr></thead><tbody><tr><td rowspan="7">3</td><td>0</td><td>0</td><td>2</td><td>3</td><td>3</td><td>0</td><td>6</td><td>21</td><td>17</td><td>50</td><td>8</td><td>0</td><td>9</td><td>9</td><td>91</td><td>28</td><td>17</td><td>15</td><td>4</td><td>4</td><td>9</td><td>16</td><td>8</td><td>9</td></tr><tr><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>1</td><td>0</td><td>1</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td></tr><tr><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td></tr><tr><td>1</td><td>1</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td></tr><tr><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td></tr><tr><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td></tr><tr><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>1</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0</td></tr><tr><td rowspan="7">520</td><td>18682</td><td>53851</td><td>80200</td><td>251082</td><td colspan="2" rowspan="7">OutOfMemoryError</td><td>28021</td><td colspan="2" rowspan="7">OutOfMemoryError</td><td>16478</td><td rowspan="7">IllegalArgumentException</td><td rowspan="7">OutOfMemoryError</td><td>37529</td><td>47549</td><td>131762</td><td>38222</td><td>43799</td><td>127213</td><td>185716</td><td>73521</td><td>97567</td><td>41362</td><td>103994</td><td>37872</td></tr><tr><td>17919</td><td>54790</td><td>80858</td><td>249545</td><td>27851</td><td>16125</td><td>39761</td><td>47767</td><td>129309</td><td>40797</td><td>52397</td><td>128099</td><td>184166</td><td>73463</td><td>97608</td><td>41716</td><td>104755</td><td>38297</td></tr><tr><td>16539</td><td>52751</td><td>83211</td><td>249944</td><td>24931</td><td>13920</td><td>38125</td><td>43881</td><td>138165</td><td>40267</td><td>45959</td><td>123474</td><td>180619</td><td>71123</td><td>98466</td><td>36946</td><td>103635</td><td>34832</td></tr><tr><td>16624</td><td>51987</td><td>82351</td><td>254334</td><td>24960</td><td>13834</td><td>37960</td><td>43967</td><td>138685</td><td>39533</td><td>45960</td><td>132114</td><td>176426</td><td>70481</td><td>98403</td><td>36961</td><td>100071</td><td>34653</td></tr><tr><td>16386</td><td>51846</td><td>81676</td><td>253514</td><td>24794</td><td>13835</td><td>37998</td><td>43980</td><td>138939</td><td>39535</td><td>45881</td><td>128861</td><td>174010</td><td>70253</td><td>97819</td><td>37170</td><td>102569</td><td>34981</td></tr><tr><td>16328</td><td>52581</td><td>82337</td><td>252253</td><td>25502</td><td>13839</td><td>38052</td><td>43998</td><td>138596</td><td>39270</td><td>45902</td><td>129694</td><td>166124</td><td>70126</td><td>98896</td><td>37285</td><td>102081</td><td>34945</td></tr><tr><td>16695</td><td>51992</td><td>82066</td><td>252656</td><td>25048</td><td>13892</td><td>37954</td><td>43977</td><td>142144</td><td>39364</td><td>42754</td><td>128938</td><td>164949</td><td>70532</td><td>98372</td><td>36817</td><td>104082</td><td>34885</td></tr><tr><td>3 Avg</td><td>0,166666666666667</td><td>0,166666666666667</td><td>0,333333333333333</td><td>0,333333333333333</td><td>0</td><td>0,166666666666667</td><td>0</td><td>0</td><td>0</td><td>0,166666666666667</td><td>0,166666666666667</td><td>0,166666666666667</td><td>0,333333333333333</td><td>0</td><td>0,5</td><td>0,166666666666667</td><td>0</td><td>0,5</td><td>0,166666666666667</td><td>0</td><td>0</td><td>0</td><td>0</td><td>0,166666666666667</td></tr><tr><td>520 Avg</td><td>16748,5</td><td>52657,8333333333</td><td>82083,1666666667</td><td>252041</td><td></td><td></td><td>25514,3333333333</td><td></td><td></td><td>14240,8333333333</td><td></td><td></td><td>38308,3333333333</td><td>44595</td><td>137639,666666667</td><td>39794,3333333333</td><td>46475,5</td><td>128530</td><td>174382,333333333</td><td>70996,3333333333</td><td>98260,6666666667</td><td>37815,8333333333</td><td>102865,5</td><td>35432,1666666667</td></tr><tr><td>avg</td><td>8374,33333333333</td><td>26329</td><td>41041,75</td><td>126020,666666667</td><td></td><td></td><td>12757,1666666667</td><td></td><td></td><td>7120,5</td><td></td><td></td><td>19154,3333333333</td><td>22297,5</td><td>68820,0833333333</td><td>19897,25</td><td>23237,75</td><td>64265,25</td><td>87191,25</td><td>35498,1666666667</td><td>49130,3333333333</td><td>18907,9166666667</td><td>51432,75</td><td>17716,1666666667</td></tr></tbody></table></details>and decided to go for DoubleOpenHashSet, as the loop part seems to outweigh the fetching part. <br>
Constatly reducing the size of the TreeSets was slower, regardless of whether I did it only when there were elements to remove or not and regardless of whether I reassigned with the headSet or cleared the tailSet.<br>
The ArrayLists were slower in both cases and for some reason (bad growth policy?) caused a heap space exception for the 520 block range, even though the data fits into memory, as shown by the other approaches.<br>
PS. your attempts to write helpful comments are mostly successfull (or always, I don't know which are such attempts and which aren't). For exaple without your last comment here, I wouldn't know about fastutils and its boxing-free collections.

leijurv:
Very cool! Thanks!

I'm fine with this, then. Maybe add in a comment to the code saying "see pull request #1978" (this is something I've done before in the code when there's a particularly tricky bit of code).

Also, did you try benching the approach I mentioned first with two full loops through the area, but no data structure storing the costs?

(also could you add a space between `while` and `(` in both places 😇)

ZacSharp:
Yes, it's listed as "Double Iteration". For the 3 block radius it seems to be really slightly faster (hard to tell from one 0 more since that can be pure luck) and for the 520 block radius it was even slower that `HashSet<Double>`.

Would a `private static final Double2DoubleOpenHashMap cachedHeuristics` be a good idea here?

ZacSharp:
@leijurv? Could you please merge this one as well or give feedback on why you don't merge it?

PR #1995, opened 2020-09-04
add mineOnlyExposedOres setting

### LINKED ISSUE #1168: Add "Ignore Not Exposed Ores" Option in AutoMine
## Describe your suggestion
This option should ignore ores that are not exposed to the air, because some servers use anti-Xray and they will replace not exposed stone with fake ores.

## Settings
There should be an option in automine that will automatically filter out not exposed ores.

## Context
For servers that use anti-Xray plugins, this would be very helpful to automine, because then we can only use legit mine, but we can also use Xray but only targeting exposed ores.

scorbett123:
adds an exposed Ores only setting.
if there is air no air on any side then it removes it from the goal.
fixes #1168

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -727,6 +727,11 @@ public final class Settings {
     public final Setting<Integer> maxCachedWorldScanCount = new Setting<>(10);
 
     /**
+     * This will only allow baritone to mine exposed ores, can be used to stop ore obfuscators on servers that use them.
+     */
+    public final Setting<Boolean> allowOnlyExposedOres = new Setting<>(false);
+
+    /**
      * When GetToBlock doesn't know any locations for the desired block, explore randomly instead of giving up.
      */
     public final Setting<Boolean> exploreForBlocks = new Setting<>(true);
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -734,7 +734,7 @@ public final class Settings {
     /**
      * When allowOnlyExposedOres is enabled this is the distance around to search.
      * <p>
-     * I recommend keeping this value low as the amount of blocks that need to be scanned increases exponentially.
+     * It is recommended to keep this value low, as it exponentially increases calculation times and also to keep the
      */
     public final Setting<Integer> allowOnlyExposedOresDistance = new Setting<>(1);
 
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -732,6 +732,13 @@ public final class Settings {
     public final Setting<Boolean> allowOnlyExposedOres = new Setting<>(false);
 
     /**
+     * When allowOnlyExposedOres is enabled this is the distance around to search.
+     * <p>
+     * I recommend keeping this value low as the amount of blocks that need to be scanned increases exponentially.
+     */
+    public final Setting<Integer> allowOnlyExposedOresDistance = new Setting<>(1);
+
+    /**
      * When GetToBlock doesn't know any locations for the desired block, explore randomly instead of giving up.
      */
     public final Setting<Boolean> exploreForBlocks = new Setting<>(true);
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -734,7 +734,7 @@ public final class Settings {
     /**
      * When allowOnlyExposedOres is enabled this is the distance around to search.
      * <p>
-     * It is recommended to keep this value low, as it exponentially increases calculation times and also to keep the
+     * It is recommended to keep this value low, as it dramatically increases calculation times.
      */
     public final Setting<Integer> allowOnlyExposedOresDistance = new Setting<>(1);
 
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -562,4 +562,11 @@ public interface MovementHelper extends ActionCosts, Helper {
     enum PlaceResult {
         READY_TO_PLACE, ATTEMPTING, NO_OPTION;
     }
+    public static boolean isTransparent(Block b) {
+
+        return b== Blocks.AIR ||
+                b == Blocks.FLOWING_LAVA ||
+                b == Blocks.FLOWING_WATER ||
+                b == Blocks.WATER;
+    }
 }
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -562,9 +562,9 @@ public interface MovementHelper extends ActionCosts, Helper {
     enum PlaceResult {
         READY_TO_PLACE, ATTEMPTING, NO_OPTION;
     }
-    public static boolean isTransparent(Block b) {
+    static boolean isTransparent(Block b) {
 
-        return b== Blocks.AIR ||
+        return b == Blocks.AIR ||
                 b == Blocks.FLOWING_LAVA ||
                 b == Blocks.FLOWING_WATER ||
                 b == Blocks.WATER;
--- a/src/main/java/baritone/process/MineProcess.java
+++ b/src/main/java/baritone/process/MineProcess.java
@@ -40,7 +40,6 @@ import net.minecraft.entity.item.EntityItem;
 import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.Vec3d;
 
 import java.util.*;
 import java.util.stream.Collectors;
@@ -411,10 +410,11 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
                 .filter(pos -> MineProcess.plausibleToBreak(ctx, pos))
 
                 .filter(pos -> {
-                    if (Baritone.settings().allowOnlyExposedOres.value)
+                    if (Baritone.settings().allowOnlyExposedOres.value) {
                         return isNextToAir(ctx, pos);
-                    else
+                    } else {
                         return true;
+                    }
                 })
 
                 .filter(pos -> !blacklist.contains(pos))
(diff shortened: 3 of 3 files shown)


Enteerman (changes requested):
LGTM, but why not have an integer setting since there are servers that hide ores except touches transparent block in a radius of two.

And, blocks like water, lava are see-through too and don't forget to include them.

scorbett123:
converted to draft due to a bug in the latest commit.

src/api/java/baritone/api/Settings.java
```diff
@@ -726,6 +726,18 @@
...
 
+    /**
+     * This will only allow baritone to mine exposed ores, can be used to stop ore obfuscators on servers that use them.
+     */
+    public final Setting<Boolean> allowOnlyExposedOres = new Setting<>(false);
+
+    /**
+     * When allowOnlyExposedOres is enabled this is the distance around to search.
+     * <p>
+     * I recommend keeping this value low as the amount of blocks that need to be scanned increases exponentially.
```
dominikaaaa, line 737: Rephrase this to be like
```diff
-      * I recommend keeping this value low as the amount of blocks that need to be scanned increases exponentially.
+      * It is recommended to keep this value low, as it exponentially increases calculation times.
```

src/main/java/baritone/process/MineProcess.java
```diff
@@ -409,6 +410,8 @@ private void addNearby() {
                 // remove any that are implausible to mine (encased in bedrock, or touching lava)
                 .filter(pos -> MineProcess.plausibleToBreak(ctx, pos))
 
+                .filter(pos -> isNextToAir(ctx, pos))
```
dominikaaaa, line 413: Shouldn't this only calculate when the setting is enabled?

src/main/java/baritone/process/MineProcess.java
```diff
@@ -452,4 +461,90 @@ public void mine(int quantity, BlockOptionalMetaLookup filter) {
...
+                        return true;
+                    }
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private static double lengthSq(double x, double y, double z) {
```
dominikaaaa, line 539: Can't you take a vec3d and square it?

src/main/java/baritone/process/MineProcess.java
```diff
@@ -452,4 +461,90 @@ public void mine(int quantity, BlockOptionalMetaLookup filter) {
...
+        return (x * x) + (y * y) + (z * z);
+    }
+
+    public static boolean isTransparent(BlockPos pos, CalculationContext ctx) {
+        IBlockState blockState = ctx.bsi.get0(pos);
```
dominikaaaa, line 544: are you *sure* `get0` is the correct method here?

dominikaaaa (changes requested):

src/main/java/baritone/process/MineProcess.java
scorbett123: what do you suggest to use instead. If you use just get() then you get illegal state exceptions due to running it on the wrong thread. This is also the same as in the method above where it sees if it is plausible to break.

leijurv: I believe `get0` is correct in this case.

src/main/java/baritone/process/MineProcess.java
```diff
@@ -452,4 +466,86 @@ public void mine(int quantity, BlockOptionalMetaLookup filter) {
...
+
+        return false;
+    }
+
+    public static boolean isTransparent(BlockPos pos, CalculationContext ctx) {
+        IBlockState blockState = ctx.bsi.get0(pos);
+        return blockState.getBlock() == Blocks.AIR ||
+                blockState.getBlock() == Blocks.FLOWING_LAVA ||
+                blockState.getBlock() == Blocks.FLOWING_WATER ||
+                blockState.getBlock() == Blocks.WATER;
```
leijurv, line 549: This helper `isTransparent` belongs in `MovementHelper`. I believe there may already be something equivalent there, but I'm not certain.

src/main/java/baritone/process/MineProcess.java
```diff
@@ -452,4 +466,86 @@ public void mine(int quantity, BlockOptionalMetaLookup filter) {
             rescan(new ArrayList<>(), new CalculationContext(baritone));
         }
     }
+
+    public static boolean makeSphere(BlockPos pos, double radius, CalculationContext ctx) {
```
leijurv, line 470: This is extreme cringe. 

Literally everything about this function is cringe.

What is the use case for this? Ignore enteerman idk who that is.

Can't this be as simple as like:

```
for (int dx = -1; dx <= 1; dx++) {
    for (int dy = -1; dy <= 1; dy++) {
        for (int dz = -1; dz <= 1; dz++) {
            if (MovementHelper.isTransparent(pos.add(dx,dy,dz))) {
                return false;
            }
        }
    }
}
```

src/main/java/baritone/process/MineProcess.java
leijurv: ya

src/main/java/baritone/process/MineProcess.java
dominikaaaa: leij's example is how I was doing spheres and enteerman or someone made fun of it :clown:

leijurv: enteerman is the real clown then?

this isn't intended to be a sphere, why would it be?

dominikaaaa: # :joy_cat:

leijurv: yeah it shouldn't be a sphere

Enteerman: i am bad at english and sorry if i didnt explain it well. It wasn't intended to be a sphere. just on some servers that blocks that are two blocks from a transparent block are all legit. I think hypixel does that in UHC so you can check that out. You can ignore me, I don't care. I sketched this out and this is how is should work: https://imgur.com/0O7BkWt

Enteerman: I dont know why this method is large but just take the 6 blocks that are adjacent to the block, and take adjacents of the adjacents, to check if it is air.

Enteerman: Or use ur method idk i dont play the game anymore, ask someone else's opinion on this

ZacSharp: for me @leijurv's method looks *way* better but it only checks the direct and diagonal neighbours. I'd rather use
 ```
 for (int dx = -radius; dx <= radius; dx++) {
     for (int dy = -radius; dy <= radius; dy++) {
         for (int dz = -radius; dz <= radius; dz++) {
             if (dx + dy + dz <= radius && MovementHelper.isTransparent(pos.add(dx,dy,dz))) {
                 return true;
             }
         }
     }
 }
 ```
to check all blocks that have a manhattan distance of at most `radius` (I think that is was @Enteerman wants to show with his image)

scorbett123:
sorry, accidentally pushed to this branch from the wrong repo, ignore latest commits.

src/api/java/baritone/api/Settings.java
```diff
@@ -726,6 +726,18 @@
...
 
+    /**
+     * This will only allow baritone to mine exposed ores, can be used to stop ore obfuscators on servers that use them.
+     */
+    public final Setting<Boolean> allowOnlyExposedOres = new Setting<>(false);
+
+    /**
+     * When allowOnlyExposedOres is enabled this is the distance around to search.
+     * <p>
+     * It is recommended to keep this value low, as it exponentially increases calculation times and also to keep the
```
ZacSharp, line 737: this sentence is incomplete and the amount of checked blocks does not increase exponentially but by (approximately) `1.3*r**3 + 2*r**2 + 2.6*r + 1`

ZacSharp (changes requested):

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -562,4 +562,11 @@ static PlaceResult attemptToPlaceABlock(MovementState state, IBaritone baritone,
     enum PlaceResult {
         READY_TO_PLACE, ATTEMPTING, NO_OPTION;
     }
+    public static boolean isTransparent(Block b) {
+
+        return b== Blocks.AIR ||
```
leijurv, line 567: space

src/main/java/baritone/process/MineProcess.java
```diff
@@ -409,6 +410,13 @@ private void addNearby() {
                 // remove any that are implausible to mine (encased in bedrock, or touching lava)
                 .filter(pos -> MineProcess.plausibleToBreak(ctx, pos))
 
+                .filter(pos -> {
+                    if (Baritone.settings().allowOnlyExposedOres.value)
```
leijurv, line 414: plz use `{` in the if

src/main/java/baritone/process/MineProcess.java
```diff
@@ -420,6 +428,22 @@ private void addNearby() {
         return locs;
     }
 
+    public static boolean isNextToAir(CalculationContext ctx, BlockPos pos) {
+        int radius = Baritone.settings().allowOnlyExposedOresDistance.value;
+        for (int dx = -radius; dx <= radius; dx++) {
+            for (int dy = -radius; dy <= radius; dy++) {
+                for (int dz = -radius; dz <= radius; dz++) {
+                    if (Math.abs(dx) + Math.abs(dy) + Math.abs(dz) <= radius
+                            && MovementHelper.isTransparent(ctx.getBlock(pos.getX()+dx, pos.getY()+dy, pos.getZ()+dz))) {
```
leijurv, line 437: run through autoformatter? put spaces before and after each `+`

leijurv (changes requested):
LGTM, fix formatting to match baritone style and i'll merge

leijurv:
@scorbett123 wait what, why did you rearrange those functions and classes? it makes the diff much bigger for no reason. can we leave the order the same and just add the new function? like how it was an hour ago?

scorbett123:
Sorry, didn't mean to, something went wrong when committing, I think it might have been my auto formatter.

leijurv:
ohh yea, that's true intellij has this `rearrange entries` option that i think you gotta untick

scorbett123:
is this better?

leijurv:
almost haha just one thing can you add curly braces to the `if`? 😹

scorbett123:
Sorry, in my fourth attempt to fix the formatting and failing with using git I obviously forgot about those.

PR #2020, opened 2020-09-16
make `#wp goto` behave like `#wp goal` and `#path`

ZacSharp:
It now is a combined `#wp goal` and `#path` just like `#goto` is `#goal` and `#path`


dominikaaaa (approved):

ZacSharp:
maybe also make `#home` be an alias for `#wp goto home` rather than `#wp goal home`

PR #2079, opened 2020-10-16
add a minimum Y level during mining.

### LINKED ISSUE #1883: Parameter to limit Y level for automine (Non legit)
## Describe your suggestion
I would like a parameter to limit Y level for example MaximumMineYLevel 50 will allow baritone only to mine on level 51 and above. 

## Settings
MaximumMineYLevel

## Context
The request is done in order to bypass some Xray which places fake blocks in deeper Y levels and with that setting Baritone would be able to avoid them while Xray mining is performed.

scorbett123:
adds a minimum Y level while mining.

fixes #1883.


l1ving (approved):

recursivetree:
Maybe add a max y level aswell

Issue #2054, opened 2020-10-05
limited distance radius


GaiaB0t:
## Describe your suggestion
this would prevent baritone from pathing more than a set radius to either a #home or from where the #mine command was initiated
this would only affect the #mine command and not goto or path
## Settings
#limit set (number of blocks from the point)
#limit home(would use home waypoint)
or #limit start(would limit from wherever you started the #mine command)
## Context
when I do #mine diamond_ore it keeps traveling far from my base even tho my block esp says there are diamonds close by.
this would ONLY affect the #mine commands
## Final checklist
- [x] I know how to properly use checkboxes 
- [x] I have not used any OwO's or UwU's in this issue.

derBobo:
how about changing the mine command to work like this #mine [optional blockradius] <block1> <count1> <block2> <count2> ... I think thats much easier to understand than adding more settings

derBobo:
this would also work for the farm command

GaiaB0t:
yea that makes more sense than what  I was saying

Colmonkey:
Could there also be an option to restrict this to selection? instead of a radius, you could create your custom selection using the #sel functions, then baritone would not do anything outside of the selection.

PR #2085, opened 2020-10-19
add range to mining.

scorbett123:
This adds a range argument to mining.

Any suggestions of how to improve the command please say. Currently it is, #mine ["radius" ["home"] distance] [quantity] block1 ...

Based from #2054.

### DIFF
--- a/src/api/java/baritone/api/process/IMineProcess.java
+++ b/src/api/java/baritone/api/process/IMineProcess.java
@@ -20,6 +20,7 @@ package baritone.api.process;
 import baritone.api.utils.BlockOptionalMeta;
 import baritone.api.utils.BlockOptionalMetaLookup;
 import net.minecraft.block.Block;
+import net.minecraft.util.math.BlockPos;
 
 import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


scorbett123:
Closing in favour of #2088

avneh:
sdd

derBobo:
oh didn't see someone already created a pr for that

scorbett123:
Don't worry, your implementation is much cleaner.

avneh:
ya i think you should use filtration instead of full so it doesnt crash

avneh:
basically it should only build with the items you have marked

scorbett123:
What?

avneh:
basically the player should choose what they want to use like if they wer
surrounded by dirt but they wanted cobblestone the dirt would and stuff
would be invis but you see cobble

avneh:
or you could turn x into y

avneh:
x = anything y = anything

derBobo:
am i the only one who is confused by avnehs comments ?

avneh:
no their are others

PR #2086, opened 2020-10-19
[Change] Ensure correct chunk provider with Sodium

CDAGaming:
This allows Sodium to work with Baritone. Despite it being only on Fabric, it is similarly in production with Forge and it's a minor enough change

### DIFF
--- a/src/main/java/baritone/utils/BlockStateInterface.java
+++ b/src/main/java/baritone/utils/BlockStateInterface.java
@@ -65,7 +65,7 @@ public class BlockStateInterface {
     public BlockStateInterface(World world, WorldData worldData, boolean copyLoadedChunks) {
         this.world = world;
         this.worldData = worldData;
-        if (copyLoadedChunks) {
+        if (copyLoadedChunks && !world.getChunkProvider().getClass().getSimpleName().equals("SodiumChunkManager")) {
             this.provider = ((IClientChunkProvider) world.getChunkProvider()).createThreadSafeCopy();
         } else {
             this.provider = (ClientChunkProvider) world.getChunkProvider();


l1ving (approved):

leijurv:
No, this is a 🗿🧠 (ROCK BRAIN) hacky fix. The entire point of `copyLoadedChunks` is to create a **thread safe copy** of the loaded chunk map, for use on the pathing thread.

I would like to point out @CDAGaming that this **exact** thing was the main issue that I pointed to way back ~15 months ago when I was saying that fabritone just made whatever changes were needed to make it compile without actually thinking through what they mean. This **literal exact thing** of "oh it's difficult to make a thread safe copy of the loaded chunk map now that it isn't just a hashmap like it was in 1.13" -> "so i will just... not copy it" "i'm sure that code wasn't needed anyway"

This directly caused (rare) random ConcurrentModificationException s and game crashes. Indirectly, it caused incorrect, suboptimal, and invalid paths to be calculated, when the loaded chunk map changes while pathing is in progress (this would happen even if the access were synchronized to the main thread). If a new chunk suddenly appears while the previous nodes into that chunk were already calculated as COST_INF, A* can no longer guarantee correctness as it is now going to take a consistently suboptimal path, because the route only appeared chronologically later in the calculation process, so the heuristic+cost metric was necessarily much higher.

I'm sure the exact same thing would happen with sodium in this PR and perhaps even worse, I'm not sure how sodium stores the map under the hood. In an ideal case we would just see incorrect ridiculous paths, and no crashes, if Sodium synchronizes accesses safely, and in a worst case (such as the chunk map introduced in 1.14.4) we would see game crashes.

The `if (copyLoadedChunks) { createThreadSafeCopy()` is not something that I wrote in there just for fun for no actual reason just to mess with you. It's not to be disabled just because it's the easiest path to making baritone work with whatever this other mod is.

fr1kin:
fabric is deprecated

PR #2088, opened 2020-10-20
improved mine command (range & quantity for every single block)

derBobo:
based on #2054  also allowing to set the mining quantity for every single block. the command now works like this `mine [[home] blockradius] block1 [quantity1] [block2 [quantity2] block3 [quantity3] ...]`

### DIFF
--- a/src/api/java/baritone/api/process/IMineProcess.java
+++ b/src/api/java/baritone/api/process/IMineProcess.java
@@ -21,6 +21,8 @@ import baritone.api.utils.BlockOptionalMeta;
 import baritone.api.utils.BlockOptionalMetaLookup;
 import net.minecraft.block.Block;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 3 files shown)


Colmonkey:
I commented this in #2054 but I realize this is probably the more appropriate place to ask this.
Could there also be an option to restrict this to selection? instead of a radius, you could create your custom selection using the #sel functions, then baritone would not do anything outside of the selection.

scorbett123:
@leijurv any reason to hold off on merging this?

leijurv:
* Merge conflict
* Breaking change to the IMineProcess API (e.g. Impact probably wouldn't compile against this anymore)
* https://github.com/cabaletta/baritone/pull/2088/files#diff-4737edbbb24d35ef7f5da331d431395daeb800e643edd7866ab68580b56550f0R96-R100 This is a nontrivial change to the behavior of MineProcess, unrelated
* Using `sqrt(x*x+z*z)>r` instead of `x*x+z*z>r*r`
* Overall kinda weird, not convinced this is useful. Seems like a "kitchen sink" type of feature. I'd worry about potential performance impact. I mean it looks fine overall, but there's a few things (e.g. https://github.com/cabaletta/baritone/pull/2088/files#diff-4737edbbb24d35ef7f5da331d431395daeb800e643edd7866ab68580b56550f0R195) where I worry about Baritone getting incrementally a little bit slower over time with each PR, for example this runs an extra bit of code for every ore every tick /shrug maybe just me being paranoid idk

scorbett123:
Thanks, someone was asking on another issue.

sharkman:
Isn't this one going to be superseded by #2597 if it ever will be implemented?

PR #2138, opened 2020-11-01
add useSwordToMine setting.

### LINKED ISSUE #2136: Sword saver
## Describe your suggestion
Baritone should avoid breaking blocks with swords just like it avoids using the wrong tool.

## Settings
This might respect the itemSaver setting from #2107 

## Context
Swords can be damaged by breaking blocks with them and because their usual slot is Baritones fallback slot this happens almost every time Baritone can't find a fitting tool.

## Final checklist
- [x] I have not used more than these two OwO's and UwU's in this issue.

scorbett123:
This setting makes it so that you can change whether baritone will use a sword whilst mining

fixes #2136


ZacSharp:
Why would I want it to use my sword? I think it should treat swords as wrong tools, which means avoiding them independently from any setting (or include all tools in this setting).

l1ving (changes requested):
Honestly this should default to false

ZacSharp (approved):
I'd still do `if (itemStack.getItem() instanceof ItemSword)` rather than `if (!Baritone.settings().useSwordToMine.value && itemStack.getItem() instanceof ItemSword)` but that's not my choice and not really important

scorbett123:
@leijurv Can you merge this one as well?

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -124,9 +125,14 @@ public int getBestSlot(Block b, boolean preferSilkTouch, boolean pathingCalculat
...
         for (int i = 0; i < 9; i++) {
             ItemStack itemStack = player.inventory.getStackInSlot(i);
+            if (!Baritone.settings().useSwordToMine.value && itemStack.getItem() instanceof ItemSword){
+                continue;
+            }
+          
             if (Baritone.settings().itemSaver.value && itemStack.getItemDamage() >= itemStack.getMaxDamage() && itemStack.getMaxDamage() > 1) {
                 continue;
             }
+
```
leijurv, line 135: why?

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -124,9 +125,14 @@ public int getBestSlot(Block b, boolean preferSilkTouch, boolean pathingCalculat
         IBlockState blockState = b.getDefaultState();
         for (int i = 0; i < 9; i++) {
             ItemStack itemStack = player.inventory.getStackInSlot(i);
+            if (!Baritone.settings().useSwordToMine.value && itemStack.getItem() instanceof ItemSword){
```
leijurv, line 128: space between `)` and `{`

src/api/java/baritone/api/Settings.java
```diff
@@ -1118,6 +1118,11 @@
      */
     public final Setting<Boolean> renderSelectionCorners = new Setting<>(true);
 
+    /**
+     * Use sword to mine.
+     */
+    public final Setting<Boolean> useSwordToMine = new Setting<>(false);
```
leijurv, line 1124: wait a sec - the current existing behavior is to allow swords, right? because when useSwordToMine is false, you continue in the toolset loop. the current behavior would be `useSwordToMine true`, right? so shouldn't this default to true?

scorbett123: Further up in this thread people stood that it should default to false, why would you want to use your sword?

leijurv: Cobwebs

scorbett123:
Thanks

PR #2142, opened 2020-11-03
Add Diagonal Jumps

Zephreo:
Adds a new movement class that allows diagonal or any angle of jumps to be considered.

Also adds 4 block jumps (using momentum jump)

2 settings for the parkour

1. allowParkourAdv, since it may add extra computations that are unnecessary
2. allowParkourMomentumOrNeo, since these jumps (4 blocks and around pillar) are much more unreliable

I use (almost) standard jump notation. So 3x2+1 is a jump with a 2 (3-1) block gap in the direction the player is jumping plus 2 blocks to the side and 1 block up. This way I can denote a 1x1 "jump" to be the block 1 block diagonal from the player. +- axis are not required.

### DIFF
--- a/src/api/java/baritone/api/utils/Rotation.java
+++ b/src/api/java/baritone/api/utils/Rotation.java
@@ -81,6 +81,22 @@ public class Rotation {
     }
 
     /**
+     * Returns a copy of the opposite rotation of the given rotation
+     *
+     * @return
+     */
+    public Rotation opposite() {
+        float newYaw = this.yaw - 180;
+        if(newYaw < -180) {
+            newYaw += 360;
+        }
+        return new Rotation(
+                newYaw,
+                this.pitch * -1
+        );
+    }
+
+    /**
      * @return A copy of this rotation with the pitch clamped
      */
     public Rotation clamp() {
--- a/src/api/java/baritone/api/utils/Rotation.java
+++ b/src/api/java/baritone/api/utils/Rotation.java
@@ -81,17 +81,11 @@ public class Rotation {
     }
 
     /**
-     * Returns a copy of the opposite rotation of the given rotation
-     *
-     * @return
+     * @return A copy of the opposite rotation of the given rotation
      */
     public Rotation opposite() {
-        float newYaw = this.yaw - 180;
-        if(newYaw < -180) {
-            newYaw += 360;
-        }
         return new Rotation(
-                newYaw,
+                this.yaw - 180,
                 this.pitch * -1
         );
     }
--- a/src/api/java/baritone/api/utils/Rotation.java
+++ b/src/api/java/baritone/api/utils/Rotation.java
@@ -81,16 +81,6 @@ public class Rotation {
     }
 
     /**
-     * @return A copy of the opposite rotation of the given rotation
-     */
-    public Rotation opposite() {
-        return new Rotation(
-                this.yaw - 180,
-                this.pitch * -1
-        );
-    }
-
-    /**
      * @return A copy of this rotation with the pitch clamped
      */
     public Rotation clamp() {
--- a/src/main/java/baritone/pathing/movement/Moves.java
+++ b/src/main/java/baritone/pathing/movement/Moves.java
@@ -22,6 +22,9 @@ import baritone.pathing.movement.movements.*;
 import baritone.utils.pathing.MutableMoveResult;
 import net.minecraft.util.EnumFacing;
 
+import java.util.Collection;
+import java.util.Collections;
+
 /**
  * An enum of all possible movements attached to all possible directions they could be taken in
  *
@@ -30,8 +33,8 @@ import net.minecraft.util.EnumFacing;
 public enum Moves {
     DOWNWARD(0, -1, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementDownward(context.getBaritone(), src, src.down());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementDownward(context.getBaritone(), src, src.down()));
         }
 
         @Override
@@ -42,8 +45,8 @@ public enum Moves {
 
     PILLAR(0, +1, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementPillar(context.getBaritone(), src, src.up());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementPillar(context.getBaritone(), src, src.up()));
         }
 
         @Override
@@ -54,8 +57,8 @@ public enum Moves {
 
     TRAVERSE_NORTH(0, 0, -1) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementTraverse(context.getBaritone(), src, src.north());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementTraverse(context.getBaritone(), src, src.north()));
         }
 
         @Override
@@ -66,8 +69,8 @@ public enum Moves {
 
     TRAVERSE_SOUTH(0, 0, +1) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementTraverse(context.getBaritone(), src, src.south());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementTraverse(context.getBaritone(), src, src.south()));
         }
 
         @Override
@@ -78,8 +81,8 @@ public enum Moves {
 
     TRAVERSE_EAST(+1, 0, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementTraverse(context.getBaritone(), src, src.east());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementTraverse(context.getBaritone(), src, src.east()));
         }
 
         @Override
@@ -90,8 +93,8 @@ public enum Moves {
 
     TRAVERSE_WEST(-1, 0, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementTraverse(context.getBaritone(), src, src.west());
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementTraverse(context.getBaritone(), src, src.west()));
         }
 
         @Override
@@ -102,8 +105,8 @@ public enum Moves {
 
     ASCEND_NORTH(0, +1, -1) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x, src.y + 1, src.z - 1));
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x, src.y + 1, src.z - 1)));
         }
 
         @Override
@@ -114,8 +117,8 @@ public enum Moves {
 
     ASCEND_SOUTH(0, +1, +1) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x, src.y + 1, src.z + 1));
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x, src.y + 1, src.z + 1)));
         }
 
         @Override
@@ -126,8 +129,8 @@ public enum Moves {
 
     ASCEND_EAST(+1, +1, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x + 1, src.y + 1, src.z));
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x + 1, src.y + 1, src.z)));
         }
 
         @Override
@@ -138,8 +141,8 @@ public enum Moves {
 
     ASCEND_WEST(-1, +1, 0) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x - 1, src.y + 1, src.z));
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(new MovementAscend(context.getBaritone(), src, new BetterBlockPos(src.x - 1, src.y + 1, src.z)));
         }
 
         @Override
@@ -150,13 +153,13 @@ public enum Moves {
 
     DESCEND_EAST(+1, -1, 0, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
             if (res.y == src.y - 1) {
-                return new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             } else {
-                return new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             }
         }
 
@@ -168,13 +171,13 @@ public enum Moves {
 
     DESCEND_WEST(-1, -1, 0, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
             if (res.y == src.y - 1) {
-                return new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             } else {
-                return new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             }
         }
 
@@ -186,13 +189,13 @@ public enum Moves {
 
     DESCEND_NORTH(0, -1, -1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
             if (res.y == src.y - 1) {
-                return new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             } else {
-                return new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             }
         }
 
@@ -204,13 +207,13 @@ public enum Moves {
 
     DESCEND_SOUTH(0, -1, +1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
             if (res.y == src.y - 1) {
-                return new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementDescend(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             } else {
-                return new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z));
+                return Collections.singletonList(new MovementFall(context.getBaritone(), src, new BetterBlockPos(res.x, res.y, res.z)));
             }
         }
 
@@ -222,10 +225,10 @@ public enum Moves {
 
     DIAGONAL_NORTHEAST(+1, 0, -1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
-            return new MovementDiagonal(context.getBaritone(), src, EnumFacing.NORTH, EnumFacing.EAST, res.y - src.y);
+            return Collections.singletonList(new MovementDiagonal(context.getBaritone(), src, EnumFacing.NORTH, EnumFacing.EAST, res.y - src.y));
         }
 
         @Override
@@ -236,10 +239,10 @@ public enum Moves {
 
     DIAGONAL_NORTHWEST(-1, 0, -1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
-            return new MovementDiagonal(context.getBaritone(), src, EnumFacing.NORTH, EnumFacing.WEST, res.y - src.y);
+            return Collections.singletonList(new MovementDiagonal(context.getBaritone(), src, EnumFacing.NORTH, EnumFacing.WEST, res.y - src.y));
         }
 
         @Override
@@ -250,10 +253,10 @@ public enum Moves {
 
     DIAGONAL_SOUTHEAST(+1, 0, +1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
-            return new MovementDiagonal(context.getBaritone(), src, EnumFacing.SOUTH, EnumFacing.EAST, res.y - src.y);
+            return Collections.singletonList(new MovementDiagonal(context.getBaritone(), src, EnumFacing.SOUTH, EnumFacing.EAST, res.y - src.y));
         }
 
         @Override
@@ -264,10 +267,10 @@ public enum Moves {
 
     DIAGONAL_SOUTHWEST(-1, 0, +1, false, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             MutableMoveResult res = new MutableMoveResult();
             apply(context, src.x, src.y, src.z, res);
-            return new MovementDiagonal(context.getBaritone(), src, EnumFacing.SOUTH, EnumFacing.WEST, res.y - src.y);
+            return Collections.singletonList(new MovementDiagonal(context.getBaritone(), src, EnumFacing.SOUTH, EnumFacing.WEST, res.y - src.y));
         }
 
         @Override
@@ -278,8 +281,8 @@ public enum Moves {
 
     PARKOUR_NORTH(0, 0, -4, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return MovementParkour.cost(context, src, EnumFacing.NORTH);
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(MovementParkour.cost(context, src, EnumFacing.NORTH));
         }
 
         @Override
@@ -290,8 +293,8 @@ public enum Moves {
 
     PARKOUR_SOUTH(0, 0, +4, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return MovementParkour.cost(context, src, EnumFacing.SOUTH);
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(MovementParkour.cost(context, src, EnumFacing.SOUTH));
         }
 
         @Override
@@ -302,8 +305,8 @@ public enum Moves {
 
     PARKOUR_EAST(+4, 0, 0, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return MovementParkour.cost(context, src, EnumFacing.EAST);
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(MovementParkour.cost(context, src, EnumFacing.EAST));
         }
 
         @Override
@@ -314,8 +317,8 @@ public enum Moves {
 
     PARKOUR_WEST(-4, 0, 0, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
-            return MovementParkour.cost(context, src, EnumFacing.WEST);
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
+            return Collections.singletonList(MovementParkour.cost(context, src, EnumFacing.WEST));
         }
 
         @Override
@@ -326,7 +329,7 @@ public enum Moves {
 
     PARKOUR_ADV_NORTH(0, 0, -5, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             return MovementParkourAdv.cost(context, src, EnumFacing.NORTH);
         }
 
@@ -338,7 +341,7 @@ public enum Moves {
 
     PARKOUR_ADV_SOUTH(0, 0, +5, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             return MovementParkourAdv.cost(context, src, EnumFacing.SOUTH);
         }
 
@@ -350,7 +353,7 @@ public enum Moves {
 
     PARKOUR_ADV_EAST(+5, 0, 0, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             return MovementParkourAdv.cost(context, src, EnumFacing.EAST);
         }
 
@@ -362,7 +365,7 @@ public enum Moves {
 
     PARKOUR_ADV_WEST(-5, 0, 0, true, true) {
         @Override
-        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
             return MovementParkourAdv.cost(context, src, EnumFacing.WEST);
         }
 
@@ -391,7 +394,7 @@ public enum Moves {
         this(x, y, z, false, false);
     }
 
-    public abstract Movement apply0(CalculationContext context, BetterBlockPos src);
+    public abstract Collection<Movement> apply0(CalculationContext context, BetterBlockPos src);
 
     public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
         if (dynamicXZ || dynamicY) {
--- a/src/main/java/baritone/pathing/movement/Moves.java
+++ b/src/main/java/baritone/pathing/movement/Moves.java
@@ -332,7 +332,7 @@ public enum Moves {
 
         @Override
         public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
-            MovementParkourAdv.cost(context, new BetterBlockPos(x, y, z), result, EnumFacing.NORTH);
+            MovementParkourAdv.cost(context, x, y, z, result, EnumFacing.NORTH);
         }
     },
 
@@ -344,7 +344,7 @@ public enum Moves {
 
         @Override
         public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
-            MovementParkourAdv.cost(context, new BetterBlockPos(x, y, z), result, EnumFacing.SOUTH);
+            MovementParkourAdv.cost(context, x, y, z, result, EnumFacing.SOUTH);
         }
     },
 
@@ -356,7 +356,7 @@ public enum Moves {
 
         @Override
         public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
-            MovementParkourAdv.cost(context, new BetterBlockPos(x, y, z), result, EnumFacing.EAST);
+            MovementParkourAdv.cost(context, x, y, z, result, EnumFacing.EAST);
         }
     },
 
@@ -368,7 +368,7 @@ public enum Moves {
 
         @Override
         public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
-            MovementParkourAdv.cost(context, new BetterBlockPos(x, y, z), result, EnumFacing.WEST);
+            MovementParkourAdv.cost(context, x, y, z, result, EnumFacing.WEST);
         }
     };
 
--- a/src/main/java/baritone/pathing/movement/Moves.java
+++ b/src/main/java/baritone/pathing/movement/Moves.java
@@ -33,8 +33,8 @@ import java.util.Collections;
 public enum Moves {
     DOWNWARD(0, -1, 0) {
         @Override
-        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
-            return Collections.singletonList(new MovementDownward(context.getBaritone(), src, src.down()));
+        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+            return new MovementDownward(context.getBaritone(), src, src.down());
         }
 
         @Override
@@ -45,8 +45,8 @@ public enum Moves {
 
     PILLAR(0, +1, 0) {
         @Override
-        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
-            return Collections.singletonList(new MovementPillar(context.getBaritone(), src, src.up()));
+        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+            return new MovementPillar(context.getBaritone(), src, src.up());
         }
 
         @Override
@@ -57,8 +57,8 @@ public enum Moves {
 
     TRAVERSE_NORTH(0, 0, -1) {
         @Override
-        public Collection<Movement> apply0(CalculationContext context, BetterBlockPos src) {
-            return Collections.singletonList(new MovementTraverse(context.getBaritone(), src, src.north()));
+        public Movement apply0(CalculationContext context, BetterBlockPos src) {
+            return new MovementTraverse(context.getBaritone(), src, src.north());
         }
 
         @Override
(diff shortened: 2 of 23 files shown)


src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+
+    private static final double PREP_OFFSET = 0.2215; // The default prep location for a jump
+
+    private static final double MAX_JUMP_MOMENTUM = 5.3; // We can't make 1bm momentum jumps greater than this distance
+    private static final double MAX_JUMP_SPRINT = 4.6; // We can't make flat sprint jumps greater than this distance
+    private static final double MAX_JUMP_WALK = 3.48; // We can make the jump without sprinting below this distance
+
+    private static final double MAX_JUMP_EDGE = 6;
+
+    //private static final double SPRINT_THRESHOLD = 3.25 + TURN_COST * Math.toRadians(30); // Distance required for a sprint jump
```
l1ving, line 77: remove this line

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        MOMENTUM, // An extra momentum jump 1bm
+        EDGE, // No run up (for higher angle jumps)
+        EDGE_NEO // Around the pillar
+    }
+
+    static {
+        int[][] validQuadrant = {{5, 0, JumpType.MOMENTUM.ordinal()},
+                {1, 1}, {2, 1}, {3, 1}, {4, 1}, {5, 1, JumpType.MOMENTUM.ordinal()},
+                {0, 2, JumpType.EDGE_NEO.ordinal()}, {1, 2, JumpType.EDGE.ordinal()}, {2, 2, JumpType.EDGE.ordinal()}, {3, 2, JumpType.EDGE.ordinal()}, {4, 2, JumpType.MOMENTUM.ordinal()}, {5, 2, JumpType.MOMENTUM.ordinal()},
+                {1, 3, JumpType.EDGE.ordinal()}, {2, 3, JumpType.EDGE.ordinal()}};
```
l1ving, line 90: i-

I'm not even going to ask

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+
+    private static void approxQuadrant(HashSet<Vec3i> output, Vec3d vector, double overlap, boolean posX, boolean posY, boolean posZ) {
+        for(double x = vector.x + (posX ? -1 : 1); (posX ? (x <= Math.floor(vector.x) + overlap) : (x >= Math.ceil(vector.x) - overlap)); x += posX ? 1 : -1) {
+            for(double y = vector.y + (posY ? -1 : 1); (posY ? (y <= Math.floor(vector.y) + overlap) : (y >= Math.ceil(vector.y) - overlap)); y += posY ? 1 : -1) {
+                for(double z = vector.z + (posZ ? -1 : 1); (posZ ? (z <= Math.floor(vector.z) + overlap) : (z >= Math.ceil(vector.z) - overlap)); z += posZ ? 1 : -1) {
+                    output.add(new Vec3i((vector.x * 2) - (x + (posX ? 1 : -1)), (vector.y * 2) - (y + (posY ? 1 : -1)), (vector.z * 2) - (z + (posZ ? 1 : -1))));
+                }
+            }
+        }
+    }
```
l1ving, line 237: is this really the best way to do this?

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+            output.add(new Vec3i(blockX - modX, blockY, blockZ));
+        }
+
+        if(localY <= overlap) {
+            output.add(new Vec3i(blockX, blockY - modY, blockZ));
+        }
+
+        if(localZ <= overlap) {
+            output.add(new Vec3i(blockX, blockY, blockZ - modZ));
+        } //*/
```
l1ving, line 313: remove this if it's not needed

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        return output;
+    }
+
+    @Override
+    public Set<BetterBlockPos> calculateValidPositions() {
+        HashSet<BetterBlockPos> out = new HashSet<BetterBlockPos>();
+        for (Vec3i vec : getLineApprox(direction, 2, 3)) {
+            BetterBlockPos pos = new BetterBlockPos(src.add(vec));
+            out.add(pos);       //Jumping from blocks
+            out.add(pos.up());  //Jumping into blocks
```
l1ving, line 324: don't add random whitespace just to align your comments

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        if (!MovementHelper.fullyPassable(context, srcX + xDiff, srcY + 1, srcZ + zDiff)) { //block in head in directly adjacent block
+            return;
+        }
+        if (!MovementHelper.fullyPassable(context, srcX + xDiff, srcY + 2, srcZ + zDiff)) { //Block above head in directly adjacent block
+            return;
+        }
+
+        IBlockState standingOn = context.get(srcX, srcY - 1, srcZ);
+        if (standingOn.getBlock() == Blocks.VINE || standingOn.getBlock() == Blocks.LADDER || standingOn.getBlock() instanceof BlockStairs || standingOn.getBlock() instanceof BlockLiquid) {
+            //System.out.println("TEST -1, " + srcX + ", " + (srcY - 1) + ", " + srcZ + ", " + " = " + standingOn.getBlock());
```
l1ving, line 413: :neutral_face:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                }
+            } else {
+                switch (type) {
+                    case EDGE: // An edge jump has only a little less momentum than a normal jump
+                    case NORMAL:
+                        maxJump = MAX_JUMP_WALK;
+                        break;
+                        //Default is -1 so these are ignored
+                    //case EDGE_NEO: // can't neo without sprint :(
+                    //case MOMENTUM: // can't momentum without sprint :(
```
l1ving, line 455: :gorilla:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                IBlockState landingOn = context.bsi.get0(destX, srcY - decendAmount - 1, destZ);
+
+                // farmland needs to be canWalkOn otherwise farm can never work at all, but we want to specifically disallow ending a jump on farmland haha
+                if (landingOn.getBlock() != Blocks.FARMLAND && MovementHelper.canWalkOn(context.bsi, destX, srcY - decendAmount - 1, destZ, landingOn)) {
+                    //if (checkOvershootSafety(context.bsi, destX + xDiff, srcY, destZ + zDiff)) {
+                    lowestCost = getMoveResult(context, destX, srcY - decendAmount, destZ, extraAscend, VecUtils.add(posbJump, 0, -decendAmount, 0), simpleDirection, type, 0, lowestCost, res);
+                    continue;
+                    //}
+                    //System.out.println("TEST 4, Flat Failed");
+                    //continue;
```
l1ving, line 504: :neutral_face:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+
+            if (!MovementHelper.fullyPassable(context, destX, srcY + 3, destZ)) {
+                continue; //If block above head at destination
+            }
+
+            if (!context.allowParkourPlace) {
+                continue; // Settings don't allow a parkour place
+            }
+
+            //System.out.println("TEST 6, Parkour Place, in dir " + simpleDirection);
```
l1ving, line 518: r e m o v e

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+            if (placeCost >= COST_INF) {
+                continue; // Not allowed to place here
+            }
+            if (!MovementHelper.isReplaceable(destX, srcY - 1, destZ, toReplace, context.bsi)) {
+                continue; // Not allowed to place here
+            }
+            //TODO
+            //if (!checkOvershootSafety(context.bsi, destX + xDiff, srcY, destZ + zDiff)) {
+                //continue; // Can place but will overshoot into a bad location
+            //}
```
l1ving, line 532: finish this or remove

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                if (MovementHelper.canPlaceAgainst(context.bsi, againstX, againstY, againstZ)) {
+                    lowestCost = getMoveResult(context, destX, srcY, destZ, extraAscend, posbJump, simpleDirection, type, placeCost, lowestCost, res);
+                    continue;
+                }
+            }
+        }
+        res = lowestCost;
+        if (res.cost < 1000) {
+            Vec3i jumpVec = new Vec3i(res.x - srcX, res.y - srcY, res.z - srcZ);
+            //System.out.println("Dir = " + simpleDirection + ", Cost: " + res.cost + ", Distance: " + getDistance(jumpVec, simpleDirection) + ", MoveDis: " + calcMoveDist(context, srcX, srcY, srcZ, extraAscend, jumpVec, simpleDirection));
```
l1ving, line 551: I suggest installing a linter

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+
+        //This distance is unitless as it contains: modifiers on ascends/descends, and the slowdowns in changing directions midair (angle)
+        return distance + TURN_COST_PER_RADIAN * angle;
+    }
+
+    @Override
+    public boolean safeToCancel(MovementState state) {
+        // since we don't really know anything about momentum, it can only be canceled during prep phase (i.e. before the jump)
+        // e.g. can't cancel while cancelling momentum or we may fall off the block.
+        return state.getStatus() != MovementStatus.RUNNING;
```
l1ving, line 597: why is space vs no space after `//` so inconsistent?

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                accuracy = 0.025; //Don't want to fall off
+                break;
+            case EDGE_NEO:
+                offset = new Vec3d(simpleDirection.getDirectionVec()).scale(0.8).add(new Vec3d(destDir.getOpposite().getDirectionVec()).scale(0.2));
+                accuracy = 0.025;
+                break;
+        }
+        Vec3d preJumpPos = offset.add(src.x + 0.5, ctx.player().posY, src.z + 0.5);
+        double distance = preJumpPos.distanceTo(ctx.playerFeetAsVec());
+        //System.out.println("Distance to prepLoc = " + distance);
```
l1ving, line 642: mfw

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                    atDestTicks++;
+                } else {
+                    atDestTicks--;
+                }
+            } else {
+                atDestTicks = 0;
+            }
+            if (ctx.playerFeet().y < src.y) {
+                // we have fallen
+                logDebug("sorry");
```
l1ving, line 661: have something more helpful in the log

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+    //0.41999, 1
+    //0.75319, 2
+    //1.00133, 3
+    //1.1661, 4
+    //1.24918 highest
+    //1.17078, 6
+    //1.01555, 7
+    //0.78502
+    //0.48071
+    //0.10408
```
l1ving, line 687: :neutral_face:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+            } else {
+                return -0.0784;
+            }
+        }
+        double vel = (getFallVelocity(ticksFromStart - 1, jump) - 0.08) * 0.98;
+        if(vel < 0.003) {
+            return 0;
+        } else {
+            return vel;
+        } */
```
l1ving, line 732: :neutral_face:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        double moveMod = 0; // Default of stopping (not moving)
+        double sprintJumpBoost = 0;
+        Map<Input, Boolean> input = state.getInputStates(); // only used in the following if statements
+        if (input.getOrDefault(Input.MOVE_FORWARD, false) || input.getOrDefault(Input.MOVE_BACK, false) || input.getOrDefault(Input.MOVE_RIGHT, false) || input.getOrDefault(Input.MOVE_LEFT, false)) {
+            if (input.getOrDefault(Input.SNEAK, false)) {
+                moveMod = 0.3; // Sneaking
+            } else if (input.getOrDefault(Input.SPRINT, false) && input.getOrDefault(Input.MOVE_FORWARD, false)) {
+                moveMod = 1.3; // Sprinting
+                if(input.getOrDefault(Input.JUMP, false)) {
+                    sprintJumpBoost = 0.2; // Multi-directional sprint?
```
l1ving, line 849: consistency in comment space

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        return null;
+    }
+
+    // Statistic tools (to find the effectiveness of each solution)
+    /*
+    Vec3d prevMotionPredict = ctx.playerFeetAsVec();
+    Vec3d prevVelocityPredict = ctx.playerFeetAsVec();
+    Vec3d totalMotionDiff = new Vec3d(0, 0, 0);
+    Vec3d totalVelocityDiff = new Vec3d(0, 0, 0);
+    //*/
```
l1ving, line 892: mfw

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        }
+
+        if (moveDist > MAX_JUMP_WALK ||
+                type == JumpType.MOMENTUM ||
+                type == JumpType.EDGE_NEO) {
+            state.setInput(Input.SPRINT, true);
+        }
+
+        final double jumpMod = 0.2; //Amount to shift the jump location by (towards the destination) (0.2 is max if block is present)
+        double xJumpMod = jumpMod - Math.abs(simpleDirection.getXOffset()) * jumpMod; // perpendicular to offset
```
l1ving, line 921: I stg

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+        prevMotionPredict = prevVelocityPredict;
+        prevVelocityPredict = getFutureLocation(ctx, state, 2);
+        totalMotionDiff = totalMotionDiff.add(motionDiff);
+        totalVelocityDiff = totalVelocityDiff.add(velocityDiff);
+
+        logDebug("distToJump = " + distToJump + ", distToJumpXZ = " + distToJumpXZ + ", distFromStart = " + distFromStart + ", distFromStartXZ = " + distFromStartXZ + ", ticksFromStart = " + ticksFromStart);
+        logDebug("Player coords = " + ctx.playerFeetAsVec());
+        logDebug("MotionDiff = " + motionDiff + ", total = " + totalMotionDiff);
+        logDebug("VelocityDiff = " + velocityDiff + ", total = " + totalVelocityDiff);
+        //*/
```
l1ving, line 1013: :neutral_face:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+
+        if (ctx.playerFeet().equals(dest)) {
+            Block d = BlockStateInterface.getBlock(ctx, dest);
+            if (d == Blocks.VINE || d == Blocks.LADDER) {
+                return state.setStatus(MovementStatus.SUCCESS);
+            }
+        } else if (!ctx.playerFeet().equals(src)) {
+            // During a momentum jump the momentum jump will position us so just jump whenever possible (i.e. as soon as we land)
+            //distToJump < 0.17
+            //logDebug("JUMPING? " + Math.abs(futureLoc.x - (src.x + 0.5)) + " or " + Math.abs(futureLoc.z - (src.z + 0.5)));
```
l1ving, line 1023: :gorilla:

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
```diff
@@ -0,0 +1,1054 @@
...
+                }
+                // prevent jumping too late by checking for ascend
+                /*
+                if (moveDist >= 3) { // this is a 2 block gap, dest = src + direction * 3
+                    double distFromStart = ctx.playerFeetAsVec().squareDistanceTo(dest.x, dest.y, dest.z);
+                    if (distFromStart < Math.pow(0.7, 2)) {
+                        logDebug("UPDATE 2");
+                        return state;
+                    }
+                } //*/
```
l1ving, line 1042: :neutral_face:

l1ving (changes requested):

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: I added a comment to explain the format

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: I was used to having no spaces but I've seen other files in the project that have spaces so I was trying to add spaces but forgot sometimes. (sorry I'm new to programming)

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: Fixed. (I was using code from MovementParkour as my base)

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: How do linter's (first time I've heard of them sorry) help here?
(I was logging each jump so that I could see what the maximum jump was)

ZacSharp: A linter tells you when you write bad code

ZacSharp:
I can't do straight descends (but does the more difficult diagonal ones) <details> <summary>screenshot</summary> screenshot </details>

It tries to land on a covered block after a diagonal descend and descending by 1 instead of 2 would be the shorter path as well <details> <summary>screenshot</summary> screenshot </details>

The reliability is also less than 50% on diagonals for me. If either fails consistently or manages to do it consistently.

ZacSharp:
Descending by more than 2 blocks and even dodging obstacles while falling (e.g. for the dropper) and parkourPlace for the new jumps would also be nice to have.

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
CDAGaming: @leijurv You have to see this xD

CDAGaming:
"The reliability is also less than 50% on diagonals for me. If either fails consistently or manages to do it consistently."

This alone, to me, is enough that this should be marked as a draft until you can bump up the success rate of this, otherwise on paper this sounds nice

CesiumCs:
it also seems to just not be able to path normally with the settings enabled
https://youtu.be/hv3piRV-95E

Zephreo:
> I can't do straight descends (but does the more difficult diagonal ones)

I currently don't replace the functionality of the main Parkour system (straight parkour) except for 4 block jumps. So basically all straight parkour is ignored.

> The reliability is also less than 50% on diagonals for me. If either fails consistently or manages to do it consistently.

Yea I think some specific jumps it has trouble with. It would be nice if you were to encounter jumps that fail to note down the block configuration so that I can recreate them (the relative jump, and what blocks are in the way) for testing.

> Descending by more than 2 blocks and even dodging obstacles while falling (e.g. for the dropper) and parkourPlace for the new jumps would also be nice to have.

Descending is currently linked to the 'maxFallHeightNoWater' setting (although I haven't tested any jumps larger than the default).

Thanks for all the feedback!

leijurv:
this is high power levels

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: > A linter tells you when you write bad code

So, I'd just get red lines or something on all my code? :)

Zephreo:
When you realise you weren't checking for blocks in the way during descends... (The major cause of issues)

It made descends that land inside of blocks valid jumps. Now fixed! yay

src/main/java/baritone/pathing/movement/movements/MovementParkourAdv.java
Zephreo: fixed

leijurv:
I normally wouldn't even consider merging something like this but holy cr*p those videos are sick!

Good thing
* Brand new movement, doesn't touch MovementParkour and like add a billion options to it or anything like that

Bad thing (but easy to fix)
* In the default state (adv parkour = false), it allocates a BetterBlockPos in cost calculation. Can you make it pass an int,int,int and it only creates the BetterBlockPos after the check for if it's enabled?

Bad thing (very hard to fix)
* None of the other movements allocate any objects in cost calculation. This was a huge performance speedup, and it's why everything takes `int, int, int` instead of `BlockPos`. This is what allows the entire pathing to average a few hundred nanoseconds per movement considered, amortized over the entire process (including heap and hashmap). Everything that you think is allocating an object, isn't really. e.g. BlockStateInterface.get0(x,y,z) is just a bunch of nested hashmap lookups, it only returns IBlockStates that already existed, never constructs new ones.

HoWeVeR, from looking at how clutch the videos are, I'm tempted to relax the idea of no movements allocating objects... I can't imagine how absurdly difficult it would be to redo this 0 alloc.

Neutral thing
* This will probably take additional effort to make it work on newer versions of the game. The big thing being the sneak changes in 1.14.4. On the other hand, I don't really care that much about newer versions of the game. At worst, this option Just Exists and glitches out when you try to use it, at best it works fine Most Of The Time. Or maybe it would be easy because you could repaste the motion predictor...

Bad thing
* Copyright infringement (DMCA) on the pasted Mojang code in MovementPrediction. This is a no no. Need to find another approach, or rework it enough that it isn't a literal skid of hundreds and hundreds of lines of code. 

🥺 wanna post more vids doe? 🥺

Zephreo:
Not sure if MovementPrediction should be touched up more, I removed a lot of it.

I removed a few of the object allocations, most of the remaining ones now are the vectors, which are hard to remove :\

Zephreo:
It's hard to think of what to record..
outpu

ZacSharp:
Some more fails/weird things <details><summary>jumps in the wrong direction</summary> screenshot screenshot </details> <details><summary>why can't it do it one block lower?</summary> screenshot screenshot </details> <details> <summary>why do a neo here? That's an unnecessary risk</summary> screenshot </details> <details> <summary>unrelated blocks deciding between success and failure</summary> fails: screenshot does not fail: screenshot </details> <details> <summary>why insist on doing a unnecessarily difficult jump when we could do an easy one?</summary> screenshot screenshot </details>
Also speed at least halves the reliability.

src/api/java/baritone/api/utils/Rotation.java
```diff
@@ -80,6 +80,22 @@ public Rotation subtract(Rotation other) {
         );
     }
 
+    /**
+     * Returns a copy of the opposite rotation of the given rotation
+     *
+     * @return
```
ZacSharp, line 86: The CI build is failing on this. Maybe just put some documentation behind the `@return`?

Zephreo: I couldn't find why it was failing... thanks!

Zephreo:
> Some more fails/weird things

thanks!

> jumps in the wrong direction
> screenshot screenshot

wasn't able to replicate the first one (but some changes may have made relevant fixes)

The second one will probably never be fixed, due to the way baritone is made I can only output one 'result' for each direction and it always outputs the lowest cost jump (don't think this was working before but now fixed) and since both of the jumps are identical they have the same cost so whichever is processed first is outputted.

> why can't it do it one block lower?
> screenshot screenshot

Jumping one block lower would incur a 4 block drop (since you jump up ~1.2 blocks) this would cause fall damage and is therefore not considered a valid jump.
This behaviour is linked and can be changed with the maxFallHeightNoWater setting.

> why do a neo here? That's an unnecessary risk

Now fixed. (Straight descends weren't implemented since that was starting to approach default MovementParkour territory)

> unrelated blocks deciding between success and failure

could not replicate. But I think I fixed it (I assume the jump was an edge type and that you weren't sprinting when the block wasn't there)

> why insist on doing an unnecessarily difficult jump when we could do an easy one?

Now fixed. (Straight descends weren't implemented since that was starting to approach default MovementParkour territory)

> Also speed at least halves the reliability.

Speed is currently not accounted for. (except for dodging obstacles)
I'll look into the jump length differences for each level of speed and slowness.

only Jump Boost is somewhat accounted for (It should check higher blocks in the way)

CDAGaming:
> The second one will probably never be fixed, due to the way baritone is made
I don't see this as more then an excuse. As a PR, you have the ability to change how baritone is made if you feel it'll benefit and make this work better. The only person preventing this is yourself with that statement.

Zephreo:
> > The second one will probably never be fixed, due to the way baritone is made
> 
> I don't see this as more then an excuse. As a PR, you have the ability to change how baritone is made if you feel it'll benefit and make this work better. The only person preventing this is yourself with that statement.

To be able to make this fix/work I would have to modify how the AStarPathFinder processes movements not only would these changes affect the execution speed, I also would have to do a lot of research into exactly how to make the change.
Currently the only way I can think of fixing it would be to add an entry into the Moves Enum for each possible jump (currently 17 for each direction (not including ascends/descends)) an extra 64 entries (for all directions) is going to be messy. (Optimisations will only improve this slightly)

leijurv:
> To be able to make this fix/work I would have to modify how the AStarPathFinder processes movements not only would these changes affect the execution speed, I also would have to do a lot of research into exactly how to make the change.
Currently the only way I can think of fixing it would be to add an entry into the Moves Enum for each possible jump (currently 17 for each direction (not including ascends/descends)) an extra 64 entries (for all directions) is going to be messy. (Optimisations will only improve this slightly)

Yeah this one gets a pass, there is really no reasonable way for a move to return a list of possible destinations and costs. It *could* be done but would introduce unacceptable compromises to the speed of normal pathfinding.

ZacSharp:
> Jumping one block lower would incur a 4 block drop (since you jump up ~1.2 blocks) this would cause fall damage and is therefore not considered a valid jump

I lowered start *and* end block by one, so the jump itself is the same

> I assume the jump was an edge type and that you weren't sprinting when the block wasn't there

The only difference between those is that single block and obviously a few seconds in time. I started all of those paths directly at the start of the shown movement and immediately paused it, so I wasn't moving (or even sprinting) before the jump in both cases. The assumption about the jump being an edge jump is half true, because somehow baritone decided to do a normal jump without the extra block and an edge jump with the extra block (or the other way round if I didn't remember properly)

I'll also test all of those again now

ZacSharp:
The neos (the ones I showed) completely stopped working.
The single unrelated block now decides between the left and right jump-targets instead of between success and failure.
The others work fine now.
Also I think considering potion effects is important because tipped arrows are actually being used on anarchy servers for exactly that reason (they break movement related tools)

How difficult would it be to not movement-cancel when continuing with a straight descend or other movements where we can't fall off?

Zephreo:
> The neos (the ones I showed) completely stopped working.

Oops, my bad forgot to test them after making changes... (fixed that now)

> The single unrelated block now decides between the left and right jump-targets instead of between success and failure.

The unrelated block ties in with the single output thing. The unrelated block is a valid jump with a lower cost than the jump that leads in the correct direction.

> The others work fine now.

Yay! :)

> Also I think considering potion effects is important because tipped arrows are actually being used on anarchy servers for exactly that reason (they break movement related tools)

I have added some preliminary improvements that should prevent a lot of overshoots with speed. Although you may still fall off when preparing for an edge jump.

> How difficult would it be to not movement-cancel when continuing with a straight descend or other movements where we can't fall off?

I reduced the momentum cancel to 2 ticks. 

~~I assume you mean the moment it lands it instantly looks back and you would either:~~
 1. ~~want excess momentum left there for another jump or~~ 
 2. ~~spend less time reducing momentum unnecessarily. (e.g. friction/drag can stop us anyway)~~

~~For the first option (using excess momentum on another jump), It would probably be quite difficult, I'm not sure if I can guarantee that the next jump will be another parkour jump unless I chain them under the same motion (which would be hard).~~

~~For the second option (saving time on unnecessary momentum reduction). It would be annoying implementing it for the case that there is an extra block to slide into.
And for the case where we would land on the destination block anyway, It wouldn't bring too many benefits (2-3 ticks is only 100-150ms).~~
~~I will probably attempt to use MovementPrediction for reducing the required checking/confirmation time to 1 tick.~~ 

(ignore that) 
Check if the lower cancelling (2 ticks) is enough. I don't think I quite get what you mean (It doesn't seem too slow to me)

ZacSharp:
I mean if the path continues in the direction of movement and the next movement is *not* a jump we could simply keep running, because our direction is already correct and there is a solid block were are moving onto, so we can't fall off.
Currently it cancels movement down to zero to continue even when landing on a flat plane. <details><summary>examples of unnecessarily canceling momentum</summary> screenshot screenshot screenshot </details>

The neos are still not working for me and I noticed that 4 block jumps are not working since cdac86302901bfbbe050d9cd675da725faa053c0 for me either<details> <summary>It's doing those again</summary> screenshot </details> <details> <summary>isn't that what this pr should eliminate?</summary> screenshot </details>

Neither speed II nor slowness could make it fail on my testing parkour

ZacSharp:
Nice, almosr everything working. Reliability is at least 90%, except for the momentum jumps without a wall behind, for which I measured 18% (9 out of 50) and the neos.
Problems are
* north neos not working
* slabs being treated as full blocks (tries to ascend from them or to jump from 0.5 locks above their edge)
* the usual "wtf why does it have to do an intermediate jump? It could just do it in one!"

For the last one I thought about maybe rotating a parkour movement if it does not find a jump in its intended direction. So if it checks a parkour east and there is nothing to jump on to the east it continues searching south, then west and north before returning nothing, so we can have up to 4 results in one direction, if there is nothing in the other directions. That would be a weird workaround but I think in most cases it is sufficient, as you rarely get 4 possible jumps that really make a difference in a real parkour situation. I don't know what this would do to the performance though.

Zephreo:
> For the last one I thought about maybe rotating a parkour movement if it does not find a jump in its intended direction. So if it checks a parkour east and there is nothing to jump on to the east it continues searching south, then west and north before returning nothing, so we can have up to 4 results in one direction, if there is nothing in the other directions. That would be a weird workaround but I think in most cases it is sufficient, as you rarely get 4 possible jumps that really make a difference in a real parkour situation. I don't know what this would do to the performance though.

Although this is a good idea, it wouldn't work. The cost calculations are deterministic and therefore for the given circumstance always return the same result (the one with the lowest cost).
2 ways to make it work.
Some kind of memory of what options were already chosen (difficult, not sure how to do)
Randomness in the result (easier).

The problem with combining your solution with randomness is that it is rather slow because you are recalculating a the movement multiple times and possibly for no gain, there is not always two lowest-cost results.

I was thinking of implementing either randomness or a goal heuristic (taking into account how far the goal is from the solutions destination). Both have some problems that I still have to work out.

(for the goal one, it's first deciding whether to do it to only the lowest-cost results, or to a range of results, or all of the results and whether it influences the external cost, a hidden internal cost, or just direct comparison (if lowest cost clashes than compare the result closest to the goal)) I've also been a little busy lately so I might take a week or two to set it up.

Zephreo:
That should fix most of the problems where it chose inefficient paths. Sorry CDAGaming, I was actually able to think of a solution...

Momentum jumps are still super unreliable, so that still needs to be fixed.

Zephreo:
I'm not sure if I can improve 4 block angled (1 block to the side) reliability much further should I remove them entirely (angled 4 blocks with no block behind) or try more stuff (would they be used often)?
Tested reliability of 20% (4/20) for angled vs 85% (17/20) for flat 4 blocks (both no block behind).

Zephreo:
parkour-place

src/main/java/baritone/pathing/movement/Moves.java
```diff
@@ -322,6 +322,54 @@ public Movement apply0(CalculationContext context, BetterBlockPos src) {
         public void apply(CalculationContext context, int x, int y, int z, MutableMoveResult result) {
             MovementParkour.cost(context, x, y, z, EnumFacing.WEST, result);
         }
+    },
+
+    PARKOUR_ADV_NORTH(0, 0, 0, true, true) {
```
leijurv, line 327: BTW can you change this to instead be the max distance along north? Because of `if ((newX >> 4 != currentNode.x >> 4 || newZ >> 4 != currentNode.z >> 4) && !calcContext.isLoaded(newX, newZ)) {` in AStarPathFinder

It **probably** is okay, becuase parkour will fail if the destination is just air, but it will probably help prevent potential weird behavior. Maybe has not come up because I imagine you aren't testing this with long-distance pathing that much, and mostly just short range?

leijurv:
@Zephreo Hey, I just wrote up a simple way to provide multiple movement results without allocating new objects per-movement or per-movement-destination

https://github.com/cabaletta/baritone/commit/4a489d29dee3e1eea8969c7e1154e0338af24518

Just a suggestion. I would be okay with this approach (so, I "pre-approve" this, or something very similar to it). 

Don't be afraid by the massive diff in AStarPathFinder, it literally is just adding `while (res != null) {` up top, and `res = res.getNext(); \n }` down below. 

It reuses the "chain" of MutableMoveResult objects so that they are only allocated once at the beginning of path calculation. 

The overhead caused by this change should be essentially zero to normal path calculation. There will be an extra `if` not taken in `reset`, and `getNext` will be called and return null to make the while loop bail out early. Two extra branches every movement is not an issue. Hell, adding a movement disabled by default already adds more than that. So I'm not worried.

I hope this helps and I hope you have some fun with it :)

leijurv:
oops, just had a 🗿 🧠  moment, also do this https://github.com/cabaletta/baritone/commit/f0082990e955ef35b02a53437728b57f9a89958c

Zephreo:
> @Zephreo Hey, I just wrote up a simple way to provide multiple movement results without allocating new objects per-movement or per-movement-destination
> 
> [4a489d2](https://github.com/cabaletta/baritone/commit/4a489d29dee3e1eea8969c7e1154e0338af24518)
> 
> Just a suggestion. I would be okay with this approach (so, I "pre-approve" this, or something very similar to it).
> 
> Don't be afraid by the massive diff in AStarPathFinder, it literally is just adding `while (res != null) {` up top, and `res = res.getNext(); \n }` down below.
> 
> It reuses the "chain" of MutableMoveResult objects so that they are only allocated once at the beginning of path calculation.
> 
> The overhead caused by this change should be essentially zero to normal path calculation. There will be an extra `if` not taken in `reset`, and `getNext` will be called and return null to make the while loop bail out early. Two extra branches every movement is not an issue. Hell, adding a movement disabled by default already adds more than that. So I'm not worried.
> 
> I hope this helps and I hope you have some fun with it :)

I think I have solved this problem by emulating the main function of AStarPathFinder by comparing costs plus the goal heuristic. 

`Goal goal = context.baritone.getCustomGoalProcess().getGoal();`
`double resHeuristic = goal.heuristic(destX, destY, destZ);`
        
`if (res.cost + resHeuristic < curLowestCost.cost + lowestCostGoalHeuristic) {`

leijurv:
That absolutely does not solve that problem lol, there are multiple possible movements from the current node that could be on the optimal path.

If you could choose and know which movement would be used in the final path, then we wouldn't need A* at all!

Zephreo:
> That absolutely does not solve that problem lol, there are multiple possible movements from the current node that could be on the optimal path.
> 
> If you could choose and know which movement would be used in the final path, then we wouldn't need A* at all!

Yes but it seems to do a good enough job at choosing the correct path? (At hopefully less complexity?)

leijurv:
> Yes but it seems to do a good enough job at choosing the correct path?

??

Can you just try it this other way where it actually feeds the movements into the A* algorithm instead of badly emulating it in place with a heuristic??

> Paths chosen during my testing with this method have been correct all the time so far
Try it in the spawn area of an anarchy server, I guarantee it'll miss possible ways out.

Even for something as simple as a path that has to backtrack for a little section, wouldn't it pick the worst backwards facing jump, just as much as it picks the best forward facing jump? Like backwards = away from the goal?

> At hopefully less complexity

Don't worry, baritone's core a* loop is a big boy, it can handle it :)

Zephreo:
I think since all of the paths that it compares are in the same general direction (north, south, etc.) it won't encounter many times where one path goes towards and another path goes away (unless they are both edge (high angle) jumps in opposite directions).

I will try out your method and if it works I'll switch to that instead. 
(Sorry don't know too much about A* and in my eyes, it was way better than the previous method)

leijurv:
> Sorry don't know too much about A* and in my eyes, it was way better than the previous method

Yeah sure so what's going on here is that A* is perfectly okay with the idea of there being many possible options to take from a given node. This is how all the different variety of movements that we have today work. It's perfectly reasonable for there to be many possible parkour jumps that we could take from a given starting spot. The algorithm can handle it just fine.

Now imagine that in the case of of a goal to the north, we need to backtrack a bit to the south. Please correct me if I'm wrong but I believe that your thing here would be biased towards taking long jumps to the north, and short jumps to the south. This could cause it to not consider a long jump to the south, even though that could conceivably be the only way to get to a goal to the north. In normal A*, it would consider all those options, but be **biased** towards the short jump to the south, but it would **eventually** get to the long jump to the south as it works its way through the binary heap of nodes in the open set. I know sounds like contrived scenario but it's perfectly plausible and I've seen backtracking need to happen quite commonly.

ZacSharp:
> Yes but it seems to do a good enough job at choosing the correct path?

It took me like 5 seconds to think of a scenario where it doesn't find a path at all because it chooses a jump into a dead end where it can't get out of rather than the only jump actually leading to the goal

SuperGamingBros4:
BUG: follow completely breaks when you use allowparkouradv and keeps on giving, "Pathing exception: java.lang.NullPointerException" until you move baritone manually into the follow radius. This should be fixed before merging.

ZacSharp:
There is also an occasional `NullPointerException` after `#cancel` and weird behavior when I told it to climb up a mountain. When I turned either `allowParkour` or `allowPlace` off, it just climbed the mountain like a normal player would do it, but with both enabled it started pillaring up to the height of the mountain and then failed pathing.

SuperGamingBros4:
I found another bug, if you make it place a block while jumping, then neo off of that placed block it doesn't place the blockl. In fact it doesn't even consider placing the block in the it's path.

ZacSharp:
This causes the `Movement became impossible during calculation` error *really* frequently. Either `MovementParkourAdv` outputs impossible movements as possible during calculation or it rejects possible movements as impossible while walking or (dunno if that can happen, but if it can that's an issue because a movement may not output a change in coordinates that can be achieved by another movement) another movement accepts the jumps and then decides they are impossible

leijurv:
```

    private Movement runBackwards(BetterBlockPos src, BetterBlockPos dest, double cost) {
        for (Moves moves : Moves.values()) {
            Movement move = moves.apply0(context, src);
            if (move.getDest().equals(dest)) {
                // have to calculate the cost at calculation time so we can accurately judge whether a cost increase happened between cached calculation and real execution
                // however, taking into account possible favoring that could skew the node cost, we really want the stricter limit of the two
                // so we take the minimum of the path node cost difference, and the calculated cost
                move.override(Math.min(move.calculateCost(context), cost));
                return move;
            }
        }
        // this is no longer called from bestPathSoFar, now it's in postprocessing
        Helper.HELPER.logDebug("Movement became impossible during calculation " + src + " " + dest + " " + dest.subtract(src));
        return null;
    }

```

Ah, it looks like this also needs to be modified in order to support multi-destination moves. Maybe `Moves.apply0` needs to instead return a `Collection<Movement>` and this function needs to filter down to any that end at the correct location?

Zephreo:
The two big problems that I currently know of are:
- Movements becoming impossible after calculation when there are a variety of different options to choose from that go in the same direction.
- Movements failing to sprint on jumps where sprint is required. This occurs when baritone is currently in the process of calculating the next segment of the path on another thread.

The first one seems to be due to the function that leijuv pointed out, but the second one I still have no idea why it happens.

Zephreo:
Not sure if the changes in Moves is what you want leijurv. Maybe I should swap it out for another function that calls Collections.singletonList() instead of manually doing it for each movement?

leijurv:
Yeah you're right. Maybe add a new `public boolean canHaveMultipleDestinations() { return false; }` in the enum, and your advanced parkour ones would override it to true, and a new method `public Collection<Movement> getMultiDestination() { throw new UnsupportedOperationException(); }` that is overridden again by advanced parkour moves?

That way there would be no cost to the case when advanced parkour is disabled :)

ZacSharp:
I found the cause for the weird stack-up-and-fail behaviour: timeouts. Somehow using `AllowParkourAdv` together with `allowPlace` slows pathing down so much that it either hits the pathing timeout and just takes the (not necessarily good) best path so far, if that is long enough, otherwise it fails or it hits the failure timeout and just fails finding any path.

Zephreo:
I wasn't able to get paths longer than 324,000 movements considered (on an amplified world up very large mountains) which took 2200ms with both allowPlace and ParkourAdv enabled, could you send a screenshot of the general path?

I wasn't able to replicate the stacking unless stacking was a pretty optimal route.

ZacSharp:
> I wasn't able to get anywhere near 24 sec with a 1,600,000 considered movements segment,
could you send a screenshot of the general path (start/end/seed)?

Maybe I should have mentioned that I bumped up all timeouts to some arbitrarily high value (11.5 days is the lowest)
The seed of my world is 0 and I let it path from my testing parkour (with a lake underneath) to the top of the next mountain
I will edit in a screenshot later today

ZacSharp:
* Can you please give me some more information on how you reproduce the issue of not sprinting while calculating?
* Between multiple 4b jumps in +x or +z direction it movement cancels until the movement times out and it recalculates the whole path.
* It tries to jump through walls in some cases, I noticed that with a -1 0 2 with blocks at 0 1 2 and 0 2 2, but there are some more.
* Parkour place seems to only work with flat jumps and neither with ascends nor decends.

Zephreo:
Edge jumps currently don't work that well with sprinting will work on that soon.

> It tries to jump through walls
Does it jump directly into a wall? or just clip the corner of a wall?

> Parkour place seems to only work with flat jumps and neither with ascends nor decends.
Currently intended might add this in later

> Can you please give me some more information on how you reproduce the issue of not sprinting while calculating?
It doesn't happen consistently and it may just be my computer I will be able to test this more on another computer in a week or so (Which is when I will look into the problem more/find how to reproduce).
Although the same problem (not sprinting) seems to also occur with my edge jump changes (not while calculating, just normally).

ZacSharp:
Sorry for taking so long to answer<details><summary>It jumps directly into a wall</summary>screenshotscreenshot</details><details><summary>Decends sometimes ignore blocks in their way</summary>screenshot</details>

ZacSharp:
All issues from my last 2 comments fixed (didn't check further back).
Neos don't work (doesn't output them) when it could descend without water.
It fails Neos when trying to place the target block against the pillar it is jumping around.

ZacSharp:
Neos close to ground still don't work. <details><summary>It fails this jump since quite a while ago, but I couldn't reproduce it earlier</summary></details> Since the 2 traverses (the 3rd makes it more likely) are required for it to fail I guess has to much momentum from the ascend and traverses and thus doesn't jump far enough.
 
And just a question: When I tested this without freeLook I noticed it turns around a lot, especially for Neos with 2 times 90° and I was wondering if that makes it harder (every human I saw doing Neos didn't noticeably turn their head).

Zephreo:
Sorry, but I think I will have to close this pull request at least for now. This is the first project I've tried to do semi-professionally (since it's possibly going to be used by people other than me) and it's rather draining. 
The following are all of the issues I know of.
The major bug, Some paths fail for seemingly no reason when close to other jumps, still requires a few more days of debugging to fix. 
The accuracy issue, some jumps fall when the starting position is in certain areas, this would require some kind of weird trickery to find the best sequence of key presses to get to a position exactly which I can't really do. Most of the jumps are within acceptable accuracies anyway >90% and I've added a setting in my dev branch for anyone who wants fake 100% accuracy (unnoticeable teleporting). Correct me if I'm wrong but I don't think this project would allow me to call ctx.player().move() directly instead of using key presses.

I may work on this in my own time without the pressure of a pull request.

> When I tested this without freeLook I noticed it turns around a lot, especially for Neos with 2 times 90° and I was wondering if that makes it harder (every human I saw doing Neos didn't noticeably turn their head).
Human players use strafing to complete a neo. for the sake of simpler/elegant code I use player rotation, it is a much harder method for humans to follow the near-perfect rotations although for baritone it's like 3 lines of code (using the APIs).

boehs:
I do wish someone would keep working on this. its a great pr and just needs some bug fixes and other improvements

HexCodeFFF:
bumping because I'd love to see this perfected and merged

CesiumCs:
He's explained why its closed and not getting any work. I believe you could still build it to play around with it, but unless someone else chooses to take on this pr, it'll stay closed.

HexCodeFFF:
right i should have made that more clear and I wasn't suggesting OP needed to fix it but I hoped that someone who has more knowledge can upgrade it to 1.19 or clean it up good enough to be added as an experimental option
I tried but I just don't know enough about java modding to do it myself
Since this pull is just some bugfixes, upgrading, and polishing away from a MAJOR feature improvement I think its worth bumping

Zephreo:
The problem with the current implementation is that it cannot accurately detect if a parkour jump is possible, so it works really great on very open parkour jumps with few obstacles but struggles with any real-world tests like climbing mountains where there are lots of blocks that only slightly obstruct the player.

I'm not really in a position where I have the time/energy to fix this but am happy to help anyone that wants an explanation.

ZacSharp:
> can upgrade it to 1.19

This is a common misconception. Pullrequests are merged on the lowest supported version (1.12.2 / master) and are then merged upwards.

PR #2216, opened 2020-11-22
Added option to prevent baritone from breaking tools

recursivetree:
I added an option to prevent baritone from breaking tools. You can toggle it with the setting "allowBreakTool". I think this is valuable, especially if you use mending tools


recursivetree:
Well, I haven't seen that there is already a pull request for this, so I'm gonna close this one

Issue #2247, opened 2020-12-08
How to create a visualization of cached chunks?


mvrozanti:
## What do you need help with?
I want to use the cached chunks to create an image to visually represent what chunks have been visited. Is this currently possible, using the *.bcr files generated?

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

l1ving:
Yea, this was possible at some point. I don't think it was ever merged into master due to stability?

https://www.youtube.com/watch?v=FyBP_IUkY9o

mvrozanti:
Sorry, that's not what I meant. 

I want to serialize the chunks to an image file, not render it in-game. The functionality should work as described in this [Impact issue](https://github.com/ImpactDevelopment/ImpactIssues/issues/2713)

mvrozanti:
Like, I'd imagine that if we take the top-side texture of each block that is immediately under the highest air block in a given X,Z coordinate (or if it _is_ the highest block at that location), we can stitch an image that represents it correctly. 

We could worry about block rotation in the future, if that works.

mvrozanti:
Only problem is, I have no idea how to read .bcr files

l1ving:
Baritone doesn't have a way to do this

mvrozanti:
I thought bcr stands for baritone cached region, does it not?

l1ving:
idk lol, but Baritone doesn't have a way to *visualize* it as far as I know

mvrozanti:
Can we extract every `BlockPos` from a bcr file? Because if so, we _can_ visualize it with a little effort

l1ving:
maybe, the code would probably help more

mvrozanti:
I'm gonna try and see how to read a bcr and reply back here

mvrozanti:
```
        ICachedRegion cr = ctx.worldData().getCachedWorld().getRegion(0, 0);
```

This is pretty much it. We just need to:

1) List every cached region for current server
2) For each region, extract every block contained in it
3) For every block in the region map the image for that block in the correct relative position
4) Save that to a png

l1ving:
I mean, would be nice to have

mvrozanti:
Textures can be found at .minecraft/versions/1.15.2.jar/assets/minecraft/textures/block/

ZacSharp:
for me #66 looks similar to this

mvrozanti:
[...]

PR #2257, opened 2020-12-12
ToMap Command

mvrozanti:
Proof of concept implementation of issue #2247

### DIFF
--- a/src/main/java/baritone/command/defaults/ToMapCommand.java
+++ b/src/main/java/baritone/command/defaults/ToMapCommand.java
@@ -20,9 +20,12 @@ package baritone.command.defaults;
 import baritone.Baritone;
 import baritone.api.IBaritone;
 import baritone.api.cache.ICachedRegion;
+import baritone.api.cache.ICachedWorld;
 import baritone.api.command.Command;
 import baritone.api.command.argument.IArgConsumer;
 import baritone.api.command.exception.CommandException;
+import baritone.cache.CachedWorld;
+import baritone.cache.WorldScanner;
 import net.minecraft.block.AirBlock;
 import net.minecraft.block.BlockState;
 import net.minecraft.client.Minecraft;
@@ -42,6 +45,8 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                } catch (IOException e) {
+                    System.out.println("Error setting color for " + blockId);
+                    e.printStackTrace();
+                }
+            }
+        }
+        logDirect("Finished coloring");
+        try {
+            ImageIO.write(result, "png", new File("/home/nexor/output.png"));
```
ZacSharp, line 136: don't hardcode output directory
Better save it to screenshots or the world's baritone folder

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    public Color averageColor(BufferedImage bi, String id) {
+        if (!cachedColors.containsKey(id)) {
+//            System.out.println(id);
```
ZacSharp, line 144: don't leave debug output as comments

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+                cachedColors.put(id, new Color(sumr / num, sumg / num, sumb / num));
+            } catch (Exception e) {
+                cachedColors.put(id, new Color(Color.TRANSLUCENT));
+            }
+        }
+        return cachedColors.get(id);
+    }
+
+    private static BufferedImage getTextureForBlockId(String blockId) throws IOException {
+        String base = "/home/nexor/.minecraft/versions/1.15.2-OptiFine_HD_U_G5_pre1/1.15.2-OptiFine_HD_U_G5_pre1/assets/minecraft/textures/block/";
```
ZacSharp, line 167: another hardcoded path

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+                });
+            }
+        };
+    }
+
+    // http://jens-na.github.io/2013/11/06/java-how-to-concat-buffered-images/
+    // /home/nexor/.minecraft/versions/1.15.2-OptiFine_HD_U_G5_pre1/1.15.2-OptiFine_HD_U_G5_pre1/assets/minecraft/textures/block
+    private void concatImages() {
+
+    }
```
ZacSharp, line 207: why is there an empty method?

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    public Stream<String> tabComplete(String label, IArgConsumer args) throws CommandException {
+        return Stream.empty();
+    }
+
+    @Override
+    public String getShortDesc() {
+        return "null";
```
ZacSharp, line 216: maybe put something more informative here

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public List<String> getLongDesc() {
+        return Arrays.asList("null");
```
ZacSharp, line 221: same here

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+public class ToMapCommand extends Command {
+
+    private HashMap<String, Color> cachedColors;
+
+    public ToMapCommand(IBaritone baritone, String... names) {
+        super(baritone, "tomap");
```
ZacSharp, line 52: either forward the names parameter or better remove it but don't leave it unused

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+        cachedColors = new HashMap<>();
+    }
+
+    @Override
+    public void execute(String label, IArgConsumer args) throws CommandException {
+        ctx.worldData().getCachedWorld().reloadAllFromDisk();
```
ZacSharp, line 58: Shouldn't it as well repack everything it can repack so the data is more up to date?

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,223 @@
...
+                    String id;
+                    if (bs == null) {
+                        id = "air";
+                    } else {
+                        try {
+                            Matcher m = p.matcher(bs.toString());
+                            if (m.find())
+                                id = m.group(1);
+                            else
+                                throw new Exception("Damn");
```
ZacSharp, line 81: Maybe give more information on what caused the exception

ZacSharp (changes requested):

ZacSharp:
I know this is a proof of concept but you should iron out at least the hardcoded paths so others can test it

mvrozanti:
I don't know how to take them out and turn them generally working

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: How?

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: This is non-important.

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: How can we fix this?

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: Just so I remembered that I need to do that at some point

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: The issue is informative enough for now

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: Same above

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: Agreed, will fix in the future

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: I don't know what repack means

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: Agreed

mvrozanti:
Can we focus on real problems on the issue assigned to this PR plz?

ZacSharp:
`net.minecraft.client.Minecraft.getMinecraft().gameDir.toPath().resolve("screenshots")` for the screenshots directory and check `baritone.cache.WorldProvider#initWorld` for how to find the place where world/server specific baritone information is stored

Ok, what problems to you mean?

src/main/java/baritone/command/defaults/ToMapCommand.java
ZacSharp: https://github.com/cabaletta/baritone/pull/2257#issuecomment-743911392

mvrozanti:
> net.minecraft.client.Minecraft.getMinecraft().gameDir.toPath().resolve("screenshots") for the screenshots directory and check baritone.cache.WorldProvider#initWorld for how to find the place where world/server specific baritone information is stored

Thanks I'm gonna work on that.

> Ok, what problems to you mean?
I mean mostly two things: 
1) how do we stitch the regions together? 
2) how do we efficiently serialize the image?

Given the code we already got.

src/main/java/baritone/command/defaults/ToMapCommand.java
ZacSharp: Don't know if https://github.com/cabaletta/baritone/pull/2257#issuecomment-743911392 is enough here but I think it should be possible to find the current version's assets

src/main/java/baritone/command/defaults/ToMapCommand.java
ZacSharp: This is the ingame help, for those that didn't read the issue (hopefully that is everyone soon)
I personally think it's better to do stuff like this before forgetting it, but maybe you are better in remembering stuff than I am

src/main/java/baritone/command/defaults/ToMapCommand.java
ZacSharp: It takes the chunks in render distance and caches them, so basically whenever you use the cache and have enough time, do a repack so you don't let the cache fool you

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: I didn't even stop to think about a description. I think this should be the last thing to be written to this file.

mvrozanti:
> `net.minecraft.client.Minecraft.getMinecraft().gameDir.toPath().resolve("screenshots")` for the screenshots directory and check `baritone.cache.WorldProvider#initWorld` for how to find the place where world/server specific baritone information is stored
> 
> Ok, what problems to you mean?

image

I'm missing this method.

ZacSharp:
Just noticed you based this on 1.15. usually pulls are based on master, so searched there and maybe this has changed from 1.12 to 1.15

ZacSharp:
It is `net.minecraft.client.Minecraft.getInstance().gameDir.toPath().resolve("screenshots")` now

l1ving:
Yes, pulls should be done to master and backported upwards

mvrozanti:
I'll work on these reviews but knowing there's still a long ways to go before merging to master

src/main/java/baritone/command/defaults/ToMapCommand.java
mvrozanti: The latter would be best, since not everyone has an unzipped jar lying around in their versions directory lol

mvrozanti:
I'm getting some very weird behavior. Only region x=0,z=0 doesn't return null, even though reloadAllFromDisk is being called and there are countless other region files on disk.

mvrozanti:
I'm getting some very weird behavior with this call:
`ICachedRegion cr = ctx.worldData().getCachedWorld().getRegion(rX, rZ);`

Only region x=0,z=0 doesn't return null, even though reloadAllFromDisk is being called and there are countless other region files on disk. I'm kinda stuck.

EDIT:

What's more weird, is that I get the following messages in the log after calling reloadAllFromDisk(): 

```
[00:31:37] [main/INFO]: [STDOUT]: Loading region 0,0 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache
[00:31:38] [main/INFO]: [STDOUT]: Loaded region successfully in 317ms
[00:31:38] [main/INFO]: [STDOUT]: Loading region 0,-1 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache
[00:31:38] [main/INFO]: [STDOUT]: Loaded region successfully in 303ms
[00:31:38] [main/INFO]: [STDOUT]: Loading region -20,46 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache
[00:31:38] [main/INFO]: [STDOUT]: Loaded region successfully in 48ms
[00:31:38] [main/INFO]: [STDOUT]: Loading region -21,46 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache
[00:31:38] [main/INFO]: [STDOUT]: Loaded region successfully in 24ms
[00:31:38] [main/INFO]: [STDOUT]: Loading region -1,0 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache
[00:31:38] [main/INFO]: [STDOUT]: Loaded region successfully in 314ms
[00:31:38] [main/INFO]: [STDOUT]: Loading region -1,-1 from disk /home/nexor/.minecraft/baritone/<my-server-ip-address>/DIM0/cache


```

So, not only getRegion is returning null for regions that have been (apparently) loaded succesfully, not all regions stored on disk are being read succesfully. What am I doing wrong here?

EDIT2: I must add, `~/.minecraft/baritone/<my-server-ip-address>/` is a symbolic link.

ZacSharp:
I think `reloadAllFromDisc` reloads all regions that are already loaded and `tryLoadFromDisk` or `getOrCreateRegion` is more what you want.

mvrozanti:
Cool let me try that

mvrozanti:
image
Colors actually work now.

mvrozanti:
Colors are now fairly consistent, thanks @ZacSharp .

But `tryLoadFromDisk` isn't exactly working... I need to test further but I think it's returning an empty CachedRegion.

ZacSharp:
tryLoadFromDisk's return type is void

mvrozanti:
> tryLoadFromDisk's return type is void

Yes. What do you mean?

ZacSharp:
You said you think it returns an empty region, so I (indirectly) said it returns nothing

mvrozanti:
I edited my previous message with what I actually meant.

mvrozanti:
I have no idea how to proceed.

ZacSharp:
And I can't test it because my hard drive is full

mvrozanti:
Well, we tried.

ZacSharp:
I haven't given up yet, it's just a bit more difficult

mvrozanti:
I kinda have. There's no input from the repository admins and baritone should definitely have its bugs fixed before prioritizing new features.

ZacSharp:
Can you please rebase this onto master so I can test it without gradle needing to cache new stuff?

mvrozanti:
[Sure](https://github.com/mvrozanti/baritone/tree/tomap-command-from-master). You're definitely going to have to tweak a few things, such as the retrieving the screenshots directory.

scorbett123:
I am not entirely sure, however I believe that there is a setting about whether to save cached chunks to the disk,  have you enabled it?

mvrozanti:
@scorbett123 AFAIK, no. But I have 2GB of data under `~/.minecraft/baritone/<my-server-ip>`. I think that's enough proof that caching to disk is happening. But I can't get what's cached to disk to runtime memory.

mvrozanti:
Funny thing is, somehow region 0,0 was loaded to runtime memory. I'm still not sure why it worked for this region and not others. The result is pretty neat:

scorbett123:
Sorry, I wrote that late last night, was thinking of the wrong project.

scorbett123:
Wait, why are you trying to merge into the 1.15.2 branch?

mvrozanti:
> Wait, why are you trying to merge into the 1.15.2 branch?

I wasn't trying to merge yet, I just made a PR for convenience for people to give feedback. Also, I'm testing for 1.15.2 only

ZacSharp:
Was the rebased branch outputting an image before the rebase? Just so I know what I screwed up by replacing the image loading and what was already broken.

mvrozanti:
No, it wasn't. I actually didn't even try to run it.

ZacSharp:
Not even before the rebase?

mvrozanti:
I edited my answer.

ZacSharp:
And I deleted my comment

mvrozanti:
I'd advise to test this on 1.15.2 though. I've only actually tested in that version.

ZacSharp:
Differences are only 5 replacements, so they should be negligible.
Bigger problem is that I don't have an assets folder somewhere and I'm testing in singleplayer.
I managed to get 6 transparent images now

mvrozanti:
>  Bigger problem is that I don't have an assets folder somewhere and I'm testing in singleplayer.
Just unzip the jar for the version you're running located in `.minecraft/versions` and edit the string in `ToMapCommand#getTextureForBlockId`

ZacSharp:
That one stings a bit but I don't think it is important right now: `Failed to find block id for minecraft:air`

mvrozanti:
I don't think you have the right version. Can you please, please test this branch? We're making this waaaaaaaay harder than it has to be.

ZacSharp:
This has to be ported to 1.12.2 at some point anyway, so why not do it now?

mvrozanti:
Because it barely works on this version!

mvrozanti:
> Failed to find block id for minecraft:air

If you're getting this message, you are most certainly not using the latest version of the code I wrote, or the rebased one.

ZacSharp:
You changed something since https://github.com/cabaletta/baritone/pull/2257#issuecomment-744074936?

mvrozanti:
I don't know man. Can't you check the git log or just download the latest?

ZacSharp:
`git pull` outputs "up to date" for both branches

mvrozanti:
This is getting out of hand..... Please:

1) Download my fork
2) switch to *one point fifteen point two* branch
3) build and move `dist/baritone-api-1.5.3.jar` to `.minecraft/libraries/cabaletta/baritone-api/1.5.3`
4) it will work

ZacSharp:
1. Done
2. Done by accident
3. Uh, I'm always using forge or a dev env, so I have to set up a new install for that
4. We'll see, didn't work from dev env

mvrozanti:
I edited my previous answer in case that doesn't work.

ZacSharp:
If you really want to you can send me a precompiled jar, but getting identical files from different places won't help.

mvrozanti:
Please, never run a precompiled jar of unknown origin. 

You say you have the latest version, but this error:


> Failed to find block id for minecraft:air

Says otherwise.

ZacSharp:
That error is gone but the screenshots are still transparent. Haven't tested a normal install yet because the game crashes when I try to start it.

mvrozanti:
Do you have discord or something?

ZacSharp:
No

mvrozanti:
Well, at the moment I can only tell you to do the necessary steps for getting the branch in this PR, which is the version that generated the picture I posted above.

ZacSharp:
Ok, so apparently I can't test this in a production enviroment because the baritone json version crashes and the forge jar is generated by the build step that just crashes. I'll now go and try to get this working without a working start. Thank you for trying to help me.

mvrozanti:
Thank **you** for trying to help us get this feature

ZacSharp:
Just a few things from the last 20 minutes:
1. How do I find out if a region is empty?
2. Player873 is banned on one of my two testing servers (on topic I know)
3. ```java
     private BufferedImage getTextureForBlockId(String blockId) throws IOException {
         String base = "/home/mcbw/.minecraft/versions/1.12.2/1.12.2/assets/minecraft/textures/blocks/";
         int x = 0/0;
         System.out.println(blockId + " -> " + String.format("%s",x));
        ...
   ```
does not cause any exception or output
4. This causes weird errors when switching server without restarting

1 and 3 are most important rn

mvrozanti:
1) I consider a region to be empty if every single block in that region (which means the blocks from northwest corner to southwest corner in all y levels) returns null.

2) I have no idea what you're talking about

3) I'm not sure what you're trying to achieve with that piece of code

4) I haven't tested switching servers without restarting

ZacSharp:
2. Not important, just some random username being banned on some random cracked server
3. I tried to get some output to see if and when it is executed. The x=0/0 was would have caused a RuntimeException, if it had been executed.

mvrozanti:
That's a weird way to test but ok. You need to make that exception pop up somehow... Have you followed the instructions on SETUP.md?

ZacSharp:
Jup, the ones for command line.

mvrozanti:
I'm going to assume "Jup" means yes. 

But you also need to follow the [More Info](https://github.com/cabaletta/baritone/blob/master/SETUP.md#more-info) section.

ZacSharp:
Your assumption is correct.
That's a bit hard to do at a restart frequency of one per 10 to 20 seconds.

mvrozanti:
>  That's a bit hard to do at a restart frequency of one per 10 to 20 seconds.

What is?

ZacSharp:
My current workflow: runClient, wait, join server, #tomap, alt + F4, alt + tab, little code change, alt + tab, repeat
And copying over files every time is really annoying

mvrozanti:
**Follow the More Info section** [here](https://github.com/cabaletta/baritone/blob/master/SETUP.md#more-info).

ZacSharp:
That's the instructions for users using impact. I'm a dev without impact, so I look at build it yourself.

mvrozanti:
Ok that makes sense then. I'm not sure why your client isn't reflecting the code changes though. I think you need to work on that first.

ZacSharp:
It is. e.g. when I put in output before the double for loop that gets sent

mvrozanti:
What are you having trouble with then?

ZacSharp:
I was just telling you that the code seems to not be run, maybe you can check if you get output from there.

mvrozanti:
Can you elaborate on that?

ZacSharp:
some example values (with applied offsets)
minX = 0
maX = 511
minZ = 2147483647
maxZ 1

mvrozanti:
minX, maxX, minZ, maxZ should result in the (block) coordinates of the northwest and southeast corners of the region, respectively. I'm about to go offline but I'm gonna check this part of the code asap

ZacSharp:
are the coordinates world coordinates or 0-511?
ok, then until tomorrow

mvrozanti:
> are the coordinates world coordinates or 0-511?

Please be more descriptive in your questions.

ZacSharp:
what is the range of mixX/maxX/minZ/maxZ? Are they relative to the world origin or to the region origin?

mvrozanti:
I can only say that `maxX-minX=512` and `maxZ-minZ=512` in **block coordinates**, which are different than chunk coordinates and region coordinates.

ZacSharp:
ok, imma go off for today

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,340 @@
...
+                                }
+                            } catch (Exception e) {
+//                            System.out.println("Couldn't get BlockState for " + x + " " + y + " " + z);
+                            }
+                        }
+                        if (!topView.containsKey(x)) {
+                            if (!topView.get(x).containsKey(z)) {
+                                topView.get(x).put(z, DEFAULT_BLOCK_ID); // this is definitely not accurate
+                            }
+                        }
```
ZacSharp, line 186: Here's the culprit: You are only trying to insert missing z-entries when the whole x-row is missing, which is guaranteed to be false.
```suggestion
                        if (!topView.get(x).containsKey(z)) {
                            topView.get(x).put(z, DEFAULT_BLOCK_ID); // this is definitely not accurate
                        }
```

mvrozanti: Damn. Sorry, I'm not following but I have a feeling you found it problem

ZacSharp: This does not fix the missing block states but it does fix the loop range. Basically whenever we don't get a blockstate that entry is missing in topView.get(x) and the code you had there wasn't fixing this, so when we later use the entries to find the minimum z position there is no key lower that Interger.MAX_VALUE, because there is no key at all.

src/main/java/baritone/command/defaults/ToMapCommand.java
```diff
@@ -0,0 +1,340 @@
...
+                logDirect("Getting top blocks for region " + rX + " " + rZ);
+                for (int x = nw[0]; x < nw[0] + 512; x++) {
+                    if (!topView.containsKey(x)) {
+                        HashMap<Integer, String> zId = new LinkedHashMap<>();
+                        topView.put(x, zId);
+                    }
+                    for (int z = nw[1]; z < nw[1] + 512; z++) {
+                        for (int y = 256; y >= 0; y--) {
+                            try {
+                                BlockState bs = cr.getBlock(x, y, z);
```
ZacSharp, line 154: this uses coordinates relative to the region, as shown in `BlockStateInterface#get0`
Swallowing exceptions seems to be a bad idea, this one causes up to 512*512*256 OutOfBounds exceptions per region and no one noticed it. Now I'm getting actual null blockstates.

mvrozanti: I don't have the time to verify this now. Are you implying that there's a problem in baritone's code itself or is it this problem restricted to ToMapCommand? Also, do you think this is easily fixable?

ZacSharp: Actuall I don't know. This might be a problem of ToMapCommand not accessing the cache correctly or baritone not returning blockstates it is supposed to return. I will try to use the normal ctx.get access method now.

mvrozanti:
@ZacSharp I haven't had the time to look into it but @FriedrichM is also lending us a hand. If can [test it](https://github.com/FriedrichM/baritone/commit/17d8cebd575960c9419e192e8c408188e765ec32)

ZacSharp:
Worked first time. Is not singleplayer safe (easy fix, already did it for testing). Maps don't look really well (as expected). No need for unzipped assets but coloring only looks for the material, resulting in a small colour palette (difficult to fix as the material colour is the only colour directly available from block states I know). Doesn't freeze the game (easy if think).

I managed to get something working yesterday as well, but the heightMap calculation seem to be wrong. If you are interested you can find it in my fork (I just did a quick commit to save it, so it contains all testing changes I had done at that point).

mvrozanti:
I just tried translating it to 1.15.2 and I'm getting blank (transparent) images as well...

mvrozanti:
This is great, thanks for the help @FriedrichM. Turns out it wasn't a bug in baritone then.

ZacSharp:
is it possible to change the branch a pr is merged into? if so please switch this pr to master.

l1ving:
No, it's not.

mvrozanti:
image

Can we get rid of the red lines though? How are they coming into view? I assume those are chunk borders; does IBlockState#getMaterial() change in this case? Why only on water?

l1ving:
Ooh that's cool

FriedrichM:
i am puzzled how these red lines arise. it never happend in my tests, maybe these are old cache files and they changed the format at some point. ill have a look at it and ill also look at how to efficiently stitch those images together

FriedrichM:
While the colormapping is in fact off (as mentioned in my code) i could not recreate the red lines even with old cache files. 

Stitching images together will be hard without using much ram and it will especially be hard without the use of a library. 
I doubt it will be worth adding a library to the jar just for this command so my suggestion is to use imagej [https://imagej.net/Fiji](https://imagej.net/Fiji) with the preinstalled stitching plugin and let the command generate a TileConfiguration.txt file. 
This file can then be used with imagej to stitch the files together.
The downsides are that imagej can be a bit hard to use and imagej doesnt seem to work with too tiled maps, say worldborder to spawn.
I looked into the code (https://github.com/fiji/Stitching) but i couldnt find the reason on first sight, i think its some variable overflowing to the negative tho because I get a negative arrayindex error.

Another suggestion would be to make an option to render all cached regions in a radius around a given coordinate, this would be more flexible than to render all of them.

mvrozanti:
As per the red lines appearing in my example: I have a mild suspicion that might be happening because of another mod of mine that represents new chunks with these red lines. I'm gonna test without it on later today

ZacSharp:
I never had such a problem with [this](https://github.com/ZacSharp/baritone/blob/tomap-command-from-master/src/main/java/baritone/command/defaults/ToMapCommand.java). Would be cool if you actually merged this into your tomap-from-master branch because it was a lot of work and it contains everything @FriedrichM did (merged his branch into mine).

FriedrichM:
add returns to the "Failed!" printouts

ZacSharp:
I thought about it and then left it the way you did it, so the user (and probably maintainers getting it as a bug) gets a proper exception.
It's easier like that but I can change it to `e.printStackTrace(); return;`

ZacSharp:
Is it worth it trying to do something like in [MapWriter](https://github.com/daveyliam/mapwriter/blob/master/src/main/java/mapwriter/BlockColourGen.java) or is that overcomplicated?

mvrozanti:
Seems like we now have 3 different ways of getting color of blocks. 

But we can also choose that later.

The stitching or tiling part now seems to be the most difficult challenge, especially from a memory efficiency aspect. These maps get out of hand really fast.

ZacSharp:
I think it is possible to do it with bitmaps, if you really want to.
The downsides are:
1. Really slow because we reload every region from disk 512 times and write to disk a lot.
2. Takes more space on disk because the image is uncompressed.
I don't know if this would be possible with compressed formats but it will always stay slow.

ZacSharp:
I've got a stitched image here, but apparently it is either to big or the wrong file type for github to render it.
image
How many regions do you actually intend to stitch? Because you can keep up to 1271 (and a half) region images in 1GB ram.

mvrozanti:
Maybe Github doesn't like .ppms :smile: 

> How many regions do you actually intend to stitch? Because you can keep up to 1271 (and a half) region images in 1GB ram.

I'd say as many as possible.

ZacSharp:
So that approach keeps a maximum of `512 pixels * 3 byte per pixel = 1537` bytes in ram, for any output size. I guess that means as long as your hard drive can handle 10,8PB (1,08×10¹⁶ byte) you can stitch the whole world using this.

l1ving:
> I've got a stitched image here, but apparently it is either to big or the wrong file type for github to render it.

Yea, github refuses to render it if it's too big. I learned that the hard way with gifs

PR #2292, opened 2020-12-24
Smooth aim

AppleCidar:
Added a setting called "smoothAim" that will determine if it should slowly move to the coords of the block.


l1ving:
You made the pull into bot-system, which is probably incorrect

PR #2308, opened 2020-12-28
allow creation of more baritone instances

unrootexe:
allows multiple baritone instances to be created through custom PlayerContext implementations

### DIFF
--- a/scripts/proguard.pro
+++ b/scripts/proguard.pro
@@ -17,6 +17,8 @@
 
 -keep class baritone.api.** { *; } # this is the keep api
 
+-keep class baritone.Baritone
+
 # service provider needs these class names
 -keep class baritone.BaritoneProvider
 -keep class baritone.api.IBaritoneProvider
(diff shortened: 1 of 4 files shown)


scripts/proguard.pro
```diff
@@ -17,6 +17,8 @@
 
 -keep class baritone.api.** { *; } # this is the keep api
 
+-keep class baritone.Baritone
+
```
l1ving, line 21: Why?

unrootexe: so i can do
BaritoneAPI.getProvider().getAllBaritones().add(new Baritone(my_custom_player_context))
in my custom utility mod and not have the Baritone class get renamed by proguard

l1ving: Just curious, what is the use case for a different player context this way?

unrootexe: allows you to control multiple players from one minecraft instance since it removes the limit of only having a primary player context

l1ving (approved):
lgtm

unrootexe:
going to need to make changes to this which probably aren't main-branch friendly, closing pr

PR #2413, opened 2021-02-01
compatibility for SodiumChunkManager

wagyourtail:
This addition adds compatibility for https://github.com/jellysquid3/sodium-fabric to baritone. It is done in this way with `@Pseudo` in order for the non fabric builds to not have issues since they obviously wouldn't be able to reference a fabric mod. this is a follow up to #2086

### DIFF
--- a/src/launch/java/baritone/launch/mixins/MixinSodiumChunkProvider.java
+++ b/src/launch/java/baritone/launch/mixins/MixinSodiumChunkProvider.java
@@ -75,18 +75,15 @@ public class MixinSodiumChunkProvider implements IClientChunkProvider {
     
     @Override
     public ISodiumChunkArray extractReferenceArray() {
+        https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.15
         if (chunkArrayField == null) {
-            boolean flag = true;
             for (Field f : this.getClass().getDeclaredFields()) {
                 if (ISodiumChunkArray.class.isAssignableFrom(f.getType())) {
                     chunkArrayField = f;
-                    flag = false;
-                    break;
+                    break https;
                 }
-            } // else
-            if (flag) {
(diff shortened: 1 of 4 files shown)


l1ving (approved):

src/launch/java/baritone/launch/mixins/MixinSodiumChunkProvider.java
```diff
@@ -0,0 +1,97 @@
...
+        } catch (InstantiationException | InvocationTargetException | NoSuchMethodException | IllegalAccessException e) {
+            throw new RuntimeException("Sodium chunk manager initialization for baritone failed", e);
+        } finally {
+            // put this in finally so we can't break anything.
+            this.lock.unlockWrite(stamp);
+        }
+    }
+    
+    @Override
+    public ISodiumChunkArray extractReferenceArray() {
```
leijurv, line 77: hello I would like to suggest a **much** cuter approach to this, booleans and branching are both cringe:

```
    @Override
    public ISodiumChunkArray extractReferenceArray() {
        https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.15
        if (chunkArrayField == null) {
            for (Field f : this.getClass().getDeclaredFields()) {
                if (ISodiumChunkArray.class.isAssignableFrom(f.getType())) {
                    chunkArrayField = f;
                    break https;
                }
            }
            throw new RuntimeException(Arrays.toString(this.getClass().getDeclaredFields()));
        }
        try {
            return (ISodiumChunkArray) chunkArrayField.get(this);
        } catch (IllegalAccessException e) {
            throw new RuntimeException(e);
        }
    }
```

wagyourtail: I wish other languages had python's for-else statement, it would look even cleaner here

PR #2462, opened 2021-02-10
Improve #modified command to include original setting value.

scorbett123:
This will help massively.

Biggest change since the initial commit.


ZacSharp (approved):
I'd use 3 imports rather than a wildcard import but that's not up to me and I'll simply approve.

scorbett123:
That is intellij auto formatting.

PR #2463, opened 2021-02-10
Add the ability to go to items while following

scorbett123:
This is useful for hunting, when you are following an entity, if it is killed you should pick up the items produced, not just leave them on the ground.

When the setting is 0 it is off.

### DIFF
--- a/src/main/java/baritone/process/FollowProcess.java
+++ b/src/main/java/baritone/process/FollowProcess.java
@@ -18,15 +18,14 @@
 package baritone.process;
 
 import baritone.Baritone;
-import baritone.api.pathing.goals.Goal;
-import baritone.api.pathing.goals.GoalComposite;
-import baritone.api.pathing.goals.GoalNear;
-import baritone.api.pathing.goals.GoalXZ;
+import baritone.api.pathing.goals.*;
 import baritone.api.process.IFollowProcess;
 import baritone.api.process.PathingCommand;
 import baritone.api.process.PathingCommandType;
 import baritone.utils.BaritoneProcessHelper;
 import net.minecraft.entity.Entity;
+import net.minecraft.entity.item.EntityItem;
+import net.minecraft.entity.item.EntityXPOrb;
 import net.minecraft.util.math.BlockPos;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


ZacSharp (approved):
(Here I'd add one line rather than removing 3 as well)

scorbett123:
Same here, autoformatting in intellij does this.

l1ving (approved):

ZacSharp:
It should only target items that were dropped nearby. (don't break through ceilings just because a zombie died there 2 minutes ago) 

Also I think this could be more generic by using a higher priority process, but I'm not sure about that

scorbett123:
> It should only target items that were dropped nearby. (don't break through ceilings just because a zombie died there 2 minutes ago)
> 
> Also I think this could be more generic by using a higher priority process, but I'm not sure about that

I agree with you, was doing this for something else, thought that I might as well create the pull. It can be done through process priorities, however I believe would have to be temporary.

src/main/java/baritone/process/FollowProcess.java
```diff
@@ -50,6 +49,13 @@ public FollowProcess(Baritone baritone) {
 
     @Override
     public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
+        for (Entity entity : ctx.world().loadedEntityList) {
+            if (entity instanceof EntityItem | entity instanceof EntityXPOrb) {
```
leijurv, line 53: delicious usage of bitwise or instead of logical or

😋

leijurv:
Wondering why this "overrides" the tracking of the actual entity? It replaces the composite goal with a singluar goal to the item. It also appears to choose the entity arbitrarily - just whichever entity happens to be first in `world.loadedEntityList`?

What do we think about just adding the XP and item entities to the goals in the existing goal composite in this function?

scorbett123:
This was made mainly in mind of hunt modules, however I was going to change it to it's own process, just not had time yet.

src/main/java/baritone/process/FollowProcess.java
scorbett123: Wait, what is the actual difference? Is there an actual performance effect? Does it sometimes give different results?

wagyourtail: yes, bitwise or has to check every part, while logical or will shortcut the first time it hit's a "true"

scorbett123: Thanks

PR #2482, opened 2021-02-17
Clear caches of schematics when moving them

### LINKED ISSUE #1925: buildRepeat with replacing blocks does not update positions of blocks to replace
## Some information
Operating system: Linux,Ubuntu / Linux,Ubuntu
Java version: 1.8 / 1.8
Minecraft version: 1.12.2 / 1.15.2
Baritone version: v1.2.14 / v1.5.2
Forge mods (if used): None / my own, private one to add a branchmineSchematic

## Exception, error or logs
There are no exceptions, errors or any log except for the chat messages:
* "Filling now"
* "Repeating in build vector Vec3i{x=1, y=0, z=0}, new origin is BlockPos{x=\*, y=\*, z=\*}"
* "Done building"

## How to reproduce
1. place a log and a dirt in this pattern, with up being positive x and right being positive z
    log dirt
    dirt log
    log dirt
    dirt log
2. select the dirt and log at the negative x side
3. run `#buildRepeat 1,0,0`
4. get some stone and run `#sel replace dirt stone`
5. see it replace first the dirt and all blocks following in the positive x direction, regardless of their type

## Modified settings
* buildRepeat 1,0,0

## Final checklist
- [x] I have not used any OwO's or UwU's in this issue, except for the two in this line.

ZacSharp:
Some of Baritone's schematics have a cache, that becomes outdated when the schematic is moved by `buildRepeat`. For expamle this caused `#replace` combined with `#buildRepeat` to always replace the same positions relative to the origin of the schematic.
Since this behaviour is actually being used, I added a setting to toggle it and set the default to not change anything.

fixes #1925

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -835,6 +835,13 @@ public final class Settings {
     public final Setting<Integer> buildRepeatCount = new Setting<>(-1);
 
     /**
+     * Don't notify schematics that they are moved.
+     * e.g. replacing will replace the same spots for every repetition
+     * Mainly for backward compatibility.
+     */
+    public final Setting<Boolean> buildRepeatSneaky = new Setting<>(true);
+
+    /**
      * Allow standing above a block while mining it, in BuilderProcess
      * <p>
      * Experimental
--- a/src/api/java/baritone/api/schematic/CompositeSchematic.java
+++ b/src/api/java/baritone/api/schematic/CompositeSchematic.java
@@ -75,9 +75,7 @@ public class CompositeSchematic extends AbstractSchematic {
     @Override
     public void reset() {
         for (CompositeSchematicEntry entry : schematicArr) {
-            if (!(entry.schematic instanceof IStaticSchematic)) {
-                entry.schematic.reset();
-            }
+            entry.schematic.reset();
         }
     }
 }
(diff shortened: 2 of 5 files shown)


leijurv:
oh, this is based

leijurv:
first, could i trouble you to please run the intellij autoformatter? `){` should be `) {`. 

Also, why does it need to do `!(schematic instanceof IStaticSchematic)`? I'd prefer if we called `.reset()` unconditionally, and if it's a static schematic it's just a noop. So `reset` is overridden only when it needs to do anything

ZacSharp:
"trouble" is the right word, that thing took me half an hour to get it working (should be easier next time though, now that I have the command)

When I wrote this a few months ago I wanted to enforce the staticness of `IStaticSchematic`s a little and when I added the setting and created the pr I didn't bother to remove it.

leijurv:
Hm, maybe good to get a third opinion: should this default to true or false? I'm not really sure. What sort of backward compatibility are we talking here? If I understand right, the current behavior is buggy when combining buildrepeat with a non-static schematic, and this fixes it? Or is there any upside to the old (current) way?

ZacSharp:
This pr is intentionally a noop unless `buildRepeatSneaky` is set to false (default is true), because I saw the current behaviour being used for simple highway/road builing by building a template and using `buildRepeat` along with `#sel replace <template material> <intended material>`

leijurv:
Crazy lmao. Alright.

PR #2484, opened 2021-02-17
Add buildAvoidBreakCheck setting

loonybtard:
Allow to deny builder break unsafe blocks (like blocks that are adjacent to liquids).

| `#set buildAvoidBreakCheck true` | `#set buildAvoidBreakCheck false` |
|----------------------------------------------|------------------------------------------------|
| example1.1 | example1.2 |
| example2.1  | example2.2 |

### todo
English is not my native language so javadoc description may be better. "Allow edits by maintainers" checked.

### Final checklist
- [ ] I have not used any OwO's or UwU's in this pull request.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -207,6 +207,11 @@ public final class Settings {
     )));
 
     /**
+     * If this is true, the builder will use avoid checks from mining(for example: does will not break block near liquid).
+     */
+    public final Setting<Boolean> buildAvoidBreakCheck = new Setting<>(false);
+
+    /**
      * A list of blocks to become air
      * <p>
      * If a schematic asks for a block on this list, only air will be accepted at that location (and nothing on buildIgnoreBlocks)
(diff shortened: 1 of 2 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -206,6 +206,11 @@
 
     )));
 
+    /**
+     * If this is true, the builder will use avoid checks from mining(for example: does will not break block near liquid).
```
ZacSharp, line 210:
```suggestion
     * If this is true, the builder won't break certain blocks (e.g. blocks near liquids).
```

ZacSharp (changes requested):
1. This only prevents breaking blocks near liquids when they should be air, but the builder will break any wrong block and because it prefers breaking over placing the liquid likely has plenty of time to flow out.
2. This does not avoid updating falling blocks next to liquids (pathing does that)
3. Give it some lava to clear out but no throwaways and see it pathing eternally<br> I think this one might be older, but without this pr it gets stuck in the lava before this can happen

ZacSharp:
@loonybtard will you work on getting this ready to be merged?
If not just tell me I and will open a pr with this feature myself.

Issue #2344, opened 2021-01-10
[1.16.4] Endless pathing


juleklO:
_OS: WinPro10 64bit
Mod version: 1.6.2
Settings used: buildInLayers true, layerorder true, buildinlayers true;_

When I select a large area to mine out the bot mines for a while and than stops in place and tried to find path. Even when I try moving the bot that has been standing in place for a while to a different spot it takes over and tries to return to the block where it was standing and trying to path. 

Additionally when a selection is made where "layer 0" is comprised exclusively of air the bot doesn't select the layer as "complete".

I will try to answer any and all questions below.

juleklO:
https://youtu.be/GRW-gAq3rIc <- this in video form with some changed settings, all in desc.

scorbett123:
I know that this is an issue (I have seen it before), however can you try without any settings changed and check for any difference in behaviour. This will just let us know whether a single setting is messing this us or if everything is broken. Do you happen to have sodium installed as well?

juleklO:
> I know that this is an issue (I have seen it before), however can you try without any settings changed and check for any difference in behaviour. This will just let us know whether a single setting is messing this us or if everything is broken. Do you happen to have sodium installed as well?

If I don't know what _sodium_ is than most likely I do not have it installed, but what can change a thing or two is that I'm using the barritone as an integration from impact client. I will test with unchanged settings and update you on the situation.

juleklO:
[](https://youtu.be/UKCdEm3GdCA) TL:DR changing settings doesn't change the behaviour

l1ving:
It does, you might need to restart pathing though

juleklO:
> It does, you might need to restart pathing though

What do you mean by _restart pathing_?

l1ving:
stop and start it 

after testing with allowbreak and the such, baritone changes it's path when settings are changed

juleklO:
> stop and start it
> 
> after testing with allowbreak and the such, baritone changes it's path when settings are changed

I did as show in the video, I changed the settings, did #cancel and than (visible in the video) #sel cleararea with the changed settings applied

scorbett123:
can you run #modified, in the video it seemed that you were thinking those were the changed settings, however this will tell you for certain all of the settings that you have changed.

juleklO:
[https://youtu.be/e0gqi1E0Eh8](url) give it a sec to upload, but still no changes

scorbett123:
[...]

PR #2504, opened 2021-02-22
Update FillSchematic.java

Altonos:
Fix by @scorbett123

Fix for:
https://github.com/cabaletta/baritone/issues/2344


scorbett123:
Now, let's see if it actually builds!!!!

PR #2526, opened 2021-03-01
Settings to alter block treatment while builing

### LINKED ISSUE #1523: Replace with Air is not working correctly
## Some information
Operating system: Win 10
Java version: 1.8.0_241
Minecraft version: 1.15.2
Baritone version: 1.5.1
Forge mods (if used): No

## Exception, error or logs
[latest - Kopie.txt](https://github.com/cabaletta/baritone/files/4446605/latest.-.Kopie.txt)
Windows: `%appdata%/.minecraft/logs/`

## How to reproduce
`#sel replace air cobblestone`
I selected an Area 1x1x1000 and tried to replace all empty blocks with cobblestone

Two unwanted behaviors I found. 

- Removing filled blocks and replace with cobblestone
- Not filling Air blocks.
- There is no error Log in the ChatDebug. Completed Successfully.
image

[Video](https://transfer.sh/eWQ7S/gP14tMCT7j.mp4)

## Modified settings

- allow DiagonalAscend
- allowDiagonalDescend
- allowInventory
- allowParkour
- allowParkourPlace
- chatControl
- chatDebug
- rendererCachedChunks

also tried without:

- avoidance
- layerOrder
- assumeWalkonWater

## Final checklist

### LINKED ISSUE #2225: Treat grass_block and others as dirt/soil
## Describe your suggestion
There should be a setting to treat grass_blocks and other soil blocks as dirt. it would be similar to buildIgnoreBlocks when baritone treats blocks as air.

## Settings
.b treatAsSoil
toggle on off
which blocks to treat as dirt

## Context
Many schematics include a deep dirt block below them to start. You don't always want to ignore all the dirt and treat it as air. Many times you want to include all that solid schematic, but it doesn't matter if its dirt, stone, iron_blocks, grass_blocks, etc.
This is especially true for surface level stuff. As it stands now, if you have a schematic with all the surface as dirt, you try to build it in a world where the surface is grass, Baritone will take its time to break each grass_block just to make a a dirt. There should be a way to tell baritone that the two are one in the same.

## Final checklist

### LINKED ISSUE #2227: Baritone stands still and does nothing when using #mine command
## Some information
Operating system: Windows 10
Java version: 1.8.0
Minecraft version: 1.16.3
Baritone version: 1.6.1
Forge mods (if used): https://www.modpackindex.com/modpack/21762/technical-electrical-vi

## Exception, error or logs
[latest.log](https://github.com/cabaletta/baritone/files/5624891/latest.log)
[debug.log](https://github.com/cabaletta/baritone/files/5624893/debug.log)

## How to reproduce
Type in #mine <any_ore> and the bot finds ore and a path to them but just stands there doing nothing

## Modified settings
None

## Final checklist

### LINKED ISSUE #2318: baritone stops tunneling if it encounter holes/ravines
## Some information
Operating system: Windows 10
Java version: 1.8.0_271
Minecraft version: 1.16.4
Baritone version: 1.6.2
Forge mods (if used): 

## Exception, error or logs

While tunneling baritone will freeze in place if it encounters a hole in the tunnel direction
How do i tell it to make a bridge under him in order to continue tunneling

[2020-12-31-2.log.gz](https://github.com/cabaletta/baritone/files/5759404/2020-12-31-2.log.gz)

## How to reproduce

#tunnel 3 1 600

## Modified settings

nothing all default

## Final checklist

### LINKED ISSUE #2323: buildSkipBlock - a Schematic block that says, I do not care what is in the world -- Ullage for Schematics
## Describe your suggestion
buildIgnoreBlocks   and   buildIgnoreExisting  do the opposite of what I am suggesting.  They ignore blocks in the world that do not match what is in the schematic.  

I want the opposite, I want to pre-process the blocks in the schematic, interpret the blocks in the schematic.  I want to add a new concept, that a particular block in the schematic tells baritone not to change anything in the world.  I do not want that to be a fixed block like Air.   I want it to be a settable block.  When this block is found in the schematic, baritone should take no action, baritone should say, I do not care what is in the world, it is OK as it is, do not try to change the world at that point.    I call this the I-do-not-care variable.   Usage would be:

#I-do-not-care  [{block_ID}/clear]

Usage to see what the I-do-not-care block is set to.
#I-do-not-care                                  

Usage to set the I-do-not-care block to be sand
#I-do-not-care sand              

Usage to clear the previous definition of the I-do-not-care block so Baritone returns to interpreting schematics literally.
#I-do-not-care clear                     

When a designer intends a schematic to use an I-do-not-care block they must choose the block at design time, and communicate this choice to users of the schematic.   The author of the schematic might save the name of the schematic to signal to the user what to set the I-do-not-care block should be:
     
My_favorite_schematic_IDC_Sand.schematic

There are so many blocks available in Minecraft that it is possible to choose any buildable block that does not appear in the intended build of the schematic.      So say I want sand to be my I-do-not-care block.   I can save the schematic, and use the schematic without the I-do-not-care-block set, and I see the oak planks being built -- this allows me to see what is in schematic.  Then I set 

#I-do-not-care sand 

and I build the schematic again -- this time not only does it not build the sand, it does not do anything at the locations that there were sand in the schematic.    

At run time, the user would know they are using a schematic with an IDC block, and they would have to set the I-do-not-care block variable appropriately before running the schematic.

The use I would put IDC blocks to is, when I want to build a schematic box to enclose a large structure, where part of the structure contains Air, I can fill the portion of the box that I do not really care about with IDC blocks.   It means I can create schematics that are effectively not right-prisms.   It means I can build triangles or pyramid-shaped schematics, and that part of the shape that is not used in the bounding-prism box is just filled with IDC blocks.   This would allow me to make effectively non-prismatic schematics with any current design tool for creating schematics. 

For instance, but certainly not limited to this use, when I am building highways in 2b2t, on axis-aligned highways, I make my schematics 6 blocks long in the highway direction.  An axis-aligned highway is a right-cylinder.   I only need to define its shape in a right-cross-section, which for minecraft, is a one-block-wide slice across the highway.   Unfortunately, one experiences glitches where blocks are laid and then disappear.  Baritone's behavior is to re-place blocks in the current schematic image that are no longer present, however, if one uses #buildrepeat, when the Baritone advances by the increment, when blocks from the last lay of the schematic, that are not in this lay of the schematic disappear, they stay as a defect after the passage of Baritone.   HighwayTools (the KAMI/Blue module) uses 2-block wide schematics, and under some conditions will leave glitches behind.   With a 6-block long schematic, every block is in 5 consecutive advances of the schematic...  In practice, only the blocks on the leading edge of the schematic are laid in a given cycle.  If there are no glitches, there is no need to re-place any blocks.  As glitches appear later and later, they may be replaced in one of the 5 previous lays.   This is a solved problem -- glitch-free axis-aligned highways -- just make the schematics longer than they have to be.   

This does not solve the problem for any highway at any angle where one does not want to determine what appears in certain blocks in the schematic -- if one does not have the use of IDC blocks.   Without IDC blocks, there is no way to have an entirely overlapping schematic that advances with buildrepeat 1,0,1 for instance without including a lot of other blocks that either Baritone must place sometimes or must remove sometimes.   When the schematic is limited only to blocks that must be determined, the blocks in the wing columns of the schematic only appear in one lay of the schematic as it advances along the other diagonal.   With the presence of glitches, there are a lot of blocks not proper in these wing columns after the pass of baritone.   The problem can be solved by enlarging the bounding prism to contain multiple appearances of all blocks in every column but then the wings of the prism extend into spaces we really do not care what is there.   No strict interpretation of the schematic will solve this problem.   

I could set a schematic for a large underground installation, and set only wall skins blocks and the contained Air blocks to be the desired block in the schematic... all the space between rooms and between tiers could be made IDC blocks, and then, if I base my schematic on a save from a world, and try to build it again at a different part of the world, Baritone is not trying to rebuild the random mineral deposits that might have appeared between rooms or between tiers in the original schematic.   Anyone who has tried to make a schematic from an existing underground build has encountered this problem.   One would be able to make standardized secret entrances (useful in Vanilla minecraft for sure) where all elements needed in some odd geometry are specified, but all unimportant parts of the bounding box are IDC blocks.   

Freeing Schematics from the Strictly-Duplicated Right-Prism Prison is what IDC blocks are about.   
      
As a further extension, I propose a second variable, As-long-as-it-is -- the full sentence would be I-do-not-care As-long-as-it-is (one of a list):

Usage:
#As-long-as-it-is [{list of block_IDs}/clear]

Usage to list the list of blocks that Baritone will apply IDC filtering to:
#As-long-as-it-is                   

Usage to set the list of blocks that Baritone will apply IDC filtering to:
#As-long-as-it-is Air Obsidian   
If the IDC block maps to a place in the world that is Air or Obsidian, do nothing.

Usage to clear the list of blocks that Baritone will apply IDC filtering to:
#As-long-as-it-is clear              

The behaviour would be, if 

#I-do-not-care sand
#As-long-as-it-is Air Obsidian

If the place in the world that sand maps to is Air or Obsidian do nothing.
In the case that the location in the world is neither Air nor Obsidian  the IDC algorithm *does* care and makes the block the first element of the #As-long-as-it-is list, in this case Air.   

I would use this particular feature for, but not limited to, when I want to dig an area in preparation to laying Obsidian.   If there is Obsidian in the schematic, and I do not have any, Baritone will fail.   If there is no Obsidian in the schematic and Obsidian is encountered in the world, Baritone will spend a lot of time mining it.   

With IDC variable, set to sand, I can just put sand in the schematic where I might eventually want to put Obsidian, and then say #As-long-as-it-is Air Obsidian -- Baritone would clear away any blocks which are not Obsidian.  I could use the build schematic, but set the IDC variable to Obsidian and the #As-long-as-it-is Air Obsidian, and get the same effect.   

Therefore:

#As-long-as-it-is Air Obsidian

is different from:

#As-long-as-it-is Obsidian Air

In the first case, the world gets more air for non-matching blocks, in the second more obsidian.  

This cannot be done by buildIgnoreBlocks   and   buildIgnoreExisting because those interpret the schematic literally and then perform judgements on the world, and only with Air either in the schematic or in the world: 

>     buildIgnoreBlocks
>     public final Settings.Setting<java.util.List<net.minecraft.block.Block>> buildIgnoreBlocks
>     A list of blocks to be treated as if they're air.
>     If a schematic asks for air at a certain position, and that position currently contains a block on this list, it will be treated as correct.
> 

>     buildIgnoreExisting
>     public final Settings.Setting<java.lang.Boolean> buildIgnoreExisting
>     If this is true, the builder will treat all non-air blocks as correct. It will only place new blocks.
> 

To do what I want to do, one has to interpret the schematic with a filter, and apply it to the world, regardless of what is in the world at that location, or only if the world does not contain certain blocks.

### LINKED ISSUE #2344: [1.16.4] Endless pathing

ZacSharp:
This adds the following new settings
* `#buildSkipBlocks` a list of blocks that are treated as if they were outside the schematic
* `#buildValidSubstitutes` a mapping of blocks to blocks that are treated as correct in their position<br>the same as what `#buildIgnoreExisting` does, but more general
* `#buildSubstitutes` a mapping of blocks to blocks that will be placed instead, preferring blocks in the order they are given in

While writing this I noticed that this additionally "breaks" buildIgnoreBlocks so that the behaviour is now the same for filling and building schematics from disk and fits the behaviour the user would expect after reading the documentation, which wasn't the case for filling.

closes #2323
closes #2225
fixes #2344 
fixes #2318
and from what @scorbett123 said I conclude this should fix #2227 as well
maybe fixes #1523
that's more closed issues for a side effect of this pr (4 duplicates/symptoms of 1 bug) than for what it was made for (2 enhancements), lol

### DIFF
--- a/src/api/java/baritone/api/schematic/FillSchematic.java
+++ b/src/api/java/baritone/api/schematic/FillSchematic.java
@@ -19,7 +19,6 @@ package baritone.api.schematic;
 
 import baritone.api.utils.BlockOptionalMeta;
 import net.minecraft.block.state.IBlockState;
-import net.minecraft.init.Blocks;
 
 import java.util.List;
 
--- a/src/api/java/baritone/api/schematic/FillSchematic.java
+++ b/src/api/java/baritone/api/schematic/FillSchematic.java
@@ -44,8 +44,6 @@ public class FillSchematic extends AbstractSchematic {
     public IBlockState desiredState(int x, int y, int z, IBlockState current, List<IBlockState> approxPlaceable) {
         if (bom.matches(current)) {
             return current;
-        } else if (current.getBlock() != Blocks.AIR) {
-            return Blocks.AIR.getDefaultState();
         }
         for (IBlockState placeable : approxPlaceable) {
             if (bom.matches(placeable)) {
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -17,7 +17,6 @@
 
 package baritone.api.schematic;
 
-import baritone.api.utils.BlockOptionalMetaLookup;
 import net.minecraft.block.Block;
 import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.IBlockState;
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -19,8 +19,10 @@ package baritone.api.schematic;
 
 import baritone.api.utils.BlockOptionalMetaLookup;
 import net.minecraft.block.Block;
+import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.init.Blocks;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
@@ -49,7 +51,7 @@ public class SubstituteSchematic extends AbstractSchematic {
         }
         List<Block> substitutes = substitutions.get(desiredBlock);
         if (substitutes.contains(current.getBlock()) && !current.getBlock().equals(Blocks.AIR)) {// don't preserve air, it's almost always there and almost never wanted
-            return current;
+            return withBlock(desired, current.getBlock());
         }
         for (Block substitute : substitutes) {
             if (substitute.equals(Blocks.AIR)) {
@@ -57,10 +59,25 @@ public class SubstituteSchematic extends AbstractSchematic {
             }
             for (IBlockState placeable : approxPlaceable) {
                 if (substitute.equals(placeable.getBlock())) {
-                    return placeable;
+                    return withBlock(desired, placeable.getBlock());
                 }
             }
         }
         return substitutes.get(0).getDefaultState();
+    }
+
+    private IBlockState withBlock(IBlockState state, Block block) {
+        Collection<IProperty<?>> properties = state.getPropertyKeys();
+        IBlockState newState = block.getDefaultState();
+        for (IProperty<?> property : properties) {
+            try {
+                newState = copySingleProp(state, newState, property);
+            } catch (IllegalArgumentException e) { //property does not exist for target block
+            }
+        }
+        return newState;
+    }
+    private <T extends Comparable<T>> IBlockState copySingleProp(IBlockState fromState, IBlockState toState, IProperty<T> prop) {
+        return toState.withProperty(prop, fromState.getValue(prop));
     }
 }
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -23,6 +23,7 @@ import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.init.Blocks;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -30,6 +31,7 @@ public class SubstituteSchematic extends AbstractSchematic {
 
     private final ISchematic schematic;
     private final Map<Block, List<Block>> substitutions;
+    private final Map<IBlockState, Map<Block, IBlockState>> blockStateCache = new HashMap<>();
 
     public SubstituteSchematic(ISchematic schematic, Map<Block,List<Block>> substitutions) {
         super(schematic.widthX(), schematic.heightY(), schematic.lengthZ());
@@ -67,6 +69,9 @@ public class SubstituteSchematic extends AbstractSchematic {
     }
 
     private IBlockState withBlock(IBlockState state, Block block) {
+        if (blockStateCache.containsKey(state) && blockStateCache.get(state).containsKey(block)) {
+            return blockStateCache.get(state).get(block);
+        }
         Collection<IProperty<?>> properties = state.getPropertyKeys();
         IBlockState newState = block.getDefaultState();
         for (IProperty<?> property : properties) {
@@ -75,6 +80,7 @@ public class SubstituteSchematic extends AbstractSchematic {
             } catch (IllegalArgumentException e) { //property does not exist for target block
             }
         }
+        blockStateCache.computeIfAbsent(state, s -> new HashMap<Block,IBlockState>()).put(block, newState);
         return newState;
     }
     private <T extends Comparable<T>> IBlockState copySingleProp(IBlockState fromState, IBlockState toState, IProperty<T> prop) {
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -18,6 +18,7 @@
 package baritone.api.schematic;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockAir;
 import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.init.Blocks;
@@ -49,12 +50,12 @@ public class SubstituteSchematic extends AbstractSchematic {
             return desired;
         }
         List<Block> substitutes = substitutions.get(desiredBlock);
-        if (substitutes.contains(current.getBlock()) && !current.getBlock().equals(Blocks.AIR)) {// don't preserve air, it's almost always there and almost never wanted
+        if (substitutes.contains(current.getBlock()) && !(current.getBlock() instanceof BlockAir)) {// don't preserve air, it's almost always there and almost never wanted
             return withBlock(desired, current.getBlock());
         }
         for (Block substitute : substitutes) {
-            if (substitute.equals(Blocks.AIR)) {
-                return Blocks.AIR.getDefaultState(); // can always "place" air
+            if (substitute instanceof BlockAir) {
+                return current.getBlock() instanceof BlockAir ? current : Blocks.AIR.getDefaultState(); // can always "place" air
             }
             for (IBlockState placeable : approxPlaceable) {
                 if (substitute.equals(placeable.getBlock())) {
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -49,23 +49,18 @@ public class SubstituteSchematic extends AbstractSchematic {
         }
         List<Block> substitutes = substitutions.get(desiredBlock);
         if (substitutes.contains(current.getBlock()) && !current.getBlock().equals(Blocks.AIR)) {// don't preserve air, it's almost always there and almost never wanted
-            System.out.println(String.format("%s is already placed", current));
             return current;
         }
         for (Block substitute : substitutes) {
             if (substitute.equals(Blocks.AIR)) {
-                System.out.println("air, lol");
                 return Blocks.AIR.getDefaultState(); // can always "place" air
             }
             for (IBlockState placeable : approxPlaceable) {
                 if (substitute.equals(placeable.getBlock())) {
-                    System.out.println(String.format("%s can be placed", placeable));
                     return placeable;
                 }
             }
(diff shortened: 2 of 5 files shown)


src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -794,6 +799,12 @@ private boolean valid(IBlockState current, IBlockState desired, boolean itemVeri
         if (!(current.getBlock() instanceof BlockAir) && Baritone.settings().buildIgnoreExisting.value && !itemVerify) {
             return true;
         }
+        if (Baritone.settings().buildSkipBlocks.value.contains(desired.getBlock()) && !itemVerify) {
+            return true;
+        }
+        if (Baritone.settings().buildValidSubstitutes.value.getOrDefault(desired.getBlock(), Arrays.asList()).contains(current.getBlock()) && !itemVerify) {
```
leijurv, line 805: `Collections.emptyList()` instead of `Arrays.asList()` to avoid alloc

src/api/java/baritone/api/utils/SettingsUtil.java
```diff
@@ -261,6 +262,36 @@ public String toString(ParserContext context, Object value) {
             public boolean accepts(Type type) {
                 return List.class.isAssignableFrom(TypeUtils.resolveBaseClass(type));
             }
+        },
+        MAPPING() {
+            @Override
+            public Object parse(ParserContext context, String raw) {
+                Type keyType = ((ParameterizedType) context.getSetting().getType()).getActualTypeArguments()[0];
```
leijurv, line 269: Is this confirmed to work in `standalone` where this is run through proguard? i.e. is this type information confirmed to be preserved?

src/api/java/baritone/api/schematic/FillSchematic.java
```diff
@@ -44,8 +43,6 @@ public BlockOptionalMeta getBom() {
     public IBlockState desiredState(int x, int y, int z, IBlockState current, List<IBlockState> approxPlaceable) {
         if (bom.matches(current)) {
             return current;
-        } else if (current.getBlock() != Blocks.AIR) {
-            return Blocks.AIR.getDefaultState();
```
leijurv, line 48: Do we know why this was originally here? IIRC this is what causes issues in higher versions of MC what with cave air and all that. Totally fine to remove if it fixes the related bug.

leijurv: ah, it's logandark code https://github.com/cabaletta/baritone/commit/991283182cde4b61bdef3743a1e2d53fc01c2bdb#diff-13b77f673395d72c1cda8da1c46c2613bc827c422cd6d0a7290b1816a31b8ca0R24

lol

src/api/java/baritone/api/schematic/SubstituteSchematic.java
```diff
@@ -0,0 +1,82 @@
...
+        Block desiredBlock = desired.getBlock();
+        if (!substitutions.containsKey(desiredBlock)) {
+            return desired;
+        }
+        List<Block> substitutes = substitutions.get(desiredBlock);
+        if (substitutes.contains(current.getBlock()) && !current.getBlock().equals(Blocks.AIR)) {// don't preserve air, it's almost always there and almost never wanted
+            return withBlock(desired, current.getBlock());
+        }
+        for (Block substitute : substitutes) {
+            if (substitute.equals(Blocks.AIR)) {
```
leijurv, line 56: Just to avoid a headache while porting up, perhaps these should both be `instanceof BlockAir`?

src/api/java/baritone/api/schematic/SubstituteSchematic.java
```diff
@@ -0,0 +1,82 @@
...
+                return Blocks.AIR.getDefaultState(); // can always "place" air
```
leijurv, line 57: I don't know for sure, but perhaps it would fix things if this were instead `return current.getBlock() instanceof BlockAir ? current : Blocks.AIR.getDefaultState();`? Very unsure of this though, feel free to ignore if I'm wrong

src/api/java/baritone/api/schematic/SubstituteSchematic.java
```diff
@@ -0,0 +1,82 @@
...
+            for (IBlockState placeable : approxPlaceable) {
+                if (substitute.equals(placeable.getBlock())) {
+                    return withBlock(desired, placeable.getBlock());
+                }
+            }
+        }
+        return substitutes.get(0).getDefaultState();
+    }
+
+    private IBlockState withBlock(IBlockState state, Block block) {
```
leijurv, line 68: Perhaps this should be cached? Worried about all these allocations within `desiredState`, which is called a LOT.

src/api/java/baritone/api/utils/SettingsUtil.java
ZacSharp: It does work, but the tab completion seems to break in `standalone` (on master as well), is that a known thing?

src/api/java/baritone/api/schematic/SubstituteSchematic.java
ZacSharp: I'm not sure about this because I
1. Don't want to special-case too many blocks (i.e. when they say "replace it with cave_air" Baritone should try doing so)
2. When used as a "build as long as you can and proceed anyway if you can't" this will likely fail in caves, because people likely didn't put `cave_air` as a substitute.

But since I consider the 2. point to be more important, I'll use what you suggested rather than the alternative for 1. (`return current.getBlock() instanceof BlockAir ? current : substitute;`)

src/api/java/baritone/api/schematic/FillSchematic.java
ZacSharp: Well, I was hoping you knew why it was here.

TrueHero1:
Hey I'm trying to build a schematic with dirt blocks but every time they turn into grass baritone goes and breaks them and replaces them could you tell me what exactly I have to type to fix this

PR #2338, opened 2021-01-07
1.16.4 - Full Litematica Building Support

EmersonDove:
Added full Litematica support. Usage is ```#build <filepath>```.

PR #2544, opened 2021-03-07
Added Litematica support

EmersonDove:
Added litematica support to 1.12.2 master.

Issues with this version:
- Some states (such as iron bars) are unrecognized
- In 1.12, the long array (`blockstates`) isn't exposed by the libraries so parsing has to be done manually
- Only reads in first region of the .litematic

#2338 Is a fully working implementation for modern versions of MC

If anyone has knowledge or cleaner solutions for any of these leave a comment so I can update it.

### DIFF
--- a/src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
+++ b/src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
@@ -1,0 +1,217 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.utils.schematic.format.defaults;
+
+import baritone.utils.schematic.StaticSchematic;
+import net.minecraft.block.*;
+import net.minecraft.block.properties.IProperty;
+import net.minecraft.nbt.*;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.block.state.IBlockState;
+
+import org.apache.commons.lang3.Validate;
+import javax.annotation.Nullable;
+import java.util.*;
+
+/**
+ * @author Emerson
+ * @since 12/27/2020
+ */
+public final class LitematicaSchematic extends StaticSchematic {
+
+    public LitematicaSchematic(NBTTagCompound nbt) {
+        String regionName = (String) nbt.getCompoundTag("Regions").getKeySet().toArray()[0];
+        this.x = Math.abs(nbt.getCompoundTag("Regions").getCompoundTag(regionName).getCompoundTag("Size").getInteger("x"));
+        this.y = Math.abs(nbt.getCompoundTag("Regions").getCompoundTag(regionName).getCompoundTag("Size").getInteger("y"));
+        this.z = Math.abs(nbt.getCompoundTag("Regions").getCompoundTag(regionName).getCompoundTag("Size").getInteger("z"));
+        this.states = new IBlockState[this.x][this.z][this.y];
+
+
+        NBTTagList paletteTag = nbt.getCompoundTag("Regions").getCompoundTag(regionName).getTagList("BlockStatePalette",10);
+        // ListNBT paletteTag = nbt.getCompound("Regions").getCompound(regionName).getList("BlockStatePalette",10);
+
+        // Create the block states array
+        IBlockState[] paletteBlockStates = new IBlockState[paletteTag.tagCount()];
+        // For every part of the array
+        for (int i = 0; i<paletteTag.tagCount(); i++) {
+            // Set the default state by getting block name
+            Block block = Block.REGISTRY.getObject(new ResourceLocation((((NBTTagCompound) paletteTag.get(i)).getString("Name"))));
+            IBlockState blockState = block.getDefaultState();
+            NBTTagCompound properties = ((NBTTagCompound) paletteTag.get(i)).getCompoundTag("Properties");
+            Object[] keys = properties.getKeySet().toArray();
+            Map<String, String> propertiesMap = new HashMap<>();
+            // Create a map for each state
+            for (int j = 0; j<keys.length; j++) {
+                propertiesMap.put((String) keys[j], (properties.getString((String) keys[j])));
+            }
+            for (int j = 0; j<keys.length; j++) {
+                IProperty<?> property = block.getBlockState().getProperty(keys[j].toString());
+                if (property != null) {
+                    blockState = setPropertyValue(blockState, property, propertiesMap.get(keys[j]));
+                }
+            }
+            paletteBlockStates[i] = blockState;
+        }
+
+
+        // BlockData is stored as an NBT long[]
+        int paletteSize = (int) Math.floor(log2(paletteTag.tagCount()))+1;
+        long litematicSize = (long) this.x*this.y*this.z;
+
+        // In 1.12, the long array isn't exposed by the libraries so parsing has to be done manually
+        String rawBlockString = (nbt.getCompoundTag("Regions").getCompoundTag(regionName)).getTag("BlockStates").toString();
+        rawBlockString = rawBlockString.substring(3,rawBlockString.length()-1);
+        String[] rawBlockArrayString = rawBlockString.split(",");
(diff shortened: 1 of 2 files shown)


src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,217 @@
...
+            {
+                if (number < 0)
+                {
+                    interval *= -1;
+                }
+
+                long i = number % interval;
+                return i == 0 ? number : number + interval - i;
+            }
+        }
```
scorbett123, line 215:
```suggestion
        public static long roundUp(double number, long interval)
        {
            if (number < 0){
                interval *= -1;
            }
            return interval * (long) Math.ceil(number/interval);
        }
```

I believe that this has exactly the same functionality, if you have any reason for wanting to use this instead please do, however I think this is nicer,

ZacSharp: No, it will throw an exception instead of returning 0 if interval is 0 and if number is 0 it returns 0 instead of number.

scorbett123: I don't believe that interval is ever 0, the only occurrence I could found was that it was 64.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,217 @@
...
+        long litematicSize = (long) this.x*this.y*this.z;
+
+        // In 1.12, the long array isn't exposed by the libraries so parsing has to be done manually
+        String rawBlockString = (nbt.getCompoundTag("Regions").getCompoundTag(regionName)).getTag("BlockStates").toString();
+        rawBlockString = rawBlockString.substring(3,rawBlockString.length()-1);
+        String[] rawBlockArrayString = rawBlockString.split(",");
+        long[] rawBlockData = new long[rawBlockArrayString.length];
+        for (int i = 0; i < rawBlockArrayString.length; i++) {
+            rawBlockData[i] = Long.parseLong(rawBlockArrayString[i].substring(0,rawBlockArrayString[i].length()-1));
+        }
```
ZacSharp, line 83: Since I hate parsing strings that aren't meant to be parsed, may I suggest using reflection or mixin to access the `data` attribute? I checked the litematica and malilib sources and the latter is what is used there.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.*;
+
+/**
+ * @author Emerson
+ * @since 12/27/2020
+ */
+public final class LitematicaSchematic extends StaticSchematic {
+
+    public LitematicaSchematic(NBTTagCompound nbt) {
+        String regionName = (String) nbt.getCompoundTag("Regions").getKeySet().toArray()[0];
```
ZacSharp, line 38: Why not iterate over all regions?
They are all part of the schematic and the issues would just be spammed with "Baritone not loading some litematic regions"

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
EmersonDove: Yes so that's the main reason why I have this as a draft pull because that iteration is so tacky. I'll look into those sources and try to find something.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
EmersonDove: Good point, I didn't do it because it involves reading the entire schematic into a larger volume. It makes more sense to add a region argument to the command?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
EmersonDove: I can test it. The nested class there is straight out of litematica, so I just left the code as-is.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
ZacSharp: I think it would be better to just keep the sub-regions, just like `CompositeSchematic` does it with it's sub-schematics.

EmersonDove: Sounds good, I'll look into that one too.

ZacSharp:
@EmersonDove are you still working on this?

EmersonDove:
@ZacSharp I've been on and off about it. I haven't fixed any the original bugs enough to commit. I talked to masa and he believes it should be possible but in general the 1.12.2 port has had a lot of issues.
I'll probably give it another try after the semester is over.

sudofox:
I ported this up to 1.17 a while ago here, did the work translating mappings etc

https://raw.githubusercontent.com/sudofox/baritone-1.17/master/src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java?token=GHSAT0AAAAAABKUY4TPGPDW7HIUI2M37YD2YQLEGKA

Unsure what needs to be done to push things forward

ZacSharp:
>  Issues with this version:
> *  Some states (such as iron bars) are unrecognized
> *  In 1.12, the long array (blockstates) isn't exposed by the libraries so parsing has to be done manually
> *  Only reads in first region of the .litematic

I't probably useable as is (never managed to get myself to have a look at it), but not ready for merging. The reason for the first point is unknown (as far as I know), the other two points have already been addressed in the discussion here.

EmersonDove:
Hey everyone looks like I should follow up here -

I wouldn't feel good merging it in with its current state - I would expect this will create GH issues than are closed.

I'm interested in pushing things forward for a 1.12 implementation with @sudofox but hit a wall with the mixins and have been working on other projects (@blankly-finance). I talked to Mesa a few months ago and there is no technical reason why this couldn't work.

If someone could tackle or give some pointers on the mixin for the `long` array that would be great.

ZacSharp:
For the mixin you probably want a simple mixin on `NBTTagLongArray` with an `@Accessor` for the data field. Since Baritone has the mixin classes in a separate source set you also need an interface for the accessor in the main source set and make the mixin class implement it. If you've got it right you can then safely cast the long array to the accessor interface and youse the mixin generated method to get the data. I think the easiest way to see how to do it is looking at the interfaces in `baritone.utils.accessor` and their implementations in `baritone.launch.mixins`.

sudofox:
Thanks for the mention Emerson. I should establish that my Java skills are quite hodgepodgey so I'll try to contribute as much as I can but otherwise I'll cheer you on from the sidelines 😅

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,217 @@
...
+        this.x = Math.abs(nbt.getCompoundTag("Regions").getCompoundTag(regionName).getCompoundTag("Size").getInteger("x"));
```
maruohon, line 39: The region size needs to be handled differently. The stored sizes per axis can be negative, if the "primary corner" (region origin" is not in the minimum corner of the region. (If they are negative, then they are at max -2, there should be no -1 values created by Litematica, as that is essentially the same as positive 1.)
image


I think for Baritone you can basically just use `Math.abs()` and then offset the origin if the values were negative for some axes. (No idea how you handle he positioning.) In Litematica this "freedom" in creating the selection with the primary corner in any relative corner causes quite some mess and complexity when handling the regions with rotations and mirroring etc.

ZacSharp: We always use the minimum corner for the position and don't have rotations, so just adding `size + 1` and then removing the sign from `size` should work fine.

ZacSharp:
I've had a look at it and I think you can copy the changes from https://github.com/ZacSharp/baritone/commit/0550fe362369d12e0aefe6ffb0b8f9190f2f7f69 as is. I also had a try at loading all regions of the schematic in https://github.com/ZacSharp/baritone/commit/3cf9e2f2580481350fcf1c4a1481164cde62b72d but that commit is not useable as is and comes with debug printouts and will likely (didn't try it) crash with map art mode enabled for any litematic that does completely fill its enclosing box volume.

ZacSharp:
I've invested some more time and used the approach from #3317 (patching the holes with air) to fix `mapArtMode`. You can find the changes at https://github.com/ZacSharp/baritone/commit/3661fea4f05755ebf285b883aabfe2abde34f76d and its parents in case you want to use them.

ThalusA:
What does this PR still need to get released?

amatheo:
News on why there is still no merge ?

ZacSharp:
What's currently missing (both done on my branch but @EmersonDove hasn't done it here yet)
* Read all regions of the .litematic instead of just the first
* Get rid of the hacky string parsing

@leijurv I'd really like to know whether static schematics are allowed to have holes (i.e. behavior of getDirect is undefined at positions within the schematic bounds as it is the case for other schematics).

scorbett123:
Superseded by #3672

PR #2567, opened 2021-03-13
Essential Needs for (Minecraft)

Yeetmyo:
You know when you try to build a ender portal by using a water. I suggest that Baritone should add a make portal command. that's not all, you know when you try to make a enchantment table and you need library books to make the enchantment table to become better at enchanting.  Baritone should make a enchatment made. This allowes the bot to get book shelfes and get make a enchantment table and build it. 

Get Item
In my opinon Baritone should add get item, for example if you want fireworks or any item, they need to get it for you. Not just blocks


scorbett123:
Wtf, why is this a pull?

PR #2622, opened 2021-03-29
update mixin dependency to newest version

wagyourtail:
update mixin version to `0.8.3`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,7 +91,7 @@ repositories {
 
 dependencies {
     runtime launchCompile('com.github.ImpactDevelopment:SimpleTweaker:1.2')
-    runtime launchCompile('org.spongepowered:mixin:0.7.11-SNAPSHOT') {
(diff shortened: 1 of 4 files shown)


leijurv:
we tried this before, a very long time ago. i truly cannot remember what went wrong, but a lot did

wagyourtail:
might need another commit, there's an issue with `ow2.asm` not having the right version so `java.lang.ClassNotFoundException: org.objectweb.asm.commons.ClassRemapper`

on a slightly unrelated note I also seem to not be able to build for forge btw, even before these changes. doing `-Pbaritone.forge_build` is still giving me notch mappings.

scorbett123:
Does this not break compatibility with like,  every other mod where baritone is built in? Such as impact etc.

l1ving:
yeah I posted a crash log on dependency bot's pull

wagyourtail:
the main reason for this is because forge is stupid and if multiple different versions of mixins are shaded and the other one isnt loaded first (ie with replaymod which is why adding ! To front works on forge) it crashes

Issue #2597, opened 2021-03-23
Blacklisted 'zones'


265866:
## Describe your suggestion
User could set different coordinate 'zones' that baritone would mine in

## Settings
#set disallowedZones (true or false)
#addzone (x1) (y1) (z1) (x2) (y2) (z2) (name)
#delzone (name)

## Context
Baritone won't dig through my own home, I know you can do blacklisted blocks but this would be much easier to do, also if you have a multi-building area, it wont dig up the terrain

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

scorbett123:
Similar to #667, maybe that one should be closed and this one left open as this is more fleshed out and has further ideas onto that. It should also be optional whether the zone is blacklisted (no mining inside it) or whitelisted (only mining inside it).

265866:
> Similar to #667, maybe that one should be closed and this one left open as this is more fleshed out and has further ideas onto that. It should also be optional whether the zone is blacklisted (no mining inside it) or whitelisted (only mining inside it).

Yeah, definitely, I’ll add those ideas to the suggestion

ZacSharp:
I think the commands should be
* one setting to toggle all zones on or of
* one to create zones (both allowed and disallowed) from selections
* one to remove a zone by name
* maybe one to toggle them by name?

The last 2/3 points should probably be subcommands of the same command; the one for creation could also be a subcommand of `sel` instead (and should be noted in `help sel` in any case)

Allowed and disallowed zones could coexist by either allowed zones overwriting disallowed zones or the other way round (another setting? a priority for each zone?).

PR #2076, opened 2020-10-12
🐛 Fix a Rendering Conflict with some other Mods

CDAGaming:
In testing within Fabritone, it was found that these two lines (Which were previously used for an unknown optifine workaround) are no longer needed, and had no known side effects as of writing.

This can be backported as needed, but details are unknown how far this can go down

Future me PS: This PR is a superseed of my last one, erased during repo refactors, so now it lives on master branch. Testing recommended and can be up-ported/down-ported as you wish.

PR #2661, opened 2021-04-13
Home protection

FloGo3308:
I permited baritone from placing and destroying blocks close to waypoints tagged with home using these settings:
- homeProtectionRange(Double, defaulted to 0.0): On what range baritone will "protect" home waypoints
- homeProtectionChatNotification(Boolean, defaulted to true): Send a chat notification if you enter or leave protected home area
- homeProtectionUpdateInterval(Integer, defaulted to 5): Interval at which baritone checks whether the player is close to a home tagged waypoint or not.(Interval is like mineGoalUpdateInterval, it checks if the amount of ticks is a multiple of the specified setting's value, so 1 means once every tick and 20 means once every 20 ticks/1 second)

I currently use BaritoneAPI.getProvider().getPrimaryBaritone() to get the baritone instance, wich may couse some issues in rare cases(would be nice if some-one could tell me how to fix :) )

Oh btw this is the first time ever I have contributed to a project on github, so please tell me if I did something wrong.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1196,7 +1196,7 @@ public final class Settings {
     /**
      * Disallow block placement and breaking within a range from the home waypoint(if it exists)
      */
-    public final Setting<Integer> homeProtectionUpdateInerval = new Setting<>(5);
+    public final Setting<Integer> homeProtectionUpdateInterval = new Setting<>(5);
 
     /**
      * A map of lowercase setting field names to their respective setting
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -17,8 +17,10 @@
 
 package baritone.api;
 
+import baritone.api.utils.NotificationHelper;
 import baritone.api.utils.SettingsUtil;
 import baritone.api.utils.TypeUtils;
+import baritone.api.utils.gui.BaritoneToast;
 import net.minecraft.block.Block;
 import net.minecraft.client.Minecraft;
 import net.minecraft.init.Blocks;
@@ -33,6 +35,7 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.function.BiConsumer;
 
 /**
  * Baritone's settings. Settings apply to all Baritone instances.
@@ -70,9 +73,9 @@ public final class Settings {
     public final Setting<Boolean> assumeExternalAutoTool = new Setting<>(false);
 
     /**
-     * If this setting is on, no auto tool will occur at all, not at calculation time nor execution time
+     * Automatically select the best available tool
      */
-    public final Setting<Boolean> disableAutoTool = new Setting<>(false);
+    public final Setting<Boolean> autoTool = new Setting<>(true);
 
     /**
      * It doesn't actually take twenty ticks to place a block, this cost is so high
@@ -596,6 +599,12 @@ public final class Settings {
     public final Setting<Boolean> renderGoal = new Setting<>(true);
 
     /**
+     * Render the goal as a sick animated thingy instead of just a box
+     * (also controls animation of GoalXZ if {@link #renderGoalXZBeacon} is enabled)
+     */
+    public final Setting<Boolean> renderGoalAnimated = new Setting<>(true);
+
+    /**
      * Render selection boxes
      */
     public final Setting<Boolean> renderSelectionBoxes = new Setting<>(true);
@@ -848,6 +857,11 @@ public final class Settings {
     public final Setting<Boolean> skipFailedLayers = new Setting<>(false);
 
     /**
+     * Only build the selected part of schematics
+     */
+     public final Setting<Boolean> buildOnlySelection = new Setting<>(false);
+
+    /**
      * How far to move before repeating the build. 0 to disable repeating on a certain axis, 0,0,0 to disable entirely
      */
     public final Setting<Vec3i> buildRepeat = new Setting<>(new Vec3i(0, 0, 0));
@@ -1057,6 +1071,20 @@ public final class Settings {
      * {@link Setting#value};
      */
     public final Setting<Consumer<ITextComponent>> logger = new Setting<>(Minecraft.getMinecraft().ingameGUI.getChatGUI()::printChatMessage);
+
+    /**
+     * The function that is called when Baritone will send a desktop notification. This function can be added to
+     * via {@link Consumer#andThen(Consumer)} or it can completely be overriden via setting
+     * {@link Setting#value};
+     */
+    public final Setting<BiConsumer<String, Boolean>> notifier = new Setting<>(NotificationHelper::notify);
+
+    /**
+     * The function that is called when Baritone will show a toast. This function can be added to
+     * via {@link Consumer#andThen(Consumer)} or it can completely be overriden via setting
+     * {@link Setting#value};
+     */
+    public final Setting<BiConsumer<ITextComponent, ITextComponent>> toaster = new Setting<>(BaritoneToast::addOrUpdate);
 
     /**
      * The size of the box that is rendered when the current goal is a GoalYLevel
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1189,6 +1189,11 @@ public final class Settings {
     public final Setting<Double> homeProtectionRange = new Setting<>(0.0);
 
     /**
+     * Sends a chat notification to let you know when you enter or leave protected home area, can be quite annoying when you're working close to a home waypoint
+     */
+    public final Setting<Boolean> homeProtectionChatNotification = new Setting<>(true);
+
+    /**
      * Disallow block placement and breaking within a range from the home waypoint(if it exists)
      */
     public final Setting<Integer> homeProtectionUpdateInerval = new Setting<>(5);
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1189,17 +1189,6 @@ public final class Settings {
     public final Setting<Integer> homeProtectionRange = new Setting<>(-1);
 
     /**
-     * Sends a chat notification to let you know when you enter or leave protected home area
-     * Can be quite annoying when you're working close to a home waypoint
-     */
-    public final Setting<Boolean> homeProtectionChatNotification = new Setting<>(true);
-
-    /**
-     * Disallow block placement and breaking within a range from the home waypoint(if it exists)
-     */
-    public final Setting<Integer> homeProtectionUpdateInterval = new Setting<>(5);
-
-    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1186,10 +1186,11 @@ public final class Settings {
     /**
      * Disallow block placement and breaking within a range from the home waypoint(if it exists)
      */
-    public final Setting<Double> homeProtectionRange = new Setting<>(0.0);
+    public final Setting<Integer> homeProtectionRange = new Setting<>(-1);
 
     /**
-     * Sends a chat notification to let you know when you enter or leave protected home area, can be quite annoying when you're working close to a home waypoint
+     * Sends a chat notification to let you know when you enter or leave protected home area
+     * Can be quite annoying when you're working close to a home waypoint
      */
     public final Setting<Boolean> homeProtectionChatNotification = new Setting<>(true);
 
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1184,6 +1184,16 @@ public final class Settings {
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
     /**
+     * Disallow block placement and breaking within a range from the home waypoint(if it exists)
+     */
+    public final Setting<Double> homeProtectionRange = new Setting<>(0.0);
+
+    /**
+     * Disallow block placement and breaking within a range from the home waypoint(if it exists)
+     */
+    public final Setting<Integer> homeProtectionUpdateInerval = new Setting<>(5);
+
+    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/api/java/baritone/api/behavior/IPathingBehavior.java
+++ b/src/api/java/baritone/api/behavior/IPathingBehavior.java
@@ -131,17 +131,6 @@ public interface IPathingBehavior extends IBehavior {
     IPathExecutor getNext();
 
     /**
-     * Returns true when close to home.
-     * @return true when close to home
-     */
-    boolean isCloseToHome();
-
-    /**
-     * Updates @isCloseToHome
-     */
-    void updateIsCloseToHome();
-
-    /**
      * Returns true when a block is close to home.
      * @return true when a block is close to home
      */
--- a/src/api/java/baritone/api/behavior/IPathingBehavior.java
+++ b/src/api/java/baritone/api/behavior/IPathingBehavior.java
@@ -140,4 +140,10 @@ public interface IPathingBehavior extends IBehavior {
      * Updates @isCloseToHome
      */
     void updateIsCloseToHome();
+
+    /**
+     * Returns true when a block is close to home.
+     * @return true when a block is close to home
+     */
+    boolean isCloseToHome(int x,  int y, int z);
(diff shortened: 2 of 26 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -1183,6 +1183,21 @@
      */
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
+    /**
+     * Disallow block placement and breaking within a range from the home waypoint(if it exists)
+     */
+    public final Setting<Double> homeProtectionRange = new Setting<>(0.0);
+
+    /**
+     * Sends a chat notification to let you know when you enter or leave protected home area, can be quite annoying when you're working close to a home waypoint
```
ZacSharp, line 1192: You should split really long lines like this one into multiple lines, so they fit onto (most) screens.

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -565,4 +577,49 @@ private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal
...
+
+    @Override
+    public boolean isCloseToHome() {
+        return isCloseToHome;
+    }
+
+    @Override
+    public void updateIsCloseToHome() {
+        try {
+            IBaritone bar = BaritoneAPI.getProvider().getPrimaryBaritone();
```
ZacSharp, line 589: `PathingBehavior` extends from `Behavior`, which has a `public final Baritone baritone`. Use that here.

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -565,4 +577,49 @@ private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal
...
+            EntityPlayerSP player = bar.getPlayerContext().player();
+            IWaypoint[] waypoints = ForWaypoints.getWaypointsByTag(bar, IWaypoint.Tag.HOME);
+            if (waypoints.length > 0) {
+                for (IWaypoint wp : waypoints) {
+                    boolean isClose = wp.getLocation().getDistance((int) Math.round(player.posX), (int) Math.round(player.posY), (int) Math.round(player.posZ)) <= Baritone.settings().homeProtectionRange.value;
+                    if (isClose) {
+//                    Helper.HELPER.logDirect("true");
```
ZacSharp, line 596: Remove debugging code in comments (ideally before committing)

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -565,4 +577,49 @@ private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal
...
+                        isCloseToHome = true;
+                        logProtectedHomeArea();
+                        return;
+                    }
+                }
+            }
+//        Helper.HELPER.logDirect("false");
```
ZacSharp, line 603: more debugging code

src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -136,7 +136,9 @@ public Block getBlock(int x, int y, int z) {
     }
 
     public double costOfPlacingAt(int x, int y, int z, IBlockState current) {
-        if (!hasThrowaway) { // only true if allowPlace is true, see constructor
+        if (!hasThrowaway || baritone.getPathingBehavior().isCloseToHome()) { /** only true if allowPlace is true, see constructor
+                                                                                * now also checks if you're not too close too home
+                                                                                */
```
ZacSharp, line 141: I think multi-line comments are better done above the line they belong to

src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -136,7 +136,9 @@ public Block getBlock(int x, int y, int z) {
     }
 
     public double costOfPlacingAt(int x, int y, int z, IBlockState current) {
-        if (!hasThrowaway) { // only true if allowPlace is true, see constructor
+        if (!hasThrowaway || baritone.getPathingBehavior().isCloseToHome()) { /** only true if allowPlace is true, see constructor
```
ZacSharp, line 139: You should either cache this (like `hasThrowaway`) so it can't change during pathing, or make it actually use the x,y,z parameters to depend on the block being checked instead of the player position

src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -150,7 +152,7 @@ public double costOfPlacingAt(int x, int y, int z, IBlockState current) {
     }
 
     public double breakCostMultiplierAt(int x, int y, int z, IBlockState current) {
-        if (!allowBreak) {
+        if (!allowBreak || baritone.getPathingBehavior().isCloseToHome()) {
```
ZacSharp, line 155: same here
either cached or block dependant

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -911,7 +912,8 @@ public double costOfPlacingAt(int x, int y, int z, IBlockState current) {
 
         @Override
         public double breakCostMultiplierAt(int x, int y, int z, IBlockState current) {
-            if (!allowBreak || isPossiblyProtected(x, y, z)) {
+//            baritone.getPathingBehavior().updateIsCloseToHome();
```
ZacSharp, line 915: remove this

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -911,7 +912,8 @@ public double costOfPlacingAt(int x, int y, int z, IBlockState current) {
...
+            if (!allowBreak || isPossiblyProtected(x, y, z) || baritone.getPathingBehavior().isCloseToHome()) {
```
ZacSharp, line 916: and another cache-or-per-block

and why is pathing disallowed to place or break near home waypoints, while the builder may place but not break there? That's a bit inconsistent

src/main/java/baritone/process/MineProcess.java
```diff
@@ -103,8 +104,9 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
                 return null;
             }
         }
-        if (!Baritone.settings().allowBreak.value) {
-            logDirect("Unable to mine when allowBreak is false!");
+//        baritone.getPathingBehavior().updateIsCloseToHome();
+        if (!Baritone.settings().allowBreak.value || baritone.getPathingBehavior().isCloseToHome()) {
+            logDirect("Unable to mine when allowBreak is false or when you\'re too close to home!");
```
ZacSharp, line 109: If you do the checks in the CalculationContexts per block, then you should remove this check and replace it by blacklisting any blocks near a home. If you decide to cache whether the player is near a home for pathing, then keep this. Just so it's consistent.

src/main/java/baritone/process/MineProcess.java
```diff
@@ -467,8 +469,9 @@ public void mineByName(int quantity, String... blocks) {
     @Override
     public void mine(int quantity, BlockOptionalMetaLookup filter) {
         this.filter = filter;
-        if (filter != null && !Baritone.settings().allowBreak.value) {
-            logDirect("Unable to mine when allowBreak is false!");
+//        baritone.getPathingBehavior().updateIsCloseToHome();
```
ZacSharp, line 472: remove

src/main/java/baritone/process/MineProcess.java
```diff
@@ -467,8 +469,9 @@ public void mineByName(int quantity, String... blocks) {
...
+        if (filter != null && (!Baritone.settings().allowBreak.value || baritone.getPathingBehavior().isCloseToHome())) {
+            logDirect("Unable to mine when allowBreak is false or when you\'re too close to home!");
```
ZacSharp, line 474: This should as well be removed depending on when home protection is checked for pathing.

src/api/java/baritone/api/Settings.java
```diff
@@ -1183,6 +1183,21 @@
...
+     */
+    public final Setting<Boolean> homeProtectionChatNotification = new Setting<>(true);
```
ZacSharp, line 1194: Why do we need chat notifications (and even a toggle setting for them) here? If you can see the chat message, you can see what Baritone is doing anyway. I think turning the chat notifications into desktop notifications would make more sense, but even then I'm not sure if we really need it.

src/api/java/baritone/api/Settings.java
```diff
@@ -1183,6 +1183,21 @@
      */
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
+    /**
+     * Disallow block placement and breaking within a range from the home waypoint(if it exists)
+     */
+    public final Setting<Double> homeProtectionRange = new Setting<>(0.0);
```
ZacSharp, line 1189: Why not an Integer? I don't think people will need sub-block precision for their home protections and ints are easier to deal with when checking whether Baritone is inside that range.

ZacSharp (changes requested):
This is btw superseeded by #2597, which I like better because it is more flexible and could easily be expanded to allow/disallow entering/leaving regions.

Didn't notice anything wrong GitHub wise.

src/api/java/baritone/api/Settings.java
FloGo3308: I thought it would be handy for people who have accidently turned the protection range up and are wondering why baritone "doesn't work"

src/api/java/baritone/api/Settings.java
FloGo3308: Yeah a double would prob be a bit overkill for such a task

src/api/java/baritone/api/Settings.java
FloGo3308: Yeah that would be better

src/main/java/baritone/process/BuilderProcess.java
FloGo3308: What? I thought I builder couldn't place either...

src/main/java/baritone/pathing/movement/CalculationContext.java
scorbett123: Make it so that it checks the xyz, not player position. Baritone would currently mine towards home, and then after entering home, would have to recalculate path, or even may just continue on the current path.

FloGo3308: ih yeah, that really has to be fixed asap

src/main/java/baritone/process/BuilderProcess.java
FloGo3308: Ahhhhhh, BuilderProcess uses CalculationContext's hasThrowaway, I just searched for allowPlace... -_-

src/api/java/baritone/api/Settings.java
FloGo3308: Changed in last commit

src/api/java/baritone/api/Settings.java
FloGo3308: Changed in latest commit

src/main/java/baritone/behavior/PathingBehavior.java
FloGo3308: Changed in latest commit

src/main/java/baritone/behavior/PathingBehavior.java
FloGo3308: Changed in latest commit

src/main/java/baritone/behavior/PathingBehavior.java
FloGo3308: Changed in latest commit

src/main/java/baritone/pathing/movement/CalculationContext.java
FloGo3308: Changed in latest commit

src/main/java/baritone/pathing/movement/CalculationContext.java
FloGo3308: Changed in latest commit

src/main/java/baritone/pathing/movement/CalculationContext.java
FloGo3308: Changed in latest commit

src/main/java/baritone/process/BuilderProcess.java
FloGo3308: Changed in latest commit

src/main/java/baritone/process/MineProcess.java
FloGo3308: The comment? Think I have removed it in my lateat commit

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -565,4 +577,66 @@ private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal
...
+            if (waypoints.length > 0) {
+                for (IWaypoint wp : waypoints) {
+                    boolean isClose = wp.getLocation().getDistance(x, y, z) <= Baritone.settings().homeProtectionRange.value;
+                    if (isClose) {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
```
ZacSharp, line 624: Make this a method of `CalculationContext` and cache the waypoint *positions* as well as `homeProtectionRangeSq` (sqared radius) in the constructor, just like it is done for settings and other things that can change while pathing.
Then compare the summed squares to the squared distance instead of taking the square root every time (it's hidden in `wp.getLocation().getDistance(x, y, z)`)

ZacSharp (changes requested):
Next time please just press the "resolve conversation" button instead of commenting "Changed in latest commit" eleven times. Everyone that commented on this pr (and everyone watching the repo I think) already receives a notification for the new commit, there's no need to spam them with redundant messages.

FloGo3308:
> Next time please just press the "resolve conversation" button instead of commenting "Changed in latest commit" eleven times. Everyone that commented on this pr (and everyone watching the repo I think) already receives a notification for the new commit, there's no need to spam them with redundant messages.

Oh my bad, sorry, i'm completely new to github :/

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -565,4 +577,66 @@ private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal
...
+        catch (Exception err) {
+            return false;
+        }
```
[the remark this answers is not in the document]
FloGo3308: Should I log them to chat using some kind of baritone exception?

src/api/java/baritone/api/behavior/IPathingBehavior.java
```diff
@@ -129,4 +129,21 @@ default boolean hasPath() {
...
+    /**
+     * Returns true when close to home.
+     * @return true when close to home
+     */
+    boolean isCloseToHome();
+
+    /**
+     * Updates @isCloseToHome
+     */
+    void updateIsCloseToHome();
```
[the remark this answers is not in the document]
FloGo3308: Ok

src/main/java/baritone/behavior/PathingBehavior.java
leijurv: Yeah, what zacsharp said is all correct.

Make local fields of `CalculationContext`. Something like `boolean homeProtectionEnabled` and `int homeProtectionRange`. I could be convinced of just `int homeProtectionRange` and `if (homeProtectionRange == -1) {` everywhere, but since this is **such** a core inner function, I would Feel Safer if the entire thing was shoved under a nice little `if (homeProtectionEnabled) {` in `CalculationContext.costOfPlacingAt` and `CalculationContext.breakCostMultiplierAt`. Then, when that boolean is false, it's nothing but a never-taken branch, which is basically free.

It might be interesting to do something similar to the `Favoring` system for avoiding mobs where it generates spheres and stores them in a fastutil hashset. It would depend on how big the radius is and how many home waypoints you have. Gut feeling is that this is correct like this and you shouldn't do the `Favoring` approach.

leijurv:
> This is btw superseeded by #2597, which I like better because it is more flexible and could easily be expanded to allow/disallow entering/leaving regions.


@ZacSharp https://github.com/cabaletta/baritone/pull/2076#issuecomment-708571322

scorbett123:
> > This is btw superseeded by #2597, which I like better because it is more flexible and could easily be expanded to allow/disallow entering/leaving regions.
> 
> @ZacSharp [#2076 (comment)](https://github.com/cabaletta/baritone/pull/2076#issuecomment-708571322)

lol

ZacSharp:
> > > This is btw superseeded by #2597, which I like better because it is more flexible and could easily be expanded to allow/disallow entering/leaving regions.
> > 
> > 
> > @ZacSharp [#2076 (comment)](https://github.com/cabaletta/baritone/pull/2076#issuecomment-708571322)
> 
> lol

makes sense I guess

FloGo3308:
Deleted this pr cuz im not interested in maintaining this pr anymore.

PR #2664, opened 2021-04-13
Add setting to only build selected part of schematica schematic

### LINKED ISSUE #2612: #schematica should optionally only build the selection
## Describe your suggestion
When running `#schematica` I think it could be poggers if it built only the little bit that is `#sel`'d.

## Settings
Idk maybe controlled by `schematicaOnlyBuildSelection`?

## Context
This would be super convenient to build small parts of large schematics so that Baritone doesn't spaz out and say "unable to do it" when there's random blocks that it doesn't know how to do

## Final checklist

ZacSharp:
I have made it like it was suggested in #2612, but I'm not really happy with it because
* I think it should be useable for `#build` as well (easy with the current implementation, just have to copy the logic to the respective method)
* I don't like the fact that it is a setting functioning as an argument (`#schematica <selectionOnly` but what signature would `#build` have?)

and less important
* I don't like the fact that it is a schematic knowing it's own (initial) position, but can't think of a cleaner solution
* a better name would be nice

closes #2612

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -848,6 +848,11 @@ public final class Settings {
     public final Setting<Boolean> skipFailedLayers = new Setting<>(false);
 
     /**
+     * Only build the selected part of the schematic when using #schematica
+     */
+     public final Setting<Boolean> schematicaOnlyBuildSelection = new Setting<>(false);
+
+    /**
      * How far to move before repeating the build. 0 to disable repeating on a certain axis, 0,0,0 to disable entirely
      */
     public final Setting<Vec3i> buildRepeat = new Setting<>(new Vec3i(0, 0, 0));
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -848,9 +848,9 @@ public final class Settings {
     public final Setting<Boolean> skipFailedLayers = new Setting<>(false);
 
     /**
-     * Only build the selected part of the schematic when using #schematica
+     * Only build the selected part of schematics
      */
-     public final Setting<Boolean> schematicaOnlyBuildSelection = new Setting<>(false);
+     public final Setting<Boolean> buildOnlySelection = new Setting<>(false);
 
     /**
      * How far to move before repeating the build. 0 to disable repeating on a certain axis, 0,0,0 to disable entirely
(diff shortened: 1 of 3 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -847,6 +847,11 @@
      */
     public final Setting<Boolean> skipFailedLayers = new Setting<>(false);
 
+    /**
+     * Only build the selected part of the schematic when using #schematica
+     */
+     public final Setting<Boolean> schematicaOnlyBuildSelection = new Setting<>(false);
```
scorbett123, line 853: Maybe something like buildOnlySelection would be better.

scorbett123 (commented):
Add this to the #build command as well, I think that that is really what leijurv wanted / implies.

ZacSharp:
Should it be applied when `mapArtMode` is applied (Files and Schematica) or for any schematic (including `#sel`, where it's a no-op, and I think even custom tunnels, which might be annoying)? I'd do it for anything, but I'd like it to be consistent with `mapArtMode` as well. 

Anything to say on the setting vs. argument point?

src/api/java/baritone/api/Settings.java
ZacSharp: Sure, I will change it tomorrow.

scorbett123:
I would say try and replicate map art mode as close as possible, but with just different behaviour.

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -150,10 +151,13 @@ public void buildOpenSchematic() {
             Optional<Tuple<IStaticSchematic, BlockPos>> schematic = SchematicaHelper.getOpenSchematic();
             if (schematic.isPresent()) {
                 IStaticSchematic s = schematic.get().getFirst();
+                BlockPos origin = schematic.get().getSecond();
+                ISchematic schem = Baritone.settings().mapArtMode.value ? new MapArtSchematic(s) : s;
+                schem = Baritone.settings().schematicaOnlyBuildSelection.value ? new SelectionSchematic(schem, origin, baritone.getSelectionManager().getSelections()) : schem;
```
scorbett123, line 156: Maybe move this to a separate function, this would mean that the same code could be used in both buildOpenSchematic and buildSchematic.

ZacSharp: Yea, thought about this and then decided that refactoring other people's code is not what this pr is there for.

leijurv:
Thanks!

leijurv:
Lol

PR #2721, opened 2021-05-07
Make baritone mor familiar with frost walker/boats

Ee0Rk:
When travelling var distances I will most likely travel over oceans/water, and so I try to put my self in a boat and turn on baritone(barely works). Or I will put on frost walker boots (Works poorly). So all I want is to put it on boats automatically and have it work in the ocean and/or maybe put on a frost walker automatically and then work on ice. When walking on normally generated ice it works just fine, but when walking on ice generated by a frost walker it always overshoots and recalcualtes. Please I (we) really need this.


ZacSharp:
If you want to make a suggestion open an issue, not a pull request.

Issue #2598, opened 2021-03-23
Discord Updates


265866:
## Describe your suggestion
Same as desktop notifications but sends to discord webhook

## Settings
#set discordNotifications (true/false)
#set webhook (webhook link)
#set discordChannelID (channel id you want it to be added to)

## Context
some people leave their PC running and aren't at their desktop to receive notifications

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

ZacSharp:
Maybe just a (hidden) `notifications` setting like `logger` so one can easily write an add-on that sends notifications to any platform?

265866:
> Maybe just a (hidden) `notifications` setting like `logger` so one can easily write an add-on that sends notifications to any platform?

I'm not entirely sure how that'd work, correct me if I'm wrong, but don't different platforms work differently? I mean if that would work that would be amazing but I'm just not entirely sure how that work haha

ZacSharp:
There would just be a setting accessible from java only with type `Consumer<ITextComponent>`.
A mod can then simply place it's own function there and do whatever it wants to the message (e.g. send it to discord or just modify it before forwarding to the previously set notification logger)

bokja:
Yeah this would be nice, just a logger that logs information to a txt and then you can hook in from any self made application.

FloGo3308:
> > Maybe just a (hidden) `notifications` setting like `logger` so one can easily write an add-on that sends notifications to any platform?
> 
> I'm not entirely sure how that'd work, correct me if I'm wrong, but don't different platforms work differently? I mean if that would work that would be amazing but I'm just not entirely sure how that work haha

I think what he means is that baritone implements some sort of thing that gives other mods acces to read the last notification(s) that baritone sent, so that other mod creators can use it to make a mod that posts it to the platform they want.
So, baritone makes the last notification(s) readable for other mods, and other mod developers read it and can do anything they want with it.

bjoernp116:
I like the idea!

ZacSharp:
I already tried implementing this, but `NotificationHelper` is a non-api class (which imo is correct), and to use `NotificationHelper#notify` as the default notification callback I have to access it from the api. Should I just move `baritone.utils.NotificationHelper` to `baritone.api.utils.NotificationHelper` or does someone know another clean(er) solution?

ZacSharp:
Almost forgot about this issue.
[...]

Issue #1593, opened 2020-04-19
Setting the Logger


Infinidoge:
## What do you need help with?
I'm not sure if I'm being stupid here, but I'm trying to change the chat logger from:
```java
Minecraft.getMinecraft().ingameGUI.getChatGUI()::printChatMessage
```
to
```java
Minecraft.getMinecraft().player::sendMessage
```
(based off of my research this should be valid, but I'm not quite sure)

in my `settings.txt` file, I added:
```
logger Minecraft.getMinecraft().player::sendMessage
```
However, whenever the game gets started up, this line gets removed (not sure if this is due to Impact or baritone, though)

And whenever I set it in-game, it spits out an error: [log](https://paste.ee/p/Vshld) (at the bottom)

Am I setting the logger correctly, and is what I'm trying to set it to valid?
Additionally, if it's impact trying to remove the line, how would I stop it from doing so?

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

Infinidoge:
Upon further inspection, the error I received happens regardless of what I put into the command to set it in game.

ZacSharp:
I'm surprised no one answered this.
The can only be changed using Baritones Java API and is never saved or loaded from saved settings.

PR #2771, opened 2021-05-24
More log callbacks

ZacSharp:
This makes the notification function a setting called `notifier` just like logging to chat uses `logger` and because I accidentally did it for toasts instead of notifications at first (don't trust the commit history), toast now have a setting called `toaster` (but it can't roast bread :disappointed:).
This makes it possible for external mods to implement #2598.

Additionally I replaced the unhandled excpetion when accessing these settings with commands with a more descriptive error message, to avoid issues like #1593

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -17,8 +17,10 @@
 
 package baritone.api;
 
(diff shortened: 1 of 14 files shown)


leijurv:
Nice, this is a very welcome fix!

PR #2778, opened 2021-05-28
Added the waypointOnDeath setting so you can disable the creation of waypoints when you die.

FloGo3308:
Added the waypointOnDeath setting so you can disable the creation of waypoints when you die.
They are not needed in some situations, especially when you use the same launcher profile for anarchy ánd crystal pvp servers(they are a complete waste of time and, although waypoint data is small, storage space when you die over and over and over again).

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1223,6 +1223,11 @@ public final class Settings {
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
     /**
+     * Creates a waypoint where you die
+     */
+    public final Setting<Boolean> waypointOnDeath = new Setting<>(true);
+
+    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/main/java/baritone/behavior/MemoryBehavior.java
+++ b/src/main/java/baritone/behavior/MemoryBehavior.java
@@ -174,25 +174,43 @@ public final class MemoryBehavior extends Behavior {
 
     @Override
     public void onPlayerDeath() {
-        Waypoint deathWaypoint = new Waypoint("death", Waypoint.Tag.DEATH, ctx.playerFeet());
-        baritone.getWorldProvider().getCurrentWorld().getWaypoints().addWaypoint(deathWaypoint);
-        ITextComponent component = new TextComponentString("Death position saved.");
-        component.getStyle()
-                .setColor(TextFormatting.WHITE)
-                .setHoverEvent(new HoverEvent(
-                        HoverEvent.Action.SHOW_TEXT,
-                        new TextComponentString("Click to goto death")
-                ))
-                .setClickEvent(new ClickEvent(
-                        ClickEvent.Action.RUN_COMMAND,
-                        String.format(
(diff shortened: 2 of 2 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -1211,6 +1211,11 @@
      */
     public final Setting<Boolean> notificationOnMineFail = new Setting<>(true);
 
+    /**
+     * Creates a waypoint where you die
+     */
+    public final Setting<Boolean> waypointOnDeath = new Setting<>(true);
```
ZacSharp, line 1217: If the goal is to prevent unnecessary waypoints `automaticWaypoints` for all automatically created waypoints would also be an option, but the only other type Baritone creates itself are beds and you usually don't come across beds on pvp servers (unless you are playing bedwars but then you shouldn't be using Baritone).

ZacSharp (approved):

src/api/java/baritone/api/Settings.java
FloGo3308: I just want to disable death waypoints, I find all the other automatic waypoints verry usefull(and I cant think of a situation where you would want to disable other automatic waypoints tbh)

ZacSharp: Then leave it as is, I just wanted to mention it.

FloGo3308: Ye ik, but just so u know my goal is to stop death waypoints and not all waypoints :)

scorbett123 (approved):

leijurv:
If this a setting that you have to explicitly turn off, I imagine that you wouldn't really want anything to be printed in chat when you die? I'm not sure I see the use case for disabling the saving of the waypoint, and disabling the chat clickable link to path to the location, but still send a message asking if you want to reenable death waypoints.

IMO if you turn off Baritone doing stuff when you die, it should fully turn off Baritone doing stuff when you die.

FloGo3308:
> If this a setting that you have to explicitly turn off, I imagine that you wouldn't really want anything to be printed in chat when you die?

I thought it would be handy for those who forgot they changed the setting.

> I'm not sure I see the use case for disabling the saving of the waypoint, and disabling the chat clickable link to path to the location, but still send a message asking if you want to reenable death waypoints.

So you get notified that it didn't save(so people wont think its a bug), and the point of this pr is not to stop the chat notification when you die, but stop the death waypoints from clogging up.

> IMO if you turn off Baritone doing stuff when you die, it should fully turn off Baritone doing stuff when you die.

https://github.com/cabaletta/baritone/pull/2778#discussion_r641834263
and btw, what other stuff doesw baritone do when you die?

leijurv:
Hm... All right I guess. If there will be a chat message, let's at least make the message still have the goto link. So something like "No death position saved. Click here to goto death."

> what other stuff doesw baritone do when you die?

Nothing, I was just talking about this PR changing it from doing something to just talking in chat (which is still something).

FloGo3308:
> Hm... All right I guess. If there will be a chat message, let's at least make the message still have the goto link. So something like "No death position saved. Click here to goto death."

good idea! i'll change it when im home

FloGo3308:
Uhhhh, whoops, totaly forgot about this pr, ill make the changes later this day when I return from vacation

PR #2796, opened 2021-06-09
Added itemSaverThreshold setting

mankool0:
This setting allows you to change the durability at which baritone should stop using a tool when using the itemSaver setting. Currently baritone will stop using the tool when it's at 0 durability which sometimes breaks it. I've set the default value to 10 to give a little buffer to avoid tools breaking.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -734,6 +734,11 @@ public final class Settings {
     public final Setting<Boolean> itemSaver = new Setting<>(false);
 
     /**
+     * Durability to leave on the tool when using itemSaver
+     */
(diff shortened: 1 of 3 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -733,6 +733,11 @@
      */
     public final Setting<Boolean> itemSaver = new Setting<>(false);
 
+    /**
+     * Durability to leave on the tool when using itemSaver
+     */
+    public final Setting<Integer> itemSaverThreshold = new Setting<>(10);
```
ZacSharp, line 739: Ideally this would be one setting (-1 ≙ itemSaver off) but it's a boolean already and I don't think you should change that.
TLDR: imo you've don it right

ZacSharp (approved):

PR #2800, opened 2021-06-12
Remove broken hit count badge

scorbett123:
Remove this broken badge. I haven't seen it working at all in the past few weeks.


leijurv:
Really sad. I think their site just went down.

Issue #2764, opened 2021-05-20
Distribute Proguard Mappings with future builds


wagyourtail:
## Describe your suggestion
distribute the mappings proguard generates with future releases of baritone.

## Context
this would allow for easier reading of the api/standalone crash reports and stack traces by actually being able to tell where they come from.

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

PR #2801, opened 2021-06-12
Add github actions.

scorbett123:
Add github actions.

This also runs the tests and will output the mapping.txt file so that it can be used for debugging. 

Further as travisci.org is now removed from the project its remaining files can be removed.

This also closes: #2764


scorbett123:
This also closes: #2764

leijurv:
wow sad that travis would do me like this

PR #2803, opened 2021-06-12
Make farm use inventory when allowInventory is enabled.

scorbett123:
Allow #farm to retrieve items from your inventory.


ZacSharp (approved):

src/main/java/baritone/behavior/InventoryBehavior.java
```diff
@@ -152,6 +153,10 @@ public boolean selectThrowawayForLocation(boolean select, int x, int y, int z) {
     }
 
     public boolean throwaway(boolean select, Predicate<? super ItemStack> desired) {
+        return throwaway(select, desired, BaritoneAPI.getSettings().allowInventory.value);
```
leijurv, line 156: `Baritone.settings()` instead of `BaritoneAPI.getSettings()`

PR #2805, opened 2021-06-13
Add setting profiles.

### LINKED ISSUE #2196: Ability to save/load modified settings
## Describe your suggestion
ability to have preset modifications that you can switch between instantaneously.

## Settings
override presets, override unedited settings

## Context
Between clearing areas, building schematics, and auto running to different locations there are several different settings you may want to change. Instead of needing to change several settings everytime, could there be a way to save presets?
A simple example is, when I'm building a schematic I want my buildbylayers to work from bottom to top, and when I'm clearing an area I usually want the opposite. There are other settings too that I change between the two. Instead of needing to change each setting everytime, it would be nice to do something like ".b mode building" that changes 6 different settings automatically, then after I could switch to ".b mode terraforming" to flatten some mountains by having those 6 settings be set different to work better in the other context.

## Final checklist

scorbett123:
Adds setting profiles.

Managed by the commands:

#settingProfiles - list setting profiles
#settingProfile <profile> - switch to another profile
#settingProfile create <profile> - create a new profile.

closes #2196

### DIFF
--- a/src/api/java/baritone/api/utils/SettingsUtil.java
+++ b/src/api/java/baritone/api/utils/SettingsUtil.java
@@ -27,16 +27,15 @@ import net.minecraft.util.math.Vec3i;
 import java.awt.*;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
+import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
-import java.util.ArrayList;
+import java.util.*;
 import java.util.List;
-import java.util.Map;
-import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


ZacSharp:
I think you should keep the normal constant settings location and only save to profiles when explicitly requested by the user. That way they won't edit their profile because they temporarily need some other value for a setting.

I'll also use this opportunity to just get rid of some thoughts on settings, so here we go:
It would be cool to have some way to temporarily override them without the value being saved to disk at all.
It would be cool to have a layered profile system were some profiles automatically activate or deactivate on certain events (e.g. being at home could force `allowPlace` and `allowBreak` to false until you leave, different servers could have different settings as well and the gamemode based settings could use this as well).
Also I'd really like to have a distinction between profiles that contain all settings (current behaviour) and such that only override settings that were explicitly set for them (otherwise the layering and automatic toggling is pointless)

ZacSharp (changes requested):
And here's a serious review:
The current behaviour is really weird and will only apply settings that are different from their default value. This can for example cause partial application in case parseAndApply fails

scorbett123:
> I think you should keep the normal constant settings location and only save to profiles when explicitly requested by the user. That way they won't edit their profile because they temporarily need some other value for a setting.

I feel like all of the commands might get quite complicated, however that could be a way forwards. What is really needed is a rework of the setting system. 

Setting categories could also be a part of this, setting descriptions (current javadoc) would also be useful (maybe displayed on hover or something like that).

scorbett123:
> And here's a serious review:
> The current behaviour is really weird and will only apply settings that are different from their default value. This can for example cause partial application in case parseAndApply fails

Isn't that an issue with existing setting behaviour though. Only settings that are different from their default value are saved. Also parseAndApply only fails when there is invalid settings after the setting file has been edited separately, therefore it is the users fault. I agree it would be better if it wasn't, however I don't really think that that is part of this pull. I feel like the full thing just needs a rework, new and better settings.

ZacSharp:
Yea, agree on the rework thing. But I just don't know how to do it.
The thing with only saving when the user requests it is actually pretty easy:
1. Add a `profileSave` command that saves all settings and copies `settings.txt` to `settings-<profile>.txt`
2. Add a `profileLoad` command that resets all settings, copies `settings-<profile>.txt` to `settings.txt` and reloads all settings

I think the "resets all settings" part from 2. is the solution to the problem I mentioned in my review as well.

PR #2809, opened 2021-06-14
Cleanup github actions

FloGo3308:
Just making it a little more consistent.
And `    - name: Set up JDK 9` was wrong cuz it runs on java 8, not 9

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 9
+    - name: Set up JDK 8
       uses: actions/setup-java@v2
       with:
         java-version: '8'
@@ -23,4 +23,3 @@ jobs:
       
     - name: Executing tests
       run: ./gradlew test
-    
--- a/.github/workflows/workflow_test.yml
+++ b/.github/workflows/workflow_test.yml
@@ -1,5 +1,4 @@
-
-name: Tests
+name: Test
 
 on:
   push:
@@ -17,9 +16,9 @@ jobs:
       with:
         java-version: '8'
         distribution: 'adopt'
-        
+
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
       
-    - name: Executing tests
+    - name: Executing test
       run: ./gradlew test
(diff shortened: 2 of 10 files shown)


.github/workflows/run_tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 9
+    - name: Set up JDK 8
```
scorbett123, line 15: Sorry, I obviously mistyped.

scorbett123 (approved):

ZacSharp (approved):
Not sure about the renamed tasks but definitely a good thing in general.

FloGo3308:
> Not sure about the renamed tasks but definitely a good thing in general.

What would you prefer then? I changed the build action to match the style of the test action becouse the build action still contained some stuff from the template.

Some questions:
- should I change the file name of the build action too?
- should I change the name of the build step to `Executing build` like the `Executing tests` the test action uses?(`Executing build` doesn't sound right though)
- why is the name(line 2) of the test action called `Tests` but the job(line 9) called `test`(without a s at the end)?(just a minor detail)

ZacSharp:
>     * should I change the file name of the build action too?

I think if the name doesn't mention gradle, then the name should not do it as well.
>     * should I change the name of the build step to `Executing build` like the `Executing tests` the test action uses?(`Executing build` doesn't sound right though)

That's sounds wrong as you said. I think this one is the only name using a progressive form.
>     * why is the name(line 2) of the test action called `Tests` but the job(line 9) called `test`(without a s at the end)?(just a minor detail)

Don't know, if you think it should be changed this is the pr for it.

As I said I'm not sure about the renames, which in this case means me having doubts about pretty much every line of this answer.
I'll send it anyway because I don't want to leave you without any answer at all.

.github/workflows/workflow_test.yml
```diff
@@ -12,15 +11,14 @@ jobs:
...
       with:
         java-version: '8'
         distribution: 'adopt'
-        
+
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
       
-    - name: Executing tests
+    - name: Executing test
```
scorbett123, line 23: Change these both to the plural tests.

scorbett123 (changes requested):

.github/workflows/workflow_test.yml
FloGo3308: wdym with plural?

scorbett123: tests not test

PR #2814, opened 2021-06-16
cleaner fabric/forge build

wagyourtail:
use `forgified-fabric-loom` (`architectury-loom` in 1.17) as a method for not having as much of a garbage if/else'd build.gradle
note the new `-Ploom.forge=true` on all non-fabric builds, (`loom.platform=forge` on 1.17)

I would get it working with 1.15, but arch-loom has issues with it...

this also will allow using the loom migrateMappings task to help switch to mojmap in1.17 https://docs.architectury.dev/docs/forge_loom/using_mcp.html


wagyourtail:
this may have literally just broken because of one of mercury's dependencies updating to java 11... I'll fix soon

scripts/proguard.pro
```diff
@@ -3,6 +3,7 @@
 
 -optimizationpasses 5
 -verbose
+-ignorewarnings #until https://github.com/wvengen/proguard-maven-plugin/pull/76
```
leijurv, line 6: This makes me nervous, because a lot of the effort in setting up the proguard plugin correctly is making it so that it doesn't throw warnings on library versions. That link seems to be about Java 9... are we making java 9 binaries now? If so, why?

build.gradle
```diff
@@ -37,44 +35,30 @@ buildscript {
         maven {
             url = 'https://libraries.minecraft.net/'
         }
-        mavenCentral()
-        jcenter()
     }
-
     dependencies {
-        classpath group: 'com.github.ImpactDevelopment', name: 'ForgeGradle', version: '3.0.115'
```
leijurv, line 45: Is this added back somewhere else? If not, how does building for vanilla work?

scripts/proguard.pro
wagyourtail: this is because some of the required libraries have `meta-inf/versions/9` folders that proguard tries to use, finds a class path conflict and warns/crashes on

build.gradle
wagyourtail: so basically, it builds vanilla by building forge but remapping to notch instead of searge. Also no ForgeGradle dep at all even for forge (that's what arch-loom is for)

scripts/proguard.pro
wagyourtail: ok so this is fixed with java 16 when on my 1.17 fork using modifications to these build scripts

wagyourtail: figured out that one of the other things fixed it somehow. not sure how but it doesn't throw errors now...

build.gradle
```diff
@@ -268,71 +188,21 @@ jar {
...
-} else {
-    task copyMcJar(type: Copy) {
-        def mcJar = { getClientJar() }
-
-        from mcJar
-        into 'build/createMcSrgJar/'
-        rename { 'client-srg.jar' }
+if (compileType.equals("OFFICIAL")) {
+    remapJar {
+        toM.set "official"
```
wagyourtail, line 193: that's what this is for leij

PR #2825, opened 2021-06-21
ParkourPlace for shorter parkour jumps

### LINKED ISSUE #2350: ParkourPlace for shorter parkour jumps
## Describe your suggestion
Baritone should be able to place the destination block for any parkour jump, instead of only for 3 block jumps.

## Settings
Obviously this would be affected by `allowPlace`, `allowParkourPlace` and `allowParkour`

## Context
When jumping against a wall baritone can't jump a little further instead of placing a block to land on, so if it needs to jump against a wall that is 1 or 2 blocks away it has to build a bridge instead.
In case MovementParkour will ever use the multi-destination movements designed for #2142 this will also help when jump-placing along a wall and it would be faster to only jump 2 blocks before getting off the wall.

## Final checklist
- [x] I have not used any additional OwO's or UwU's in this issue.

Zephreo:
Allows parkour place to consider 2 and 1 block jumps
Fixes #2350

### DIFF
--- a/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
@@ -104,17 +104,23 @@ public class MovementParkour extends Movement {
                 maxJump = 3;
             }
         }
+        
+        // check parkour jumps from smallest to largest for obstacles/walls and landing positions
+        int verifiedMaxJump = -1;
         for (int i = 2; i <= maxJump; i++) {
+        	verifiedMaxJump = i - 1;
             int destX = x + xDiff * i;
             int destZ = z + zDiff * i;
+            
+            // check head/feet
             if (!MovementHelper.fullyPassable(context, destX, y + 1, destZ)) {
-                maxJump = i - 1;
                 break;
             }
             if (!MovementHelper.fullyPassable(context, destX, y + 2, destZ)) {
-                maxJump = i - 1;
                 break;
             }
+            
+            // check for ascend landing position
             IBlockState destInto = context.bsi.get0(destX, y, destZ);
             if (!MovementHelper.fullyPassable(context.bsi.access, context.bsi.isPassableBlockPos.setPos(destX, y, destZ), destInto)) {
                 if (i <= 3 && context.allowParkourAscend && context.canSprint && MovementHelper.canWalkOn(context.bsi, destX, y, destZ, destInto) && checkOvershootSafety(context.bsi, destX + xDiff, y + 1, destZ + zDiff)) {
@@ -124,9 +130,10 @@ public class MovementParkour extends Movement {
                     res.cost = i * SPRINT_ONE_BLOCK_COST + context.jumpPenalty;
                     return;
                 }
-                maxJump = i - 1;
                 break;
             }
+            
+            // check for flat landing position
             IBlockState landingOn = context.bsi.get0(destX, y - 1, destZ);
             // farmland needs to be canWalkOn otherwise farm can never work at all, but we want to specifically disallow ending a jump on farmland haha
             if (landingOn.getBlock() != Blocks.FARMLAND && MovementHelper.canWalkOn(context.bsi, destX, y - 1, destZ, landingOn)) {
@@ -137,20 +144,23 @@ public class MovementParkour extends Movement {
                     res.cost = costFromJumpDistance(i) + context.jumpPenalty;
                     return;
                 }
-                maxJump = i - 1;
                 break;
             }
+            
             if (!MovementHelper.fullyPassable(context, destX, y + 3, destZ)) {
-                maxJump = i - 1;
                 break;
             }
+            
+            // reset verifiedMaxJump so that if we exit out of the for loop here it is not changed
+            verifiedMaxJump = maxJump;
         }
+        
         // parkour place starts here
         if (!context.allowParkourPlace) {
             return;
         }
-        // check parkour jumps from largest to smallest
-        for (int i = maxJump; i > 1; i--) {
+        // check parkour jumps from largest to smallest for positions to place blocks
+        for (int i = verifiedMaxJump; i > 1; i--) {
             int destX = x + i * xDiff;
             int destZ = z + i * zDiff;
             IBlockState toReplace = context.get(destX, y - 1, destZ);
--- a/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
@@ -106,9 +106,8 @@ public class MovementParkour extends Movement {
         }
         
         // check parkour jumps from smallest to largest for obstacles/walls and landing positions
-        int verifiedMaxJump = -1;
+        int verifiedMaxJump = 1; // i - 1 (when i = 2)
         for (int i = 2; i <= maxJump; i++) {
-        	verifiedMaxJump = i - 1;
             int destX = x + xDiff * i;
             int destZ = z + zDiff * i;
             
@@ -151,8 +150,7 @@ public class MovementParkour extends Movement {
                 break;
             }
             
-            // reset verifiedMaxJump so that if we exit out of the for loop here it is not changed
-            verifiedMaxJump = maxJump;
+            verifiedMaxJump = i;
         }
         
         // parkour place starts here
--- a/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementParkour.java
@@ -108,10 +108,12 @@ public class MovementParkour extends Movement {
             int destX = x + xDiff * i;
             int destZ = z + zDiff * i;
             if (!MovementHelper.fullyPassable(context, destX, y + 1, destZ)) {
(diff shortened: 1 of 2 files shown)


ZacSharp:
You have tested this and it worked?
Because I can't get it to do a 2 block parkour place (even tried fixing parkour cost to 1 tick).

Zephreo:
Maybe it's cheaper to do other actions in your tests? 
image

ZacSharp (approved):
I had the thought about the movement cost as well and that's why I tried making any parkour move cost 1 tick, but apparently I just didn't try enough.<details><summary>I now got it do a few</summary></details><details><summary>it seems to not like them</summary></details><details><summary>and often did stuff like this</summary></details>Anyway, it works and the looks good to me

Zephreo:
The reason why it prefers 3 block parkour places is since the current MovementParkour only returns a single result to the pathfinder (from largest to smallest parkour places if valid) I didn’t modify this since I didn’t notice it having a significant effect and it would make a (although approved by leijuv) core change to the pathfinder.

Zephreo:
Another reason 3 block parkour jumps are preferred is due to the cost calculation of the parkour system favouring sprinting (as it’s faster) over walking. This is another issue that could be resolved as all parkour jumps take a fixed time based on height (12 ticks for flat jumps) not based on distance.

Zephreo:
Ok I found a problem that caused walls that collided with 3 block jumps effecting the smaller jumps, I’ll test my fix in a few hours when I’m at a computer.

Zephreo:
Just tested and it works great. It should be able to make jumps that go towards walls now instead of just making the ones that go beside the wall. 

<details>
  <summary>(Like the last jump in this image)</summary>

image

</details>

ZacSharp:
Now it behaves like I expected<details><summary>like this</summary>

</details><details><summary>there's just a slight problem with the path crossing itself</summary>




</details>I suspect this happens because it plans to place a block below the intersection but once it's done with the jump it notices that the block is already there and fails.<br>However, I was able to reproduce this on master as well, so it's an older bug and not introduced by/to be fixed in this pr.

ZacSharp (approved):

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -120,57 +122,62 @@ public static void cost(CalculationContext context, int x, int y, int z, EnumFac
...
+            	continue;
+            }
+            if (!checkOvershootSafety(context.bsi, destX + xDiff, y, destZ + zDiff)) {
+            	continue;
+            }
+            for (int j = 0; j < 5; j++) {
+                int againstX = destX + HORIZONTALS_BUT_ALSO_DOWN_____SO_EVERY_DIRECTION_EXCEPT_UP[j].getXOffset();
+                int againstY = y - 1 + HORIZONTALS_BUT_ALSO_DOWN_____SO_EVERY_DIRECTION_EXCEPT_UP[j].getYOffset();
+                int againstZ = destZ + HORIZONTALS_BUT_ALSO_DOWN_____SO_EVERY_DIRECTION_EXCEPT_UP[j].getZOffset();
+                if (againstX == x + xDiff * 3 && againstZ == z + zDiff * 3) { // we can't turn around that fast
```
leijurv, line 171: I believe this `3` might need to be changed. The purpose of this `if` is to prevent Baritone from attempting to spin around in midair and place the parkour-place block behind it, which flags anticheat. (or, at least, it flagged NCP in late 2018 when I tested this).

leijurv:
> the current MovementParkour only returns a single result to the pathfinder (from largest to smallest parkour places if valid) I didn’t modify this since I didn’t notice it having a significant effect and it would make a (although approved by leijuv) core change to the pathfinder.

This was previously semi-approved by me, yes. I'm still open to the idea it's just a little bit worrisome because of how much of a complication it would be to the pathfinder, and the path stitcher. I imagine it might be an ongoing headache. 

PR generally looks good. Some formatting issues, looks like the indent is strange? Maybe a mixture of tabs vs spaces?

Screen Shot 2021-06-30 at 12 36 12 PM
^ looks like the indent for the `if` is smaller than the indent for the `for`?

leijurv:
Cool. Ready to merge except for one last question:

I don't quite like how `maxJump = i - 1; break;` is repeated each time.

Could we achieve the same effect by doing something like this? (just off the top of my head, please check if this actually works lol)

```
        int verifiedMaxJump = 2;
        for (int i = 2; i <= maxJump; i++) {
            ...
            if (something) {
                // removed all the updates to maxJump
                break;
            }
            ....
            verifiedMaxJump = i; // so it only sets verifiedMaxJump if the loop gets all the way to the bottom with no problems / breaks / returns
        }
        ...
        for (int i = verifiedMaxJump; i > 1; i--) { // start at verifiedMaxJump instead of maxJump
```

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -104,17 +104,23 @@ public static void cost(CalculationContext context, int x, int y, int z, EnumFac
                 maxJump = 3;
             }
         }
+        
+        // check parkour jumps from smallest to largest for obstacles/walls and landing positions
+        int verifiedMaxJump = -1;
```
leijurv, line 109: If you change this to 1

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -104,17 +104,23 @@ public static void cost(CalculationContext context, int x, int y, int z, EnumFac
...
         for (int i = 2; i <= maxJump; i++) {
+        	verifiedMaxJump = i - 1;
```
leijurv, line 111: Then you can remove this line, no?

Zephreo: no this sets verifiedMaxJump so that if it breaks it keeps the value

Zephreo: (just realised the 2 comments are connected lol)

Zephreo: To emulate what was there before verifiedMaxJump is set to `i-1` before the `break;` occurs, this way if it `break;`'s then it would keep the verifiedMaxJump value of `i-1` otherwise if it does not break then it exits out of the loop with `verifiedMaxJump = maxJump` effectively doing nothing.

Zephreo:
Not sure why the build task is failing... I can run `gradlew run` just fine but `gradlew build` fails. Any Ideas?

Zephreo:
sorry bad practice to revert like this (reverted default verifiedMaxJump to 1)

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
leijurv: Right, I'm imagining that if you change the above line to `= 1`, and change the `verifiedMaxJump = maxJump` to `verifiedMaxJump = i` at the bottom of the loop, then by the time it hits a `break` verifiedMaxJump will already equal i-1. It will have been set by the previous time through the loop having run `verifiedMaxJump = i`, or if it's the first time through, by `verifiedMaxJump = 1`.

Zephreo: this wouldn't work because there are cases where maxJump does not change at all exiting the loop while still having verifiedMaxJump the same as maxJump.

leijurv: And that is the same as the previous behavior seen here https://github.com/cabaletta/baritone/blob/2159773723c7b392a8ec2c67c57cca50e4ffb3d9/src/main/java/baritone/pathing/movement/movements/MovementParkour.java If the loop on 107 exits normally (no break), then the loop on 153 starts at normal maxJump. That would be the same, because the first loop goes up to maxJump, so the last thing it would do is set `verifiedMaxJump = i` but `i` equals `maxJump`.

leijurv: I applied what I think to that link https://github.com/cabaletta/baritone/blob/2159773723c7b392a8ec2c67c57cca50e4ffb3d9/src/main/java/baritone/pathing/movement/movements/MovementParkour.java

Look at: https://pastebin.com/KfiTNkrq

Zephreo: The end `verifiedMaxJump = maxJump` could be encased in an if statement that checks if it is the last for loop iteration since that is the only time it is used

```
if (lastLoop) {
    verifiedMaxJump = maxJump;
} else {
    verifiedMaxJump = i;
}
```
Something like this? 

I see now... very smart.

`lastLoop = i == maxJump`

ok should work

graemeg:
> Not sure why the build task is failing... I can run `gradlew run` just fine but `gradlew build` fails.
> edit: seems that the master branch is failing as well

Not to highjack this thread, but myself and a few others have reported the same problem is Discord. For some reason, a few days ago the `gradlew build` command fails on the `proguard` step.

Thanks for the tip though, I didn't realise the `gradlew run` still works. At least that allows me to continue my development and testing.

PR #2833, opened 2021-06-24
✨ Add layerHeight setting

### LINKED ISSUE #1756: Modify height for layers
Make the height of the Layer modifiable

Would be really useful for excavating stuff much faster and without the bot going straight down nor being slowed down/limited by only being able to mine 1 block thick layer

ZacSharp:
Because it hurts watching it clear 1 block layers of a 257x257x257 sphere.

There's a rough alternative implementation at https://github.com/ZacSharp/baritone/tree/layerHeight2, which has the benefit of not changing the layer associated with a certain height in blocks, making it more usable when changing `layerHeight` halfway through a build.

closes #1756

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -851,6 +851,11 @@ public final class Settings {
     public final Setting<Boolean> layerOrder = new Setting<>(false);
 
     /**
+     * How high should the individual layers be?
+     */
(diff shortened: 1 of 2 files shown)


leijurv:
Oh, this is interesting!

leijurv:
Need to think about it more but this looks **quite** fun

leijurv:
Might as well at this point.

PR #2859, opened 2021-07-10
Revamped readme(fixed)

MoKetab:
Fixed link issues from prior pull req.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,93 +1,71 @@
-# Baritone
+***
+
+<center> <h1>Baritone</h1> </center>
+
+***
+
 <p align="center">
   <a href="https://github.com/cabaletta/baritone/releases/"><img src="https://img.shields.io/github/downloads/cabaletta/baritone/total.svg" alt="GitHub All Releases"/></a>
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


wagyourtail:
you know you could've just amended the previous pr right?
too late now but in the future...

Flurrrr:
only 2 badges?!?!

ZacSharp (changes requested):
Not going to comment on lines, it just looks like a mess.

I don't care about the two big fun badges, but the rest is useful information. The mc version badges used to link to the appropriate release, which was a good thing IMHO.

You cleaned up the mess of a text hinting at various versions and tutorials and whatever, which is good, but by doing so you have left out the information about Baritones origins and speed as well as leijurvs bitcoin address (which definitely should be a little more emphasized).

Also I think we should not prefer just a few utility clients by mentioning them twice, just keep them with a badge for every client that wants it.

And IMO you used way too many separator lines, one below each section or none at all are enough.

You also removed a lot of things leijurv might care about, such as the YourKit section, Impact being overly advertised, the bragging about having committed daily for a year and possibly the stargazer graph as well.

PR #2867, opened 2021-07-14
Faster movement

ghostmantisboy:
Allow baritone to sprint jump while on flat land, it only tries to when it has 3 meat shanks. This is to get around faster, mostly and also because I wanna see the line have strange spikes in it.


scorbett123:
Create an issue not a pr. Why do people keep doing this.

PR #2879, opened 2021-07-18
Fix proguard from being broken.

scorbett123:
Currently every build is failing as soon as it is run.

This seems to be to do with a update in one of the dependencies changing the version of patchy, which wasn't later changed in proguard.pro leading to a null pointer exception.


wagyourtail (approved):
LGTM

Issue #2942, opened 2021-08-18
Tools for specific block types


CriminosaMC:
## Describe your suggestion
The feature should allow you to specify a slot for a tool for a specific block type - for example, redstone ore would use slot 3 which would have a silk touch pickaxe there.

## Settings
A setting to change this per block

## Context
I'd like this to be added because at the moment I have to manually babysit Baritone and place the correct tool (in this case, a silk touch pickaxe) whenever Baritone finds redstone ore.

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

RevanMardagh:
has this been added yet?

wagyourtail:
well, based on the fact that the issue is still open. no

Arsenaloro:
Make it faster!

PR #2702, opened 2021-04-28
farm: Fortune support

JamesTheAwesomeDude:
I'm trying to add [Fortune](https://minecraft.gamepedia.com/Fortune) support to `.b farm`.

Since breaking 0-hardness crops (i.e. Wheat, Carrots, Potatoes, and Nether Wart) does not use a tool's durability, should I selectively ignore the `cost` ca. the `else if` statement here in order to prefer (e.g.) a Golden Hoe with Fortune 3 over a Netherite Pickaxe for mining those blocks in particular?

Or should this logic be injected directly into [`FarmProcess.java#PathingCommand`](https://github.com/cabaletta/baritone/blob/v1.6.3/src/main/java/baritone/process/FarmProcess.java#L236-L239), instead?

PR #2948, opened 2021-08-19
Auto tool priority system

PanForPancakes:
list of blocks that baritone will mine with silk touch
https://github.com/cabaletta/baritone/issues/2942

### DIFF
--- a/src/api/java/baritone/api/utils/ToolPriorityType.java
+++ b/src/api/java/baritone/api/utils/ToolPriorityType.java
@@ -44,6 +44,6 @@ public enum ToolPriorityType {
                 return tpt;
             }
         }
-        return DEFAULT;
+        throw new IllegalArgumentException(String.format("Unknown ToolPriorityType %s", text));
     }
 }
--- a/src/api/java/baritone/api/utils/ToolPriorityType.java
+++ b/src/api/java/baritone/api/utils/ToolPriorityType.java
@@ -1,0 +1,49 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.api.utils;
+
+public enum ToolPriorityType {
+    DEFAULT("default"),
+    CAN_BREAK_BLOCK("can_break_block"),
+    BEST_SILKTOUCH_FOR_BLOCK("best_silktouch_for_block"),
+    FASTEST("fastest"),
+    LEAST_DURABLE("least_durable"),
+    EXPENSIVE("expensive"),
+    CHEAP("cheap"),
+    ENCHANTED("enchanted"),
+    NOT_ENCHANTED("not_enchanted");
+
+    private final String name;
+
+    ToolPriorityType(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public static ToolPriorityType fromString(String text) {
+        for (ToolPriorityType tpt : ToolPriorityType.values()) {
+            if (tpt.name.equalsIgnoreCase(text)) {
+                return tpt;
+            }
+        }
+        return DEFAULT;
+    }
+}
--- a/src/main/java/baritone/utils/ToolSet.java
+++ b/src/main/java/baritone/utils/ToolSet.java
@@ -143,7 +143,7 @@ public class ToolSet {
         IBlockState blockState = b.getDefaultState();
 
         int bestItem = player.inventory.currentItem;
-        for (ToolPriorityType tpt : Baritone.settings().toolPriority.value) {
+        for (ToolPriorityType tpt : Baritone.settings().autoToolPriority.value) {
             switch (tpt) {
                 case DEFAULT: {
                     double highestSpeed = Double.NEGATIVE_INFINITY;
--- a/src/main/java/baritone/utils/ToolSet.java
+++ b/src/main/java/baritone/utils/ToolSet.java
@@ -138,8 +138,12 @@ public class ToolSet {
             double speed = calculateSpeedVsBlock(itemStack, blockState);
             boolean silkTouch = hasSilkTouch(itemStack);
 
-            if(doesBlockRequireSilkTouch && silkTouch) {
-                speed *= 1.2;
+            if(silkTouch) {
+                if(doesBlockRequireSilkTouch) {
+                    speed *= 1.2;
+                } else {
+                    speed *= 0.8;
+                }
             }
 
             if (speed > highestSpeed) {
--- a/src/main/java/baritone/utils/ToolSet.java
+++ b/src/main/java/baritone/utils/ToolSet.java
@@ -138,9 +138,8 @@ public class ToolSet {
             double speed = calculateSpeedVsBlock(itemStack, blockState);
             boolean silkTouch = hasSilkTouch(itemStack);
 
-            if(doesBlockRequireSilkTouch && !silkTouch)
-            {
-                continue;
+            if(doesBlockRequireSilkTouch && silkTouch) {
+                speed *= 1.2;
             }
 
             if (speed > highestSpeed) {
--- a/src/main/java/baritone/utils/ToolSet.java
+++ b/src/main/java/baritone/utils/ToolSet.java
@@ -123,6 +123,8 @@ public class ToolSet {
         int lowestCost = Integer.MIN_VALUE;
         boolean bestSilkTouch = false;
         IBlockState blockState = b.getDefaultState();
+        boolean doesBlockRequireSilkTouch = Baritone.settings().preferSilkTouchForBlocks.value.contains(b);
+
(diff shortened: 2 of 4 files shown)


src/main/java/baritone/utils/ToolSet.java
```diff
@@ -132,8 +134,15 @@ public int getBestSlot(Block b, boolean preferSilkTouch, boolean pathingCalculat
...
                 continue;
             }
+
             double speed = calculateSpeedVsBlock(itemStack, blockState);
             boolean silkTouch = hasSilkTouch(itemStack);
+
+            if(doesBlockRequireSilkTouch && !silkTouch)
+            {
+                continue;
+            }
```
scorbett123, line 144: Why don't make it so that there is a higher speed if it does have silk touch, otherwise it won't be able to mine the blocks if you don't have a item with silk touch on it in your inventory.
```suggestion
            if(doesBlockRequireSilkTouch && silkTouch) {
                speed *= 1.2;
            }
```

Also formatting

scorbett123 (commented):
Thanks, that's better, I am still wondering though whether this is the best method of implementing this, it probably is but if more enchantments are added it would become really cluttered. @ZacSharp any thoughts?

PanForPancakes:
> Thanks, that's better, I am still wondering though whether this is the best method of implementing this, it probably is but if more enchantments are added it would become really cluttered. @ZacSharp any thoughts?

https://github.com/cabaletta/baritone/pull/2702
this draft pull request is probably that one "enchantment"

ZacSharp (changes requested):
As long as it is configurable and *does not leave `getBestSlot`* (i.e. the true cost is returned) I actually like the idea of treating tools as if they were slightly better/worse, because it allows to prefer enchantments over speed within some limits and keeps the different preferences in separate if-statements. However it does not allow preferring while still guaranteeing to choose the fastest tool and misusing the cost like this where we could just as well switch to a proper selection strategy feels wrong.
I think we should instead by some means generate a tuple of preference values and use that as the sort key.

Here are some examples:
sort key | meaning
---------|-------
[speed,silktouch,material,slot] | always use the fastest tool, if there are multiple prefer those with silktouch and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.<br>This is the current strategy>
[speed+silktouch,silktouch,material,slot] | Use the fastest tool unless there is one with silktouch that is not too much worse, if there are multiple prefer those with silktouch and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.<br>This is the strategy with this pr>
[slot] | use the tool with the lowest slot number
[silktouch,speed,material,slot] | always silktouch tools if you can, if there are multiple prefer faster tools and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.
[-material,enchanted,slot] | Good for bragging. Take the most expensive tools you have, if there are multiple prefer those with any enchantment and if you still have a choice just take the one with the lowest slot number.

We could then add api functionality to extend this tuple with more values or add functionality to register new value generators (e.g. `Function<ItemStack,Tuple<Number,Number>>`) the user can then arrange using a setting (slot will always be the implicit last entry).

This "rewiew" starts getting big, so I'll just stop here, keep the rest of my thought until someone asks for them and hope what I've written so far makes sense to you.

PanForPancakes:
toolPriority system:

`#set toolPriority can_break_block,fastest,least_durable`
Baritone will first get every valid instrument in hotbar.
Removes instruments that can't break block. (iron pick if mining obsidian as example)
Removes slowest instruments and only keep fast ones.
Removes most durable instruments.
And at end it will select fastest and closest to 0 slot instrument

`#set toolPriority can_break_block,best_silktouch_for_block,cheap,least_durable` and `#set preferSilkTouchForBlocks diamond_ore,lapis_ore` can be used to mine with "bare minimum" tools, stone pickaxes to mine stone/iron, silk touch iron pick to mine diamonds/lapis (tested, it works)

src/api/java/baritone/api/utils/ToolPriorityType.java
```diff
@@ -0,0 +1,49 @@
...
+        return name;
+    }
+
+    public static ToolPriorityType fromString(String text) {
+        for (ToolPriorityType tpt : ToolPriorityType.values()) {
+            if (tpt.name.equalsIgnoreCase(text)) {
+                return tpt;
+            }
+        }
+        return DEFAULT;
```
ZacSharp, line 47: If the user asks for a type nobody added that's an error, just like block arguments/settings don't fall back to air when they can't find a block with the provided name.

ZacSharp (changes requested):
Sorry, my description was intended to explain the effects and not to be an implementation guide (I didn't even expect it to just be implemented without at least some sort of discussion).
To achieve what I suggested you can still just iterate over the item once, but instead of looking at the cost to judge whether the current item is better than the current best you compare the values returned by the first priority function and if those are equal you continue with the next priority and if those are equal you compare the next one and so forth, like the old implementation uses silktouch as the secondary measure.
That way you can keep all preference implementations out of `getBestSlot` and implement them somewhere else (probably still `ToolSet` or a new class) and register them via the api.

Also I would prefix all of the new setting names with "autoTool" so the show up right next to the `autoTool` setting in `#set` and `#mod` and to make up for the face that we don't have setting categories.

And @scorbett123 could you please review my suggestions before we go for them?

src/api/java/baritone/api/utils/ToolPriorityType.java
```diff
@@ -0,0 +1,49 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.api.utils;
+
+public enum ToolPriorityType {
```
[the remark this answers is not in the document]
PanForPancakes: No idea how to implement that, can anyone help with it?

scorbett123:
> I think we should instead by some means generate a tuple of preference values and use that as the sort key (like python sorts tuples, basically equivalent to a series of stable sorts using the n-th entry as the key, starting with the last one).
> 
> Here are some examples:
> 
> sort key	meaning
> [speed,silktouch,material,slot]	always use the fastest tool, if there are multiple prefer those with silktouch and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.
> This is the current strategy>
> [speed+silktouch,silktouch,material,slot]	Use the fastest tool unless there is one with silktouch that is not too much worse, if there are multiple prefer those with silktouch and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.
> This is the strategy with this pr>
> [slot]	use the tool with the lowest slot number
> [silktouch,speed,material,slot]	always silktouch tools if you can, if there are multiple prefer faster tools and out of those take the cheapest material. If you still have a choice take the one with the lowest slot number.
> [-material,enchanted,slot]	Good for bragging. Take the most expensive tools you have, if there are multiple prefer those with any enchantment and if you still have a choice just take the one with the lowest slot number.
> We could then add api functionality to extend this tuple with more values or add functionality to register new value generators (e.g. `Function<ItemStack,Tuple<Number,Number>>`) the user can then arrange using a setting (slot will always be the implicit last entry because we iterate in that order).

Yea, this seems cool, however it will probably need a custom command to change it, however the ability to apply a specific weighting would also be cool, for example, it will prefer to use tools with a specific enchantment, however if they are really slow then it will instead use a quicker tool.
For example:

| Sort Key  | Meaning |
| ------------- | ------------- |
|  [silkTouch: 1, speed: 0.25] | Prefer to use silk touch, however if using another tool is 5 times faster, then use it  |


@ZacSharp Do you have a discord so that you can join the baritone discord and we can discuss this in more detail?

scorbett123:
For now this need significant work on how we want the new tool priority system to work so I am going to close this pull request.

If how we decide to make this work fits with this pr then it can be reopened and changed.

We need to discuss this in much more detail before it is implemented properly.

scorbett123:
Let just discuss this here for now, if you have any suggestions about where to discuss this, then I am happy to use it.

PanForPancakes:
> [silkTouch: 1, speed: 0.25]
Parsing this is totally possible, however how it's gonna be implemented?
Right now I don't see any way how it can be better than ZacSharp suggested method and it looks like it's gonna be very complicated for average user. Whoever sees this in settings will be at least very confused.

PanForPancakes:
Also with this

> [silkTouch: 1, speed: 0.25]

original idea with having setting for using silk touch tool for some blocks is starting to get lost or will be bad at the end.

scorbett123:
> Also with this
> 
> > [silkTouch: 1, speed: 0.25]
> 
> original idea with having setting for using silk touch tool for some blocks is starting to get lost or will be bad at the end.

sorry yea, silk touch for some blocks has got lost, it will be implemented in the new version though. We will try and keep it easy and not "bad" but we will be sure to have it implemented in a hopefully easier to use, and overall better method than the current tool set allows you to implement.

ZacSharp:
I don't know whether we can achieve that "prefer silktouch but only up to 50% slower" behaviour and similar in a generic way with when doing so, but a simple option would be to combine multiple ratings using weighted addition and then using the result as one of the list entries.

What about irc (Libera.Chat) for discussing this further?

scorbett123:
ok, I have downloaded a client for it, how would I get to a point of being able to message you?

ZacSharp:
So far (for one day) I'm using their webchat, so I haven't actually followed those steps myself, but according to [this guide](https://libera.chat/guides/connect) you just conect your client to `irc.libera.chat` on port `6697`.
Once you are connected I have created `##baritone` as a public place to discuss this or, if you want a private conversation, you can message me (ZacSharp there as well) directly.

scorbett123:
I should have joined that now and sent a message, the public channel should be fine. Have you received my message?

ZacSharp:
@scorbett123 I now have a few ideas on how to do the more advanced sorting, so can we maybe meet again?

scorbett123:
> @scorbett123 I now have a few ideas on how to do the more advanced sorting, so can we maybe meet again?

Sorry really busy at the moment, maybe in a few days

Issue #324, opened 2019-02-01
Keep a radius of cached regions loaded


leijurv:
Probably just like 1 in every direction (so 9 loaded).

Currently they're only loaded on the chunk packer thread when a chunk is loaded that is in an unloaded cache region.

This kind of defeats the purpose of the cached renderer if it only renders loaded regions, and regions are only loaded if the server tells us about a chunk that's inside them. 

Not really an issue unless you're standing near a border tho (coords divisible by 512), but still

leijurv:
Just to be client-freecam-method-agnostic there should be something that asynchronously loads this radius around the mc.player and the mc.renderviewentity both once a tick and once a frame. (if it's already loaded it's just 9 hashmap lookups in a new thread, no big deal).

Leisterbecker:
Can I work on that issue?

ZacSharp:
Someone already did it in #2967

Leisterbecker:
Ah okay, thanks for clarification :)

PR #2967, opened 2021-08-29
Keep a radius of cached regions loaded

UPSman123:
This implements an enhancement from #324. I created the ChunkLoader
class. This handles loading chunks around the player and render view
entity in a given radius. The chunks are loaded from disk using the
CachedWorld.tryLoadFromDisk function. If a chunk has not yet been
added to the cache (in memory or on disk) then this will not cache it.

Curently the class is initialized from the Baritone constructor. I don't know if this is okay or if it should be done elsewhere.
I couldn't find a good place for it because it needs to be registered with the GameEventHandler.

I also noticed that the ICachedWorld interface does not include the tryLoadFromDisk function even though it is public.
I don't know if this is intentional or not. In this commit I decided to just cast the ICachedWorld to a CachedWorld.

### DIFF
--- a/src/main/java/baritone/Baritone.java
+++ b/src/main/java/baritone/Baritone.java
@@ -24,6 +24,7 @@ import baritone.api.event.listener.IEventBus;
 import baritone.api.utils.Helper;
 import baritone.api.utils.IPlayerContext;
 import baritone.behavior.*;
+import baritone.cache.ChunkLoader;
 import baritone.cache.WorldProvider;
 import baritone.command.manager.CommandManager;
 import baritone.event.GameEventHandler;
@@ -120,6 +121,9 @@ public class Baritone implements IBaritone {
         this.worldProvider = new WorldProvider();
         this.selectionManager = new SelectionManager(this);
         this.commandManager = new CommandManager(this);
+
+        ChunkLoader chunkLoader = new ChunkLoader(this, 2); // load 3 by 3 area of chunks around the player.
+        this.gameEventHandler.registerEventListener(chunkLoader);
     }
 
     @Override
--- a/src/main/java/baritone/cache/ChunkLoader.java
+++ b/src/main/java/baritone/cache/ChunkLoader.java
@@ -1,0 +1,85 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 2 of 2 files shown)


src/main/java/baritone/cache/ChunkLoader.java
```diff
@@ -0,0 +1,85 @@
...
+import baritone.api.event.listener.AbstractGameEventListener;
+import net.minecraft.client.entity.EntityPlayerSP;
+import net.minecraft.entity.Entity;
+import net.minecraft.world.World;
+
+import static baritone.api.utils.Helper.mc;
+
+/**
+ * Loads an area around the player and renderViewEntity to the cache.
+ * The
```
ZacSharp, line 33: Don't think you meant to leave that here.

src/main/java/baritone/cache/ChunkLoader.java
```diff
@@ -0,0 +1,85 @@
...
+    private void loadChunks() {
+        EntityPlayerSP player = mc.player;
+        Entity renderViewEntity = mc.getRenderViewEntity();
+        // Don't do anything before the player is loaded.
+        if (player == null || renderViewEntity == null) return;
+        WorldData worldData = this.worldProvider.getCurrentWorld();
+        // Don't do anything if the world doesn't exist. Like in the main menu.
+        if (worldData == null) return;
+        loadAroundChunk(player.chunkCoordX, player.chunkCoordZ);
+        // Only load around the renderViewEntity if it's in a different chunk.
```
ZacSharp, line 67: Maybe also don't do anything if the position didn't change since last tick?
Don't know whether the overhead of reloading every tick is more than the overhead of checking this.

src/main/java/baritone/Baritone.java
```diff
@@ -120,6 +121,9 @@
         this.worldProvider = new WorldProvider();
         this.selectionManager = new SelectionManager(this);
         this.commandManager = new CommandManager(this);
+
+        ChunkLoader chunkLoader = new ChunkLoader(this, 2); // load 3 by 3 area of chunks around the player.
```
ZacSharp, line 125: #324 suggested loading 3 by 3 *regions* around the player, not chunks.
Loading 3 by 3 chunks changes almost nothing, it now just loads the next region 16 blocks before you reach the region border instead of when you reach it.

src/main/java/baritone/cache/ChunkLoader.java
```diff
@@ -0,0 +1,85 @@
...
+        if (player.chunkCoordX != renderViewEntity.chunkCoordX || player.chunkCoordZ != renderViewEntity.chunkCoordZ) {
```
ZacSharp, line 68: You should compare region coords here because you can only load entire regions at once.

ZacSharp (changes requested):

UPSman123:
Thanks for the code review.
I was a bit confused about the difference between regions and chunks but I properly looked into it and I get it now. I changed it to work with regions instead.
I also added a check to see if the player moved to a different region. This won't prevent regions around the renderViewEntity from being loaded if it is in a different region but in that case it's only a few hashmap lookups and shouldn't influence performance significantly. I could add code to check for the renderViewEntity as well but I feel like it will just make things more messy without any real gain.

ZacSharp (changes requested):
Just two oversights and some formatting

UPSman123:
I fixed it. Thanks again for looking it over.

ZacSharp (approved):
Looks good to me.

PR #2991, opened 2021-09-09
Slight performance optimization in GoalAxis

StylexTV:
After creating this [pr](https://github.com/cabaletta/baritone/pull/2954), I decided to take a look at the other goals and didn't find much besides this performance tweak. The performance gain is pretty small and I didn't really know if I should even make this change, but decided to anyway.
Note: This change (probably) doesn't affect the result of the calculation, it just makes it faster.

Also, I didn't create an issue first, so I hope it's ok to just create this pr.

### DIFF
--- a/src/api/java/baritone/api/pathing/goals/GoalAxis.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalAxis.java
@@ -33,11 +33,10 @@ public class GoalAxis implements Goal {
         int x = Math.abs(x0);
         int z = Math.abs(z0);
 
-        int shrt = Math.min(x, z);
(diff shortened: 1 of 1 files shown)


ZacSharp (approved):
Does not seem to break anything and I can't see why it should behave differently.
As you said the performance gain is probably not a lot , but since the heuristic is the bottleneck of most a* implementations and is called a **lot** I'd take it anyway.

It's totally fine to just create a pull request without opening an issue first.

scorbett123:
In fact, it is preferable to just create a pr, it means that there are less issues cluttering my emails, and the issues tab.

PR #3015, opened 2021-10-02
Immersive portal fix 1.16

SteamDoge9000:
A request for a release to be made wtih this as a immersive portal fix of some kind

either that or add maybe some method in the next 1.16.5 release to where if it detects immersive portals that it will use this instead, and then else it uses the regular

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1218,7 +1218,7 @@ public final class Settings {
     // here be dragons
 
     Settings() {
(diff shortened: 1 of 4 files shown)


ZacSharp:
Did you want to open an issue? This is a pull request.

Also immersive portals compatibility was already worked on and "suggesting" it again is just annoying.

SteamDoge9000:
sorry, i just dont know what todo, as for I cant find a download to the fix

SteamDoge9000:
I checked the respitory that I tried to merge but was unable to find a download for me to plug into the game, some help would acutaly be nice

Again, sorry for any inconvience and/or anoyance that I might have caused

PR #3037, opened 2021-10-16
✨ Add "#set add", "#set remove" and "#set clear"

### LINKED ISSUE #2306: Command "#set add" to add an item to a list in settings
New commands : `#set add`, `#set remove`.

Example : 

```
#set remove acceptableThrowawayItems stone
```

```
#set add acceptableThrowawayItems andesite,diorite,granite
```

In [USAGE.md](https://github.com/cabaletta/baritone/blob/master/USAGE.md#commands), `acceptableThrowawayItems` is listed as a setting that's "*fun / interesting / important that you might want to look at changing in normal usage of Baritone*".

However, during "normal usage of Baritone", using `#acceptableThrowawayItems dirt,cobblestone,netherrack` is a bit clunky if you want to add or modify your list.

Even in abnormal usage while editing `settings.txt`, it looks like this : 

```properties
acceptableThrowawayItems Optional[ResourceKey[minecraft:item / minecraft:dirt]],Optional[ResourceKey[minecraft:item / minecraft:cobblestone]],Optional[ResourceKey[minecraft:item / minecraft:netherrack]]
```

<details>
<summary>Off-topic</summary>
I'd suggest using JSON or YAML to store settings
</details>

The problem is that there's no way to add/modify a list, only to override it. And copying old values has to be done manually, which is tedious if there's many, many items in that list. I think a command to add/remove items from a list in the settings would be very appropriate. This way, during normal usage of Baritone, one could add/remove an item from the list without having to copy everything.

## Final checklist

ZacSharp:
Behaviour for lists and simple maps is simply appending or removing the values, for lists in maps the new value is appended to or removed from the value list.
* `#set add <setting> <value>` adds a value to a list setting or merges it into a map setting
* `#set remove <setting> <value>` removes the value from the setting, doing the exact opposite of `#set add`
* `#set clear <setting>` removes all entries from list and map settings

This is a draft because there are a few things I'd like to get sorted first (ideas are welcome, that's why this pr exists already):
* [x] `#set remove` and `#set clear` now allow for any list or map or mapped list to be empty, leading to edge cases like having an empty list of throwaways or the builder having to substitute some block for nothing (not air, nothing)
  EDIT: `#set <list setting> ,` can also set any setting to an empty list
* [ ] Empty lists/maps serialize to an empty string, which is not a valid value
* [ ] The way the parsing is done is more of a hack to work around the setting parsers only being able to parse a complete value but not e.g. just a key for a map
* [ ] I'd like to have `#set clear <setting> <key>`, `#set add <setting> <key> <value>` (currently redundant with `#set add <setting> <value>`), `#set remove <setting> <key>` and `#set <setting> <key> <value>` for `Map<*,List<*>>` type settings
* [x] The tab completion could be better, but I think that's a problem for settings in general and not specific to this pr. It just continues what is there.
* [x] ~Map parsing only works when the keys and values have the same type (did it wrong when adding maps, want to fix it now) (EDIT: see #3474)~ Fixed in #4324

closes #2306

### DIFF
--- a/src/api/java/baritone/api/schematic/SubstituteSchematic.java
+++ b/src/api/java/baritone/api/schematic/SubstituteSchematic.java
@@ -52,6 +52,9 @@ public class SubstituteSchematic extends AbstractSchematic {
             return desired;
         }
         List<Block> substitutes = substitutions.get(desiredBlock);
+        if (substitutes.isEmpty()) {
+            return desired;
+        }
         if (substitutes.contains(current.getBlock()) && !(current.getBlock() instanceof BlockAir)) {// don't preserve air, it's almost always there and almost never wanted
             return withBlock(desired, current.getBlock());
         }


wagyourtail:
> #set remove and #set clear now allow for any list or map or mapped list to be empty, leading to edge cases like having an empty list of throwaways or the builder having to substitute some block for nothing (not air, nothing)
I'd say either add an optional argument to map/list settings where you can specify the values that can't actually be removed/cleared, or fix it on the other end when settings are parsed in the various places baritone uses them in code. (the first way might be messier but easier to implement)

ZacSharp:
Unless I overlooked something `buildSubstitutes` was the only one causing problems and they should all work with empty lists now.

PR #3041, opened 2021-10-17
Always use the renderViewEntity for relative coordinates

ZacSharp:
I also thought about doing it for a few other things that happen relative to the player like `#tunnel` or `#thisway` and decided against it, keeping an explicit distinction rather than "here it makes sense to do it relative to the camera but not there".


ZacSharp:
I made this to make the behaviour of relative coordinates consistent, but if you want I can revert it for things like `#goal`. Just tell me which you don't want to use the camera position.

I also thought about letting the user decide by adding another type of coordinate, but I'm not really a fan of it.

PR #3042, opened 2021-10-18
Waypoint improvements

### LINKED ISSUE #2952: Allow rename of automatically created waypoints (on death/bed)
## Describe your suggestion
I do like having the waypoints (on death and sleeping on a bed) created automatically. Still, I would also like to rename them to more meaningful names representing the coordinates they store. I use multiple beds since I would spawn in one and do whatever, then go to another location (say, another farm far away) and sleep in a bed located there. Having those named the same (as it's automatically generated today) makes it hard to know which is where. Also, auto-creation of said beds would create duplicate waypoints, just different timestamps (which could be disabled using the `waypointOnDeath` mentioned in #2778.

## Settings
NA

## Context
Allow for meaningful names in `#waypoints`

## Final checklist

ZacSharp:
* You can now omit any argument from `#wp save`. Tab completion for the new forms is not implemented and not completely possible to implement.
* The usage section in `#help wp` now shows both tags and names can be used 
* `#wp info` now has a button to get the command that could have created the waypoint (kind of closes #2952)
* repeatedly clicking on a bed now creates exactly one waypoint for the head part of the bed
* deleted waypoints are now cached so you can restore them using `#wp restore`. The syntax using timestamps is not listed in the usage section, just like for all other `#set` subcommands that can use timestamps.

### DIFF
--- a/src/main/java/baritone/command/defaults/WaypointsCommand.java
+++ b/src/main/java/baritone/command/defaults/WaypointsCommand.java
@@ -21,6 +21,7 @@ import baritone.Baritone;
 import baritone.api.IBaritone;
 import baritone.api.cache.IWaypoint;
 import baritone.api.cache.Waypoint;
+import baritone.api.cache.IWorldData;
 import baritone.api.command.Command;
 import baritone.api.command.argument.IArgConsumer;
 import baritone.api.command.datatypes.ForWaypoints;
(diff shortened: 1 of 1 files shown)


ZacSharp:
Using `#wp restore` you can transfer waypoints between worlds.
I will change this tomorrow.

PR #3067, opened 2021-11-03
fix Null params on getCollisionShape causing NPEs

### LINKED ISSUE #2148: Null getCollisionShape returned when standing on irregular shaped blocks
## Some information
Operating system: Windows 10
Java version: 1.8
Minecraft version: 1.16.3
Baritone version: 1.6.1
Forge mods (if used): https://www.curseforge.com/minecraft/modpacks/all-the-mods-6 + Baritone + ForgeHax

## Correlated between
- https://github.com/TeamPneumatic/pnc-repressurized/issues/649
- https://github.com/BluSunrize/ImmersiveEngineering/issues/4408

## Exception, error or logs
https://paste.ubuntu.com/p/KDgYj76K9B/ - crashes PneumaticCraft [here](https://github.com/TeamPneumatic/pnc-repressurized/blob/da1e1e6fb5b95117c4e382388d3b1c654a478b17/src/main/java/me/desht/pneumaticcraft/common/block/BlockPneumaticCraftCamo.java#L86)
https://pastebin.com/X5EpVU5f - crashes PneumaticCraft [here](https://github.com/TeamPneumatic/pnc-repressurized/blob/da1e1e6fb5b95117c4e382388d3b1c654a478b17/src/main/java/me/desht/pneumaticcraft/common/block/BlockPneumaticCraftCamo.java#L86)
https://pastebin.com/3Nmjbn5E - crashes Immersive Engineering [here](https://github.com/BluSunrize/ImmersiveEngineering/blob/549dfa3182fa5ef0842c51f438a4a43ac1413f41/src/main/java/blusunrize/immersiveengineering/common/blocks/IETileProviderBlock.java#L378)

As you can notice, both mods crash on the same Forge method call. 

## How to reproduce
It seems like whenever the player is standing on top of a non-standard-shaped block, a Block#getCollisionShape() of null is returned and the next mod that tries to use it crashes. This happened to me on two separate mods - Immersive Engineering Conveyor Belt, and PneumaticCraft Air Pressurizer, both of which look non-standard.

wagyourtail:
player can get stuck on the "irregular" blocks if looking at the center of the block doesn't actually target it for breaking. I see that as "acceptable" but lmk if I should dynamically add them to blocksToAvoidBreaking as they come up

fixes #2148

### DIFF
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -39,6 +39,7 @@ import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.IBlockReader;
 
+import java.util.List;
 import java.util.Optional;
 
 import static baritone.pathing.movement.Movement.HORIZONTALS_BUT_ALSO_DOWN_____SO_EVERY_DIRECTION_EXCEPT_UP;
@@ -522,7 +523,12 @@ public interface MovementHelper extends ActionCosts, Helper {
                 || block instanceof ShulkerBoxBlock) {
             return false;
         }
-        return Block.isOpaque(state.getCollisionShape(null, null));
+        try {
+            return Block.isOpaque(state.getCollisionShape(null, null));
+        } catch (Exception ignored) {
+            // if we can't get the collision shape, assume it's bad and add to blocksToAvoid
+        }
+        return false;
     }
 
     static PlaceResult attemptToPlaceABlock(MovementState state, IBaritone baritone, BlockPos placeAt, boolean preferDown, boolean wouldSneak) {


ZacSharp (commented):
I think it should appear in the log when this fallback is used because it might cause unexpected behaviour. I don't know how often this can be expected to trigger and if it can trigger a lot I don't think logging a line every time is a good idea.

ZacSharp:
Does this problem only occur in 1.14+?

wagyourtail:
technically only 1.14 plus, collision couldn't (technically, i'm sure mods get around it as microblocks still existed) depend on world/blockpos in 1.13 and below

wagyourtail:
for logging with the fallback then we don't want to spam it just log when used once, so a HashSet of blocks maybe.
the only unexpected behavior I "expect" is getting stuck on weird shaped blocks... 

in testing with chisel&bits it does become possible to get stuck on it because it doesn't "look" at the hitbox properly for mining, this is probably fixable by adjusting the mining function to actually look at the center of the bounding box and not the blockpos
image

ZacSharp:
Aiming past blocks is a real issue not unexpected behaviour. I thought about things like c&b structures with a solid top or covered pipes where users might think Baritone should walk over it.

wagyourtail:
yeah, I guess that should be fixed... btw I do have to port to 1.16 for every test, 1.14 has no forge xd

SteamDoge9000:
when can this be expected for 1.16?

PR #3084, opened 2021-11-14
Settable IWorldScanner

wagyourtail:
Provide functionality for the `IWorldScanner` to be overridden.
this would allow (for example) a different mod to feed in ore locations for `MineProcess`

### TODO:
[] - make `MineProcess`'s `prune` function able to not remove ores that don't match the filter when the world scanner has been overriden


ZacSharp:
I don't think the call to `repack()` in `PathCommand` should be redirected, because it is meant to update the cache for the pathfinder and not to update the scanner for some following scan, like in `MineProcess` and `FarmProcess` (also why does `#path` repack but `#goto` doesn't?).
I'm not sure about the one in `RepackCommand` because its purpose could be either, maybe we should call it on both in case the scanner is overridden.

PR #3115, opened 2021-12-05
put a baritone installer inside baritone

wagyourtail:
:)

### DIFF
--- a/src/installer/java/baritone/install/BaritoneInstaller.java
+++ b/src/installer/java/baritone/install/BaritoneInstaller.java
@@ -1,0 +1,190 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


src/installer/java/baritone/install/BaritoneInstaller.java
```diff
@@ -0,0 +1,190 @@
...
+                ex.printStackTrace(new PrintWriter(writer));
+                JOptionPane.showMessageDialog(this, "Error installing Baritone: " + ex.getMessage() + "\n\n" + writer, "Error", JOptionPane.ERROR_MESSAGE);
+            }
+        });
+    }
+
+
+    public String findMinecraftDir() {
+        if (OSProber.getOS() == OSProber.OS.WINDOWS) {
+            return "C:\\Users\\" + System.getProperty("user.name") + "\\AppData\\Roaming\\.minecraft";
```
Soulevoker, line 171: Wouldn't it be better to pull the %appdata% variable?

Musings:
I'm not sure that if someone moved their AppData folder to a different drive and created a Symbolic Link that the appdata variable will be updated as well, and I'm not sure if the File class will automatically follow a sym link.

Sumandora: try to use user.home property instead of user.name. user.home gives you the same folder as doing %userprofile% in for example cmd would give.

MoMMde: and when e.g. mincraft is installed on the root user(i know you shouldn't do this) it doesn't work becouse the user home of `root` is `/root/` not `/home/root/`

Starmania:
Add a possibility to change the .mineecraft folder ( I think at MultiMC when I say that )

PR #3049, opened 2021-10-21
Merge features upwards

ZacSharp:
I don't know how you will merge this so I will only make one pr.
* 1.17.1 is at 0c0a0f2d1e5fa2be80cef7516cd6505b43f90d74 (this pr)
* 1.16.5 is at c2d22d1997caa631ddbd7061a907f26b3ee224b0
* 1.15.2 is at 4523a2d30e75b367649f89c5c760be1b58179ec0
* 1.14.4 is at 98f3373866def62ae2f8b15e40cec8e4fa7b423b
* 1.13.2 is at bb91dc98de8758926f431ed10be838fa3c223a32

The build fails for 1.15.2 and 1.16.5 because proguard can't handle mixin 0.8.4 and the build.grade specifies 0.8.+

PR #3134, opened 2021-12-11
buildSkipBlocks and buildValidSubstitutes

m-sarabi:
Added those missing settings for 1.17.1
I don't code java. Please if you know how to, update buildSubstitutes options for 1.17.1 and above


ZacSharp:
#3049 already ports `buildSkipBlocks` and `buildValidSubstitutes` and many other features to all supported versions so I think doing these two specifically is unnecessary.

Sorry for answering only after you had done your port.

PR #3148, opened 2021-12-17
Option to ignore direction of certain blocks during building

Entropy5:
Toggling this option will let baritone place glazed terracotta, bone blocks and hay bales freely.
Current behavior (with buildignoreexisting false) often includes continuous breaking and placing the same block, as baritone does not understand how to place a block so it faces the right way.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -259,6 +259,11 @@ public final class Settings {
     public final Setting<Boolean> buildIgnoreExisting = new Setting<>(false);
 
     /**
+     * If this is true, the builder will ignore directionality of certain blocks like glazed terracotta.
+     */
+    public final Setting<Boolean> buildIgnoreDirection = new Setting<>(false);
+
+    /**
      * If this setting is true, Baritone will never break a block that is adjacent to an unsupported falling block.
      * <p>
      * I.E. it will never trigger cascading sand / gravel falls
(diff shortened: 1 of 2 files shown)


ZacSharp (changes requested):
Nice someone actually went ahead and implemented this.
Two things though:
1. Place the check right after the check in `return current.equals(desired);` ([here](https://github.com/cabaletta/baritone/blob/3adfa16db8305a07200b29bcde32e3912699ee79/src/main/java/baritone/process/BuilderProcess.java#L864))
2. Instead of checking for specific blocks you should check whether `current.getBlock() == desired.getBlock()` (you already know `current.equals(desired) == false` because of 1.) and if the blocks are the same, check whether all properties except the rotation match.

Entropy5:
@ZacSharp I tested this and it seems to work without any additional property checks
https://github.com/Entropy5/baritone/commit/32a7657c4ffd933cba4ed8d289e4e6bb32b6cd39
Thoughts?

ZacSharp:
The problem with that approach is that it ignores *all* properties, not just direction so it is definitely not a good implementation for `buildIgnoreDirection`.
On the other hand a setting to ignore any properties might be useful in some cases, because Baritone can't really deal with anything beyond choosing the correct item and properties are often for things that can't be dealt with by choosing a different item (redstone power level, cauldron level, door state, lever flicking, plant groth, etc.). I never used the builder to build non-trivial schematics so I don't know whether such a setting  would help.

wagyourtail:
maybe pull this out to a `buildIgnoreNBT` option or something?

Entropy5:
@ZacSharp I realised this once baritone started placing granite instead of stone
made a new pull request https://github.com/cabaletta/baritone/pull/3159

PR #3151, opened 2021-12-17
Make rightClickContainerOnArrival click lit furnaces

ZacSharp:

### DIFF
--- a/src/main/java/baritone/process/GetToBlockProcess.java
+++ b/src/main/java/baritone/process/GetToBlockProcess.java
@@ -242,7 +242,7 @@ public final class GetToBlockProcess extends BaritoneProcessHelper implements IG
         if (!Baritone.settings().rightClickContainerOnArrival.value) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


wagyourtail:
maybe just make it click any block that has a block entity? not sure if that's a bad idea...

ZacSharp:
Don't think that's a good idea since there are tileentities one would not expect when looking for containers (though clicking beds could be declared a feature). Maybe there is something more specific we could check for. I'll have a look at it tomorrow.

ZacSharp:
There is definitely more tileentities than we want to click and while all containers seem to extend from `BlockContainer` that baseclass is also used by all blocks with a tileentity exept for beds and comparatos.
Going up the inheritance chain of the tileentity classes looks more promising and most of them should be covered by `TileEntityLockable`, with just a few like workbench,  anvil and enchantment table missing because they don't have persistent contents. I think it might be possible to get them by checking for the block being an `IInteractionObject` or `IInventory`, but these two interfaces are implemented by many more classes and interfaces in various packages so maybe it's better to just special case them.

PR #3159, opened 2021-12-21
Better buildignoredirection

Entropy5:
zac suggested doing this more broadly on the previous pull request


ZacSharp (changes requested):
Looks good, except for the formatting.
The `if`s in `sameWithoutOrientation` almost physically hurt me.

ZacSharp (approved):

leijurv:
Feels asymmetric. You can't check if two maps are equal by only iterating through map1. That will catch if map1 has something that map2 doesn't, but it won't catch if map2 has something that map1 doesn't.

Entropy5:
i see no difference in behavior but this is more correct i guess

ZacSharp:
The blocks are checked to be equal right at the top of the method and property keys are per block, so any two blockstates passing the first check should be guaranteed to have the same keys in their property maps.

In either case I don't think we should copy both maps into one here.

And unless there are conflicts you don't need to merge from upstream every time.

Entropy5:
ok i have no idea what im doing nor why it just merged the same thing again

Issue #186, opened 2018-09-25
Multiple goals simultaneously


ThatGrzegorzB:
Possible an option to run multiple goals simultaneously like following a player but mining blocks so depending on which goal it would prioritize, for example if follow player was first it would only mine if it was in range of the player, but if mine was first it would first mine everything it has then go to the player and start mining again

ThatGrzegorzB:
Yes preety much

leijurv:
@ThatGrzegorzB which one? @Thunder33345 gave two possibilities (1: it considers both goals and could go to either one, but it prefers the higher priority one, 2: it goes to the higher priority goal, and only after it gets there does it consider and path to the lower priority goal). Which one was your "Yes preety much" in response to?

ThatGrzegorzB:
Number 2, but it only considers the lower priority goal if the higher priority goal is finished, my idea was using this to follow players and for example mine or find blocks navigate to them and mine another block until the block it was supposed to find is gone

SylveonBottle:
I'd say we should keep #176 separate from #186. Queuing goals to complete in order, and doing things simultaneously seem a bit different.

`#mine log 64 & #goto home` = baritone should walk home, but occasionally mine logs if it gets the chance.  
`#mine log 64 ; #goto home` = baritone should mine logs, then once it's finished walk home.

Though, priority may be depneding on the order.
`#kill pig & #goto home` = Baritone should kill all pigs in sight. If none are found, start walking home until it finds more pigs.
`#goto home & #kill pigs` = Baritone should walk towards home, but if it happens to walk near a pig go ahead and kill it.

You could probably queue it beyond this too, alongside #176:
`#kill pig & #goto home ; #mine diamond_ore` = Baritone should kill pigs that it finds, walking home when there are no targets. Once finally home, start mining diamonds.

Combine with #708 to make a minimum amount of stuff.

`#goto home & #kill pig10 ; #mine diamond_ore 32 ; #goto home` = Baritone should walk home, but kill pigs when it can. If the user has 10 porkchops, ignore pigs and just head home. If the user eats and has less than 10 porkchop (which would happen once the previous pigs are out of render distance), just keep going home until the next time it finds pigs. Once home, mine 32 diamonds. Once mined & collected, head back home once more.

This results in a super long trip back home where baritone is restocking on food by itself, and once home will collect 32 diamonds before heading back home again.

Lengthy post but the possibilities are endless if you manage to add this sort of "baritone goal language".
[...]

Issue #176, opened 2018-09-20
Add a way to queue goals


leijurv:
e.g. "baritone please don't be smart and try and go diagonal. please just go from 0,0 to 25000,0 to 25000,2000 to 24500,2000 to the portal there" btw those aren't my real base coords i promise

plootia:
can confirm, don't bother going to those coords

ThatGrzegorzB:
please yes

ZeroMemes:
I assume the next goal in the queue would only be set if the current one is successfully reached, and if the current one fails all succeeding ones are dropped?

ZeroMemes:
Bumping this, it is something that I’m actually capable of doing

VolantTheFox:
bump! any news on this? it would be hella useful to go to a certain set of coords, then #mine <block> <amount>, and THEN go back to home base.

mrphax0r:
Would it take much more than a list tho?

GunnerBasil:
Bump. I would love this. like mine 3 diamonds, and make a pickaxe or something

dominikaaaa:
don't comment useless shit for no reason, unless you have something to contribute to the conversation, as it sends a notification. this isn't a forum

SylveonBottle:
Duplicate of #269, [according to l1ving](https://github.com/cabaletta/baritone/issues/269#issuecomment-751557525)

SylveonBottle:
I found this issue personally resolved by using [AltoClef](https://github.com/gaucho-matrero/altoclef).

More specifically, [this fork of altoclef](https://github.com/Meloweh/altoclef) which allows running chain commands.

Just incase anyone from the mentioned issues would like a temporary solution / something to toy with.

walksanatora:
yeah but we shouldn't have to use forks/addons to use a feature that should be in the base mod

SiffrinSuffrin:
Would also be cool to automatically set up bounding boxes, ie
#queue goto 0 0 0
#queue sel pos1
#queue goto 100 0 100
#queue sel pos2
#queue sel cleararea
#queue home

bp88dev:
Seems like this suggestion has lost its momentum, but I think this would be a great addition to Baritone.

ghost:
waiting for this solid 7years!!!

PR #3177, opened 2022-01-04
Add Task queues

ELordd:
This is a feature i wanted since last 3 months, i want the ability to do a task queue, for example, i do #taskqueue mine diamond_ore and then #start, in the meanwhile i can add more queues like #home, and once its done with the first task, it automatically goes to the other task so i dont have to take a year typing one command, i would really appreciate if you add this as a feature. this would be be very useful since i have more time to do stuff irl while baritone is working for me.


ZacSharp:
Why do people open pull requests when they want to open an issue?
Also this has already been suggested quite a few times.

ZacSharp:
Duplicate of #186

ZacSharp:
Duplicate of #176

PR #3077, opened 2021-11-11
1.18 support

wagyourtail:
@leijurv giv me a branch, xd

it works

- [x] - forge 1.18 support

blocks to fix

- [x] - azelea saplings

download artifacts at https://github.com/cabaletta/baritone/actions/runs/1702733433

PR #2293, opened 2020-12-24
SmoothAim while mining

AppleCidar:
Not the cleanest code, but nobody else would do it so I did. it works how its suppose to.
set to 1 to disable. issue #2034

PR #3201, opened 2022-01-14
SmoothAim upgraded to 1.18

SM1LLER:
Just implemented SmoothAim from this #2293 and using #3077 made it work on 1.18.
Use "#setting smoothAim int". It disabled if set '1'. Int is number of ticks baritone will need to rotate camera. Be careful with using it, cause it slows rotating.


wagyourtail:
for features, baritone only accepts PR's to the master branch, (or the lowest version that makes sense).

Saloouu:
Sorry guys but when i type #setting SmoothAim int ,the commands doesn't exsist.
I read the mapping, and this 'settings' it was not found.

SM1LLER:
> Sorry guys but when i type #setting SmoothAim int ,the commands doesn't exsist. I read the mapping, and this 'settings' it was not found.

It's not a part of baritone, i implemented it for me. U can find fork of baritone in my reps and there is branch with smoothAim source for baritone. Or u can dm me in Discord and i'll give u builded version with smoothAim and waitTicksBeforeAim settings.

PR #3237, opened 2022-02-01
Fix the builder place-break loop with carpet and similar blocks

ZacSharp:
Pathing sometimes planned a `MovementPillar` on a carpet, which was then broken by the movement and immediately placed back by the builder and broken again by the movement and so on.
This prevents considering non `canWalkOn` blocks as a throwaway for the schematic and also places the goal next to the blocks instead of on it. Additionally the builder is now allowed to place `canWalkThrough` blocks at eye height because it can't block itself with those but it helps with decorative blocks or carpet on stairs.
I also fixed the problem of standing partially inside a block by making the builder move to the centre of `pathStart` when it can't do anything.

New problems:
It cannot place unsupported carpet (and other `canWalkThrough` blocks which need support) anymore. It previously had trouble with those and with the schematic requesting air below it got stuck in a place-break loop for the supporting block, but the problem was obviously visible. Now it just stands there doing nothing but making sure it's in the centre of the block.
I could force it to place a block below by setting the goal into the block if it is unsupported, but that would only help for `canWalkThrough` blocks needing support from below and also cause blocks being placed under any other `canWalkThrough` block like torches or signs on walls.

### DIFF
--- a/src/main/java/baritone/process/BuilderProcess.java
+++ b/src/main/java/baritone/process/BuilderProcess.java
@@ -204,6 +204,9 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
         if (state.getBlock() == Blocks.AIR) {
             return null;
         }
+        if (!MovementHelper.canWalkOn(new BlockStateInterface(ctx),x,y,z,state)) { // throwaways have to be canWalkOn so we have to place a wrong block here, air is separate because it is so common
+            return null;
+        }
         return state;
     }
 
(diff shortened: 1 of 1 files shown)


leijurv:
This a bit too complicated with a bit too many tradeoffs for me to just merge it, can someone else test this and approve or disapprove?

ZacSharp:
Any good idea on how to detect the need for a supporting block? IMHO something to do that is definitely needed before this should be merged because standing around without any visible problem is much worse than visibly trying to place a block that can't be placed.
Are there other tradeoffs I missed?

TheContainer:
Why is this dead now for more than a year?

ZacSharp:
Because while this does fix the original problem it also introduces a major behavioral regression (see "New problems" in the original comment and my second comment).
If you have any ideas on how to address these new problems feel free to contribute them.

TheContainer:
> 



> Because while this does fix the original problem it also introduces a major behavioral regression (see "New problems" in the original comment and my second comment). If you have any ideas on how to address these new problems feel free to contribute them.

Ah thanks for the explanation

MrFums:
Do you potentially have a version that is built for 1.20 please?

PR #2430, opened 2021-02-05
immersive portal fix (part 1)

wagyourtail:
fix for immersive portals with baritone, 

It works, but baritone will try to path through the portal and stop when it changes dimensions, need to get the portal on the avoid block list somehow, not quite sure since it doesn't appear to actually be blocks.

PR #3247, opened 2022-02-04
Ported ImmersivePortals-Fix to 1.18.1

Tannex:
Spent a little bit of time to get @wagyourtail 's immersive portals fix (#2430 ) working on 1.18.1 Fabric.

Before merging anything, someone should probably check if it breaks the forge binary, as I haven't tested anything related to Forge.

I verified that the binary works on the `Valhelsia: Enhanced Vanilla - 1.18` (v1.0.7)  modpack.


scorbett123:
Sorry, but we always have to merge into the oldest branch that it makes sense to modify, this means that when the original pull is merged it will them be pulled into the 1.18 branch. This process wouldn't work so smoothly if these commits were merged into 1.18 first, so therefore this pull won't be merged.

However thanks for creating it as it will give us help porting versions in the future to use this as a reference.

Tannex:
> Sorry, but we always have to merge into the oldest branch that it makes sense to modify, this means that when the original pull is merged it will them be pulled into the 1.18 branch. This process wouldn't work so smoothly if these commits were merged into 1.18 first, so therefore this pull won't be merged.
> 
> However thanks for creating it as it will give us help porting versions in the future to use this as a reference.

Makes sense, just felt like it would be wrong not to share it in case it is useful to others.

PR #3249, opened 2022-02-05
Sort results of the `find` command by distance to player

EthanZeigler:
Sorts the results of the `find` command by the distance to the player. Makes reading the output much less frustrating.

Adds `distanceFrom` and `distanceFromSquared` utilities to `BetterBlockPos`

- Add distance between BetterBlockPos
- Sort find command results by distance
- Update find command usage documentation
- I've spent too much time in python, aka adding ;
- Comply to the Compatator lambda interface

### DIFF
--- a/src/api/java/baritone/api/utils/BetterBlockPos.java
+++ b/src/api/java/baritone/api/utils/BetterBlockPos.java
@@ -115,6 +115,34 @@ public final class BetterBlockPos extends BlockPos {
         return oth.getX() == x && oth.getY() == y && oth.getZ() == z;
     }
 
+    /**
+     * Squared distance from this object to another. Given a null
+     * BetterBlockPos, the returned distance is -1.
+     * 
+     * This is more performant than distanceFrom due to sqrt's relatively
+     * high cost and should be used unless the exact distance is needed. For
+     * comparisons, use this.
+     */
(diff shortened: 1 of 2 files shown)


ZacSharp:
If you want to make it really useful make the coordinates clickable to run `#goal <x> <y> <z>` when clicked (and maybe paste the coordinates into chat when shift clicked, iirc there's a click event for that as well)

EthanZeigler:
> If you want to make it really useful make the coordinates clickable to run `#goal <x> <y> <z>` when clicked (and maybe paste the coordinates into chat when shift clicked, iirc there's a click event for that as well)

I'll look into doing that sometime this week. It would be a good improvement.

scorbett123:
If possible please could you also make it output no blocks found if it doesn't find any of the blocks, otherwise it sometimes looks like it's not doing anything to users.

leijurv:
BlockPos already has a distance? I don't remember what it is off the top of my head but I'm certain this could instead be something like `.sorted(Comparator.comparingDouble(pos -> ctx.player().distanceSq(pos)))` or maybe `ctx.playerFee().distanceSq(pos)` or something like that.

leijurv:
closing due to inactivity, reopen if/when my previous comment is addressed

PR #3255, opened 2022-02-08
replaymod replay crash fix 1.17+ 

wagyourtail:
use world not worldData to prevent npe when in replaymod replay

oopsie while merging up

### DIFF
--- a/src/main/java/baritone/utils/BlockStateInterface.java
+++ b/src/main/java/baritone/utils/BlockStateInterface.java
@@ -43,7 +43,7 @@ public class BlockStateInterface {
 
     private final ClientChunkCache provider;
     private final WorldData worldData;
-    protected final BlockGetter world;
+    protected final Level world;
     public final BlockPos.MutableBlockPos isPassableBlockPos;
     public final BlockGetter access;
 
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/utils/BlockStateInterface.java
```diff
@@ -97,9 +97,9 @@ public BlockState get0(BlockPos pos) {
     }
 
     public BlockState get0(int x, int y, int z) { // Mickey resigned
-        y -= worldData.dimension.minY();
+        y -= world.dimensionType().minY();
```
wagyourtail, line 100: does JIT optimize these function calls away, or should I move this to a `private final int`

dimensionType just returns a final field
minY is also just returning a final field

Martine1231: can you explain how you fixed this in low IQ speech. I appreciate it

wagyourtail: world -> worldData as world can sometimes be null

Martine1231: Capture
checking through but I cant find a file for worldData

wagyourtail: :rock: :brain:

EvilEpicPanda:
I'm confused. Issues get closed citing this issue. Is there a fix for this or are we waiting for it to be fixed?

wagyourtail:
well, has there been a release of baritone for (insert version you still experience this on) since feb 8 2022 when this was fixed

EvilEpicPanda:
> well, has there been a release of baritone for (insert version you still experience this on) since feb 8 2022 when this was fixed

Yes. I was using 1.9.4, latest release.

SunfxDEV:
Im also using 1.9.4 with a modpack and im getting the same errors, any fix?

PR #3276, opened 2022-02-19
Clarify License.

Dev380:
Added a clear statement of the licence under the README, and adding an agreement for contributors to licence their code under the LGPL v3.

Changing licence files to follow GNU's preferred format.
https://www.gnu.org/licenses/gpl-howto.html

> You should also include a copy of the license itself somewhere in the distribution of your program. All programs, whether they are released under the GPL or LGPL, should include the text version of the GPL. In GNU programs we conventionally put the license in a file called COPYING.
...
If you are releasing your program under the Lesser GPL, you should also include [the text version of the LGPL](https://www.gnu.org/licenses/lgpl.txt), usually in a file called COPYING.LESSER. Please note that, since the LGPL is a set of additional permissions on top of the GPL, it's crucial to include both licenses so users have all the materials they need to understand their rights.


leijurv:
Screen Shot 2022-07-05 at 3 29 50 PM

This indent and codeblock thingy looks bad and weird. I'm not sure we want this disclaimer in the readme

0-x-2-2:
Closed due to inactive.

PR #3309, opened 2022-03-09
fix inner class attribute being stripped

wagyourtail:
this fixes Setting.value issue in ide's when using the api jar as a dependency

why didn't I do this sooner. I've known about this issue for like a year...


scorbett123 (approved):
Thanks for looking into this and fixing it.

ZacSharp (approved):

wagyourtail:
It just keeps the relationship that its inner, its an attribute so I dont think you can specify per class what to keep...

i dont really know proguard /shrug

PR #3317, opened 2022-03-12
Add support for the Structure Block file format (.nbt files)

### LINKED ISSUE #750: allow #build to use .nbt files
Currently #build uses .schematic files, An old and now broken format. .nbt files (ones a structure block would save) use 1.14 type string IDs and no more bytelists. They’re both NBT type files at heart, so no new libraries or code will be needed to read the files.

Aryezz:
This adds support for .nbt files with the `#build` command. Small demo here: https://youtu.be/Gnp5JsKG9Ug

Closes #750 

Notes:
Tile-Entity data doesnt get read. Schematica supports this and since this is basically copy and paste from its source, it would not be difficult to add support (but i dont know enough about Baritone to know if it would make sense).
Check out the Schematica source here: https://github.com/Lunatrius/Schematica/blob/master/src/main/java/com/github/lunatrius/schematica/world/schematic/SchematicStructure.java#L30-L43

Also note that this is the first time i've ever done anything with Java, so you might wanna check everything twice :)

### DIFF
--- a/src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
+++ b/src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.utils.schematic.format.defaults;
+
+import baritone.utils.schematic.StaticSchematic;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.gen.structure.template.Template;
+
+import java.lang.reflect.Field;
+import java.util.List;
+
+/**
+ * @author Sam
+ * @since 03/11/2022
+ */
+public final class StructureNBT extends StaticSchematic {
+
+    public StructureNBT(NBTTagCompound schematic) {
+        Template template = new Template();
+        template.read(schematic);
+
+        BlockPos size = template.getSize();
+        this.x = size.getX();
+        this.y = size.getY();
+        this.z = size.getZ();
+
+        List<Template.BlockInfo> blocks;
+        try {
+            Field f = template.getClass().getDeclaredField("field_186270_a"); // Template.blocks
+            f.setAccessible(true);
+            blocks = (List<Template.BlockInfo>) f.get(template);
+        } catch (final Exception e) {
+            // This should never happen since SRG names are canonical
+            throw new IllegalStateException("reflection error");
+        }
+
+        this.states = new IBlockState[this.x][this.z][this.y];
+        for (Template.BlockInfo block : blocks ) {
+            this.states[block.pos.getX()][block.pos.getZ()][block.pos.getY()] = block.blockState;
+        }
+    }
+
+    @Override
+    public boolean inSchematic(int x, int y, int z, IBlockState currentState) {
+        // Filtering out Structure Void
+        if (x >= 0 && x < widthX() && y >= 0 && y < heightY() && z >= 0 && z < lengthZ()) {
+            return this.states[x][z][y] != null;
+        }
+        return false;
+    }
+}
--- a/src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
+++ b/src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
@@ -21,6 +21,7 @@ import baritone.api.schematic.IStaticSchematic;
 import baritone.api.schematic.format.ISchematicFormat;
 import baritone.utils.schematic.format.defaults.MCEditSchematic;
 import baritone.utils.schematic.format.defaults.SpongeSchematic;
+import baritone.utils.schematic.format.defaults.StructureNBT;
 import net.minecraft.nbt.CompressedStreamTools;
 import net.minecraft.nbt.NBTTagCompound;
 import org.apache.commons.io.FilenameUtils;
@@ -64,6 +65,18 @@ public enum DefaultSchematicFormats implements ISchematicFormat {
                 default:
                     throw new UnsupportedOperationException("Unsupported Version of a Sponge Schematic");
             }
+        }
+    },
+
+    /**
+     * The vanilla structure block file format. Commonly denoted by the ".nbt" file extension.
+     *
+     * @see <a href="https://minecraft.fandom.com/wiki/Structure_Block_file_format>Structure Block file format</a>
+     */
+    STRUCTURE("nbt") {
+        @Override
+        public IStaticSchematic parse(InputStream input) throws IOException {
+            return new StructureNBT(CompressedStreamTools.readCompressed(input));
         }
     };
 


src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
```diff
@@ -0,0 +1,68 @@
...
+
+        List<Template.BlockInfo> blocks;
+        try {
+            Field f = template.getClass().getDeclaredField("field_186270_a"); // Template.blocks
+            f.setAccessible(true);
+            blocks = (List<Template.BlockInfo>) f.get(template);
+        } catch (final Exception e) {
+            // This should never happen since SRG names are canonical
+            throw new IllegalStateException("reflection error");
+        }
```
ZacSharp, line 52: This is a no go. It only works when loading Baritone as a Forge mod in a non-dev enviroment.
Cases where Baritone has to work:
* Baritone dev env, mcp or official names
* Vanilla minecraft, obfuscated names
* Forge, srg names
* Fabric, yarn names
* As a library in other dev envs, mcp or official or yarn names

Of course it is an option to look up all possible names and check them all but that would mean manually validating the official, yarn, srg and obf name for every minecraft version to come. Instead you should use mixin so the mixin preprocessor and the transformers in Forge and fabric take care of the different mappings just like for any other name.

Look at the interfaces in `src/main/java/baritone/utils/accessor` and their implementations `src/launch/java/baritone/launch/mixins` for working and used examples. (and feel free to ask questions)

ZacSharp (changes requested):

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
Aryezz: Ooooh, very interesting, i'll look into it.
Kinda thought that hardcoding fieldnames was whack but didn't know about Mixins.

Aryezz: I made a new commit but im not sure if i've done it right. It seems to work, but it feels like it shouldn't...

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public boolean inSchematic(int x, int y, int z, IBlockState currentState) {
+        // Filtering out Structure Void
+        if (x >= 0 && x < widthX() && y >= 0 && y < heightY() && z >= 0 && z < lengthZ()) {
+            return this.states[x][z][y] != null;
+        }
+        return false;
+    }
```
ZacSharp, line 58: If this check actually filters out null states, can you maybe try using a schematic with null states with `mapArtMode` enabled?

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
ZacSharp: It's just injecting a bit of code at runtime.
When the game is running mixin merges the `MixinTemplate` class into the `Template` class so it implements the interfaces and has the methods (`@Accessor` is a shorthand for a getter on the field with the corresponding name) and fields.

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
Aryezz: What would the right aproach to fix this be? We want to prevent any `null` of leaking through `getDirect` and `getColumn`, but there would have to be some sort of placeholder (because otherwise there would be a hight shift in the column).

The easiest solution i can think of is to fill `this.states` with Air blocks instead of null while still keeping track of the structure void. That way if the blocks get read without checking `inSchematic` there would not be a NullPointerException, but the blocks would hopefully get ignored when building.

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
Aryezz: I understand whats happening, but it just seems kinda weird since im never actually referencing the field that i want the getter to return, it just reads the function name and parses out the field name, which to me seems weird.

But hey, on the fabric wiki it said thats the way to do it, and it works, so who am i to judge ._.

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
Aryezz: Ok, so i fixed this by overriding `getDirect`, `getColumn` to use air as placeholders and `desiredState` to return the current block.
There is still `null` in `this.states`, but since that can only be accessed from within subclasses (as far as i understand) it shouldn't be a problem.

ZacSharp (approved):
In my opinion this can be merged now, though I still have some thoughts/questions regarding the problem with `MapArtSchematic`.

If I have understood this correctly (can't test currently) `minecraft:structure_void` is not saved to *.nbt structure files so those contain holes where the structure void was, just like a litematic contains "holes" in areas not covered by one of its regions. If that's wrong please tell me and ignore the rest of this comment.
The current approach to patch those holes with air works without problem when the schematic is used as a normal dynamic schematic and the only use as a static schematic within Baritone is `MapArtSchematic` which still behaves correctly with the filled holes and I don't think there is a lot of usages by other parties, so I think we can use it. However I wonder whether static schematics are really supposed to bypass the `inSchematic` check. Before the attempts to read *.nbt and *.litematic files all supported formats were only capable of saving a single cuboid and `MapArtSchematic` can only be instantiated with schematics created by Baritone itself so when it was written it was safe to assume that no blockstate would ever be null, regardless of whether `getDirect()` and `getColumn()` are actually allowed to return null states.
My suggestion would be to explicitly allow those methods to return null for positions where `inSchematic` returns false (and maybe add a variant of `inSchematic` to `IStaticSchematic` that does not take the current state and returns true by default for compatibility)

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
ZacSharp: It's not tedious.
My current fight against the gradle cache is tedious.

Aryezz:
> If I have understood this correctly (can't test currently) `minecraft:structure_void` is not saved to *.nbt structure files so those contain holes where the structure void was

Yes, thats right. I guess it could be added manually, but the idea is that structure void blocks don't get saved when exporting.

ZacSharp (approved):
I have now tested this and it seems to work.

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
```diff
@@ -0,0 +1,79 @@
...
+            // Use air blocks as placeholder for Structure Void
+            if (column[i] == null) { column[i] = Blocks.AIR.getDefaultState(); }
+        }
+        return column;
+    }
+
+    @Override
+    public boolean inSchematic(int x, int y, int z, IBlockState currentState) {
+        // Filtering out Structure Void
+        return (super.inSchematic(x, y, z, currentState) && this.states[x][z][y] != null);
```
leijurv, line 77: remove outer parentheses

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public IBlockState getDirect(int x, int y, int z) {
+        // Use air blocks as placeholder for Structure Void
+        return desiredState(x, y, z, Blocks.AIR.getDefaultState(), null);
+    }
+
+    @Override
+    public IBlockState[] getColumn(int x, int z) {
+        IBlockState[] column = this.states[x][z];
```
leijurv, line 66: This doesn't make a copy, this actually takes the `states[x][z]`. That means that when you do `column[i] = Blocks.AIR.getDefaultState()`, it actually updates the value in `states[x][z][i]` from `null` to `Blocks.AIR.getDefaultState()`. I worry this could mess up the logic in `inSchematic` which checks if `this.states[x][z][y] != null`, as well as the logic in `desiredState` which also checks for null.

leijurv (changes requested):

src/main/java/baritone/utils/schematic/format/defaults/StructureNBT.java
Aryezz: I added a shallow clone in ad4f257, since it's only a 1D array and the objects just get replaced it should be enough (except if anything that uses `getColumn` changes the BlockStates but that would be a problem with all Schematic Formats).

rycbar0 (commented):
This looks good to me but i dont understand why you overide this mothods in your StrucureNBT class? isnt the structure void suposed to not replace the block that is already there? i would suggest re-merg the master branch and delete the override methods. after that i think this can be merged.

ZacSharp:
If you mean `getDirect` and `getColumn` swapping out `null` for air, that's because `MapArtSchematic` (and possibly third parties) assume the states are not null.
This works within Baritone because `MapArtSchematic` filters out the air anyway and `BuilerProcess` can and does use `inSchematic`. For third parties it can cause structure void to become air, but in my opinion that's better than crashing them.

Adding `boolean inSchematic(int x, int y, int z)` to `IStaticSchematic` (with the same default implementation as in `ISchematic`) would probably be a good idea though. That way, if we keep patching holes with air, old code won't crash while new code can check for the holes and exclude them from whatever it's doing.

PR #3326, opened 2022-03-16
allow baritone to walk on waterlogged topslabs/stairs

wagyourtail:
should this apply to anything else?

it appears to have an issue with going down waterlogged stairs and I'm not sure why...


StealthyPony:
Trapdoor (all variants)

wagyourtail:
in my testing it appeared to work with trapdoor...

leijurv:
I remember testing this in late 2019-ish, and making stairs `canWalkOn` made it do stupid stuff with failing parkour jumps. I think this needs a check to see if the stair is upside down. Right side up stairs shouldn't be canWalkOn if I remember correctly

PR #3327, opened 2022-03-16
If context.player() returns none we don't want to use its position for pruning.

### LINKED ISSUE #3289: baritone.dm.a() returns null
Sometimes when I try to load a world, I get this:

```
[01:11:34] [pool-14-thread-2/ERROR]: Uncaught exception in thread "pool-14-thread-2"
java.lang.NullPointerException: Cannot invoke "net.minecraft.class_746.method_19538()" because the return value of "baritone.dm.a()" is null
	at Not Enough Crashes deobfuscated stack trace.(1.18.1+build.22) ~[?:?]
	at baritone.dm.a(Unknown Source) ~[baritone-standalone-fabric-1.8.2.jar:?]
	at baritone.eo.c(Unknown Source) ~[baritone-standalone-fabric-1.8.2.jar:?]
	at baritone.eo.b(Unknown Source) ~[baritone-standalone-fabric-1.8.2.jar:?]
	at baritone.eo.d(Unknown Source) ~[baritone-standalone-fabric-1.8.2.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
```

scorbett123:
This fixes a really rare error which will crash the region saving thread, thus making it not save regions when they are explored.

This crash of that thread only occurs when time loading the minecraft world is > 30 seconds, making it pretty rare, however in these cases IPlayerContext.player() returns null, so we need to protect against this in guessing the position.

closes #3289

### DIFF
--- a/src/main/java/baritone/cache/CachedWorld.java
+++ b/src/main/java/baritone/cache/CachedWorld.java
@@ -209,7 +209,7 @@ public final class CachedWorld implements ICachedWorld, Helper {
     private BlockPos guessPosition() {
         for (IBaritone ibaritone : BaritoneAPI.getProvider().getAllBaritones()) {
             IWorldData data = ibaritone.getWorldProvider().getCurrentWorld();
-            if (data != null && data.getCachedWorld() == this) {
+            if (data != null && data.getCachedWorld() == this && ibaritone.getPlayerContext().player() != null) {
                 return ibaritone.getPlayerContext().playerFeet();
             }
         }


ZacSharp (approved):

leijurv:
Nice!

PR #3356, opened 2022-03-25
add `allowBreakAnyway` setting

### LINKED ISSUE #3316: add allowOnlyBlocksCanBreaking and BlocksCanBreakingOnly in #set
## allowOnlyBlocksCanBreaking
type: Boolean
Default Value: false
### description: 
**""set blocks can break in BlocksCanBreakingOnly""** (*or*) **""use '#set BlocksCanBreakingOnly' to set blocks can break""**
## BlocksCanBreakingOnly
type: List\<block>
Default Value: none
# 
#### When allowOnlyBlocksCanBreaking is true, BlockToAvoidBreaking will temporarily be invalid.

wagyourtail:
add a setting to have blocks that baritone is allowed to break even with `allowBreak` set to `false`

closes #3316

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -52,7 +52,7 @@ public final class Settings {
     /**
      * Blocks that baritone will be allowed to break even with allowBreak set to false
      */
-    public final Setting<List<Block>> forceAllowBreak = new Setting<>(new ArrayList<>());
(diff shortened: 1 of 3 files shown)


ZacSharp:
I think `MineProcess` should react to this appropriately (allow mining with `allowBreak` disabled as long as the target blocks are in `allowBreakAnyway`).
Automatically whitelisting the target block would be even better, but I don't think temporarily changing the setting value is a good idea.

Issue #3096, opened 2021-11-20
1.17 build fails on proguard


ZacSharp:
## Some information
Operating system: Ubuntu 20.04.3 LTS
Java version: 16.0.2+7 (Adopt-Hotspot)
Minecraft version: 1.17.1
Baritone version: 1.7.1 (749fd4fac3448d82201f9566dc4db37a289a2071)
Other mods (if used): none

## Exception, error or logs
<details><summary>gradle log</summary><pre>
2021-11-16T23:34:15.0995133Z Downloading https://services.gradle.org/distributions/gradle-7.0.2-bin.zip
2021-11-16T23:34:16.7526020Z ..........................................................................................................
2021-11-16T23:34:17.8785428Z 
2021-11-16T23:34:17.8820790Z Welcome to Gradle 7.0.2!
2021-11-16T23:34:17.8827250Z 
2021-11-16T23:34:17.8828556Z Here are the highlights of this release:
2021-11-16T23:34:17.8838196Z  - File system watching enabled by default
2021-11-16T23:34:17.8862364Z  - Support for running with and building Java 16 projects
2021-11-16T23:34:17.8870001Z  - Native support for Apple Silicon processors
2021-11-16T23:34:17.8871199Z  - Dependency catalog feature preview
2021-11-16T23:34:17.8872188Z 
2021-11-16T23:34:17.8873506Z For more details see https://docs.gradle.org/7.0.2/release-notes.html
2021-11-16T23:34:17.8874187Z 
2021-11-16T23:34:17.9780996Z Starting a Gradle Daemon (subsequent builds will be faster)
2021-11-16T23:34:46.9794039Z > Task :buildSrc:compileJava
2021-11-16T23:34:47.0781219Z > Task :buildSrc:compileGroovy NO-SOURCE
2021-11-16T23:34:47.0809778Z > Task :buildSrc:processResources NO-SOURCE
2021-11-16T23:34:47.0810434Z > Task :buildSrc:classes
2021-11-16T23:34:47.0810878Z > Task :buildSrc:jar
2021-11-16T23:34:47.0811880Z > Task :buildSrc:assemble
2021-11-16T23:34:47.0812912Z > Task :buildSrc:compileTestJava NO-SOURCE
2021-11-16T23:34:47.0813919Z > Task :buildSrc:compileTestGroovy NO-SOURCE
2021-11-16T23:34:47.0814978Z > Task :buildSrc:processTestResources NO-SOURCE
[...]

PR #3358, opened 2022-03-25
Build fix

stackmagic:
The build failed for me and also the docker build would fail. The built artifacts are always named "baritone....jar" but the BaritoneGradleTask would look for a jar named "code....jar" (in the case of docker) because that's the name of the directory where the sources were in.

Setting the rootProject's name property to "baritone" fixes this and the build will work even when the project is located in a directory named differently than "baritone".


ZacSharp:
Can you please check whether this only applies to 1.18?
There are reasons (namely #3096) for me to think this applies to 1.17.1 as well. If I'm right about that please change the pr to target the 1.17.1 branch instead.

wagyourtail:
this can probably be pulled back to master and not cause any issues..

stackmagic:
I'll check what happens on master (that's 1.12 I presume?)

Issue #928, opened 2019-09-11
Option to preview path before executing


user670:
## Describe your suggestion
I've had a few instances where Baritone's path finding break some of my structures and contraptions. It might arguably be helpful to allow an optional secondary confirmation to path finding commands, showing path (but not actually moving the player) before the player confirms (and before mistakes were made). Also can be used when the player doesn't trust Baritone's movement and want to take controls themselves.

## Settings
A setting like secondaryConfirmation, which when enabled, will cause path finding commands to wait for a #confirm command before actually taking over the controls.

## Context
part of base broken by path finding, also distrust of the bot when moving through areas with mobs (at night or in caves)

## Final checklist
- [x] I have not used any OwO's or UwU's in this issue.

jonnyawsom3:
sounds good, almost like what should happen for any travel command where you have to say "path" before continuing

iamlcarus:
I inadvertently made a duplicate of this issue, that being the above. Take it as my show of support for this enhancement, that I'd also like to see it added myself.
You can already sort of do it in a very janky way, by telling the AI to do something and then immediately pausing it. This will stop the AI from moving, but still show what the UI will do. This solves the "preview" part of things, but since this image doesn't update while paused, its not the best way for the player to use commands while still having control, as they'd have to continually manually update it themselves by re-executing and pausing the command.

PR #3359, opened 2022-03-25
Add Highlight Command

stackmagic:
I hope this is a desired feature in baritone. I made this because running the 'advanced' clients is a hassle and most are way out of date.

This allows the player to highlight any nearby blocks the exact same way they are shown while mining - except we're not pathing to the blocks and trying to break them. The player remains in full control.

The Implementation is based on extending the MineProcess. I had to make the MineProcess non-final and make a bunch of properties/methods protected in order to re-use existing code.

Let me know if there's a better way to implement this or if you want me to change anything, maybe refactor the common code out into a common superclass for the two Process classes. I didn't  do this refactoring as I first wanted to get your feedback and change as little as possible in this first iteration.

### DIFF
--- a/src/api/java/baritone/api/IBaritone.java
+++ b/src/api/java/baritone/api/IBaritone.java
@@ -59,6 +59,12 @@ public interface IBaritone {
     IMineProcess getMineProcess();
 
     /**
+     * @return The {@link IHighlightProcess} instance
+     * @see IHighlightProcess
+     */
+    IHighlightProcess getHighlightProcess();
+
+    /**
      * @return The {@link IBuilderProcess} instance
      * @see IBuilderProcess
      */
--- a/src/api/java/baritone/api/process/IHighlightProcess.java
+++ b/src/api/java/baritone/api/process/IHighlightProcess.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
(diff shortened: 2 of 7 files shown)


ZacSharp:
This looks like an ugly way to sneak hack client features into Baritone.
Imo a feature that would fit Baritone much better is #928 and it's duplicates.

scorbett123:
I think in its current state this is out of scope for baritone.

If this was made into a setting which allowed the preview of the path as @ZacSharp suggested I think then we could consider this more. This would entail it working as a setting, not its own independent process.

stackmagic:
> This looks like an ugly way to sneak hack client features into Baritone.

True. Honestly, I'd be very happy if all I need for some casual 2b2t is baritone, tweakeroo and a few other mods on an otherwise vanilla client. But I suspect adding tracers or highlighting players/mobs is probably out of scope for baritone?

> Imo a feature that would fit Baritone much better is #928 and it's duplicates.

I'll have a go at that. Sounds like less code changes too.

> EDIT: In any case the pr should target master and not 1.18.2

Thanks for letting me know!

scorbett123:
> True. Honestly, I'd be very happy if all I need for some casual 2b2t is baritone, tweakeroo and a few other mods on an otherwise vanilla client. But I suspect adding tracers or highlighting players/mobs is probably out of scope for baritone?

Baritone isn't meant to be a be all and end all solution to anarchy servers. It is meant to be highly focussed on pathfinding, and only really have features related to this.

PR #3360, opened 2022-03-25
Merge master into 1.16.5

ZacSharp:
I'm currently trying to get architectury-loom to work with my setup (It's a little more difficult than for ForgeGradle) and if I manage to do so I will also continue merging to 1.17.1, 1.18.1 and 1.18.2.
If prs are merged I will update and force push. I promise I won't let this one die like the last one.


ZacSharp:
Oh and tell me whether I should include fixes in this pr. The failing checks should all be from the commits I merged.

leijurv:
I think we'd want to skip the deprecated branch (1.13 1.14 1.15). But I guess it doesn't matter, because the PR is only into 1.16.5?

leijurv:
@ZacSharp 1.16.5 doesn't build anymore 😭  https://github.com/cabaletta/baritone/runs/5778320718?check_suite_focus=true

ZacSharp:
It stopped building a few months ago (when mixin 0.8.4 released) but the CI never had a reason to run. That's why I asked about including fixes, but I guess I can just as well create a separate pr either to upgrade proguard or to fix mixin to 0.8.3 (which is what I'm doing for myself all the time).
`runClient` is also broken btw.

ZacSharp:
I merged via the deprecated branches because I'm doing it anyway. If you don't want to have the intermediary merges I can squash them away next time.

leijurv:
Understood, if merging up one-branch-at-a-time is the way you like to do it, then that's fine. Just wanted to let you know that that's not needed anymore.

PR #3368, opened 2022-03-31
Remove false information from `#help mine`

ZacSharp:
Baritone no longer caches ores on Minecraft 1.13+
Metadata values were removed in Minecraft 1.13
lit_redstone_ore was merged into redstone_ore in Minecraft 1.13

### DIFF
--- a/src/main/java/baritone/command/defaults/MineCommand.java
+++ b/src/main/java/baritone/command/defaults/MineCommand.java
@@ -70,7 +70,7 @@ public class MineCommand extends Command {
                 "Also see the legitMine settings (see #set l legitMine).",
                 "",
                 "Usage:",
-                "> mine diamond_ore - Mines all diamonds it can find.",
+                "> mine diamond_ore - Mines all diamonds it can find."
         );
     }
 }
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/command/defaults/MineCommand.java
```diff
@@ -65,14 +65,12 @@ public String getShortDesc() {
...
-                "The specified blocks can be ores (which are commonly cached), or any other block.",
+                "The specified blocks can be ores, or any other block.",
                 "",
                 "Also see the legitMine settings (see #set l legitMine).",
                 "",
                 "Usage:",
-                "> mine diamond_ore - Mines all diamonds it can find.",
-                "> mine redstone_ore lit_redstone_ore - Mines redstone ore.",
-                "> mine log:0 - Mines only oak logs."
+                "> mine diamond_ore - Mines all diamonds it can find."
```
scorbett123, line 73: If we are updating this why don't we add a example with a quantity
```suggestion
                "> mine diamond_ore - Mines all diamonds it can find."
                "> mine 5 diamond_ore - Mines 5 diamonds"
```

ZacSharp: Agree on that, but won't create another pr just for a single line.
Fixing dozens of minor issues with one pr each feels wrong. Can we maybe have some list where we can collect them and then from time to time fix them at once? Just from a quick look I already found another one in the usage help of `#build` and my not even a week old list of such things already contains a couple more (one wrong javadoc, some click events not working)

scorbett123: I'll create an issue where we can list small documentation problems and fix them all in one go after we have found enough to make it worth it

PR #3371, opened 2022-04-02
Update Proguard

ZacSharp:
The current 0.8.+ version of Mixin uses a classfile format that Proguard 6 can't handle.
Alternatively we could fix our Mixin dependency to 0.8.3


wagyourtail:
fixing the mixin dependency to the java 8 version is the proper solution imo

wagyourtail:
#3372 was merged

PR #3373, opened 2022-04-03
Rework buildscript to work with maven publish

wagyourtail:
rework of the buildscript to simplify so `gradlew build` will build forge/fabric jars together

current issues:

- [ ] vanilla jar not remapping refmap properly
- [ ] proguard not finding mixin dependency


wagyourtail:
https://jitpack.io/com/github/wagyourtail/baritone/c3c88c50d1/build.log
:sunglasses:

PR #3387, opened 2022-04-11
Merge master into 1.16.5

ZacSharp:
Please don't merge immediately before or after updating master.


wagyourtail:
gh action only builds tweaker jar, forge/fabric missing

ZacSharp:
That's the part that is really annoying about updating things. If I miss a mapping change the compiler yells at me, if I forget functionality changes the mistakes sneaks in silently.
Thanks for pointing that one out, I'll make another pr to fix it (plus some `component.setStyle()` calls).

PR #3388, opened 2022-04-11
Improve output to user in eta command

scorbett123:
Currently it outputs in ticks, however it is easier to understand when there is also a second estimate.

### DIFF
--- a/src/main/java/baritone/command/defaults/ETACommand.java
+++ b/src/main/java/baritone/command/defaults/ETACommand.java
@@ -47,21 +47,17 @@ public class ETACommand extends Command {
         }
         IPathingBehavior pathingBehavior = baritone.getPathingBehavior();
 
-        Optional<Double> ticksRemainingInSegment = pathingBehavior.ticksRemainingInSegment();
-        Optional<Double> ticksRemainingInGoal = pathingBehavior.estimatedTicksToGoal();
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/command/defaults/ETACommand.java
```diff
@@ -45,12 +46,22 @@ public void execute(String label, IArgConsumer args) throws CommandException {
...
-                "Next segment: %.2f\n" +
-                "Goal: %.2f",
-                pathingBehavior.ticksRemainingInSegment().orElse(-1.0),
-                pathingBehavior.estimatedTicksToGoal().orElse(-1.0)
-        ));
+
+        Optional<Double> ticksRemainingInSegment = pathingBehavior.ticksRemainingInSegment();
+        Optional<Double> ticksRemainingInGoal = pathingBehavior.estimatedTicksToGoal();
+
+        if (ticksRemainingInGoal.isPresent() && ticksRemainingInSegment.isPresent()) {
```
ZacSharp, line 53: `ticksRemainingInSegment` shouldn't be required to show `ticksRemainingInGoal` because if planning ahead is too slow there won't be a current segment but Baritone is actually pathing and the full path eta should be shown.
I'd suggest using "N/A" as the default rather than not showing them or defaulting to -1

ZacSharp (changes requested):

ZacSharp (approved):

PR #3389, opened 2022-04-11
Add only mining in of selections

### LINKED ISSUE #2249: Legitmine for a Selection
## Describe your suggestion
This expands on the #sel command by adding a way to legitmine in the selection I have and that selection only. So basically, I select a region and then legitmine in that region only.

## Settings
Just like there is a "#sel replace block with block", this could also add a functionality called "#sel mine". If legitmine is activated, then "#sel mine" will do what legitmine does, just inside the selected area. If not, "#sel mine" will do what "#sel replace" does, but with the added ability to specify more than one ore to mine.

## Context
I am currently making a perimeter in Minecraft. I want to be able to mine out the ores the TNT machine reveals, and those ores only. However, the TNT machine also reveals a lot of caves and ravines. #mine gets confused and starts heading down the caves. I came back from being AFK and found that my player was hundreds of blocks away from where I wanted to mine. And also, I know plenty of people who could benefit from something like this. Want to mess with your friends by using automine in a game that has a limited area to mine for ores? Then this #sel legitmine will come in handy.

## Final checklist
- [x] I know how to properly use checkboxes 
- [x ] I have not used any OwO's or UwU's in this issue.

### LINKED ISSUE #2597: Blacklisted 'zones'

### LINKED ISSUE #3320: Adding nearby item picking up or that @mine works only in selected area
## Describe your suggestion
I need to @mine but only in selected plot of land

Currently we are able to select area and command baritone to dig certain blocks in it but **it is not picking items up**.
Would be great if you would be able to add some _#set_ for picking up list of items that are nearby, or somehow merge @mine with
area selection.

## Settings

_@set mineOnlyInSeletedArea true/false_

or something along

_@set nearbyItemPickup (Boolean)_
_@set nearbyItemPickup (List<>)_
_@set nearbyItemPickupRange (integer)_

## Context
I dont know basically anything about programming. I am lazy ass bitch who uses baritone to farm/do stuff on some kids server while i sleep or watch some videos.
It is pain in the ass that i have to alt tab to minecraft every 4min to collect items that have fallen to the ground before they despawn.

_#set nearbyItemPickup_ would be great multipurpose addition. It could be used from picking up samplings from fallen trees to collecting loot on some farms.

I dont know if nearbyItemPickup would be importand enough to make it into commang like @farm, @follow, @mine, @goto or others, so my proposition is to add it as setting that is executed in the background. It would execute main command for some time for example **@sel ca** some amount of blocks and then collect everything that fallen out of them, just like in @mine.

## Final checklist
- [ ] I know how to properly use check boxes 

scorbett123:
Filter out any positions that aren't valid according to only mining inside selections, or only mining outside of selections.

I'm not really sure about the setting naming here, or even if this is the best way to do the settings, so any input on this would be appreciated.

partially fixes #3320 (this contains two suggestions)
~~fixes #2597~~ 
fixes #2249

~~This could also be seen as a fix for mining around "home" (#2054) however that could be made to work nicer~~

### DIFF
--- a/src/main/java/baritone/process/MineProcess.java
+++ b/src/main/java/baritone/process/MineProcess.java
@@ -425,7 +425,7 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
                     if (Baritone.settings().mineOnlyInsideSelection.value) {
                         boolean isInSelection = Arrays.stream(ctx.baritone.getSelectionManager().getSelections()).anyMatch(selection -> selection.aabb().contains(new Vec3d(pos.getX(), pos.getY(), pos.getZ())));
 
-                        return !Baritone.settings().mineOnlyInsideSelection.value || isInSelection;
+                        return isInSelection;
                     }
                     return true;
                 })


src/api/java/baritone/api/Settings.java
```diff
@@ -840,6 +840,17 @@ public final class Settings {
...
+
+    /**
+     * Only mine blocks out of the selected area. Can't be combined with {@link #onlyMineInSelection}
+     */
+    public final Setting<Boolean> onlyMineOutOfSelection = new Setting<>(false);
+
+    /**
+     * Only mine blocks in the selected area. Can't be combined with {@link #onlyMineOutOfSelection}
+     */
+    public final Setting<Boolean> onlyMineInSelection = new Setting<>(false);
```
ZacSharp, line 852: I think "out of" should be "oudside" and maybe "in" should then be "inside" for consistency.
Also I think it would be better to name those `mineOnly...` so they are prefixed with the command they apply to and show up next to `mineGoalUpdateInterval`, `mineScanDroppedItems`, `mineDropLoiterDurationMSThanksLouca` and possible futures settings for `#mine` in `#set list`.

src/main/java/baritone/process/MineProcess.java
```diff
@@ -420,6 +422,22 @@ private static List<BlockPos> prune(CalculationContext ctx, List<BlockPos> locs2
                     }
                 })
 
+                .filter(pos -> {
+                    if (Baritone.settings().onlyMineInSelection.value || Baritone.settings().onlyMineOutOfSelection.value) {
+                        boolean isInSelection = Arrays.stream(ctx.baritone.getSelectionManager().getSelections()).anyMatch(selection -> selection.aabb().contains(new Vec3d(pos.getX(), pos.getY(), pos.getZ())));
```
ZacSharp, line 427: This will potentially run many times per tick, not sure how much we can or should try to optimize it.

ZacSharp (commented):
This is super simple and that's a pretty good argument on its own so feel free to ignore the rest of this comment, I'm just making things more complex in some sense.

I have two main concerns
1. This per game run instead of persistent per dimension. People asking for a protection radius around home waypoints likely don't want to select their base every time they join the server.
2. This does not help for `#farm` escaping and breaking things and also doesn't stop `#goto` from making its own "door" a.k.a. hole-in-the-wall

I'd like to have a system that can be used by any command/process, is persistent per dimension and if possible doesn't require two settings per action supporting it.
My current and not yet thought through idea is to have a command to turn selections into named areas which are saved together with other dimension specific data (like waypoints). Since location dependent settings are very unlikely to happen those areas can then be given tags which are used by whatever wants to use them.
For example mine could prune anything inside a region tagged with "pleaseDontMineHere" (name intentionally bad) or have a setting to list the appropriate tags (but then we could just as well not use tags and list the region names instead, maybe that would also allow for better efficiency).

If we are to do this in a persistent way we definitely need a way to distinguish between different functions of selections or the user will have to redo their selection whenever they do something else, which is not much better than the current state.

src/main/java/baritone/process/MineProcess.java
scorbett123: From my tests this running multiple times a tick doesn't really matter as it is quite quick, it would slow the game down slightly, however I can't see an easy optimisation which would be worth the effort.

scorbett123:
Looking at making this work for everything and per waypoint, maybe MovementHelper.avoidBreaking could go through all the waypoints with a tag "don't mine for X blocks from here", and return true if it is in any of those regions. However this is pretty likely to cause more lag.

I think mining only in a selection is something that you probably don't want to be persistent, as #2249 probably doesn't want it to be persistent, but out of selections should probably be persistent, using the waypoint system.

For now I'm going to remove the "onlyMineOutOfSelection" option, however I still think the onlyMineInSelection is a good addition.

src/api/java/baritone/api/Settings.java
```diff
@@ -840,16 +840,10 @@ public final class Settings {
...
-     * Only mine blocks out of the selected area. Can't be combined with {@link #onlyMineInSelection}
-     */
-    public final Setting<Boolean> onlyMineOutOfSelection = new Setting<>(false);
-
     /**
-     * Only mine blocks in the selected area. Can't be combined with {@link #onlyMineOutOfSelection}
+     * Only mine blocks in the selected area.
      */
-    public final Setting<Boolean> onlyMineInSelection = new Setting<>(false);
+    public final Setting<Boolean> mineOnlyOutsideOfSelection = new Setting<>(false);
```
ZacSharp, line 846: I think you made an oopsie here.

scorbett123: Yep, already fixed it, noticed it whilst testing making this setting also work for farm

src/main/java/baritone/process/MineProcess.java
```diff
@@ -420,6 +421,15 @@ private static List<BlockPos> prune(CalculationContext ctx, List<BlockPos> locs2
                     }
                 })
 
+                .filter(pos -> {
+                    if (Baritone.settings().mineOnlyInsideSelection.value) {
+                        boolean isInSelection = Arrays.stream(ctx.baritone.getSelectionManager().getSelections()).anyMatch(selection -> selection.aabb().contains(new Vec3d(pos.getX(), pos.getY(), pos.getZ())));
+
+                        return !Baritone.settings().mineOnlyInsideSelection.value || isInSelection;
```
ZacSharp, line 428:
```suggestion
                        return isInSelection;
```

ZacSharp: It won't ever be false inside this if unless someone changed it on another thread.

scorbett123: Thanks for pointing that out, that's just a remnant from how I was doing it

pizzav-xyz:
when will this be added?

BaoLam-hue:
@ZacSharp please review this, We need this.
You approved it yet not merging it is so weird.

PR #3396, opened 2022-04-16
1.18.2

SundayMC:
Hi, I was wondering if it was possible to add a #gotobiome. Similar to #goto but to go to a biome.


scorbett123:
Why do people keep creating prs instead of issues.

This has also already been suggested

PR #3400, opened 2022-04-17
prevent air from causing issues in acceptable throwaway items.

### LINKED ISSUE #3399: Please add a brief but descriptive title
## Some information
Operating system: Windows
Java version: dunno, not important for this one
Minecraft version: 1.18.2
Baritone version: 1.8.3
Other mods (if used): 
appleskin-fabric-mc1.18.2-2.4.0.jar         architectury-4.1.32.jar
baritone-standalone-fabric-1.8.3.jar        beehive-tooltips-fabric-1.4.2.jar
BetterF3-1.2.3-Fabric-1.18.2.jar            cloth-config-6.2.57-fabric.jar
Clumps-fabric-1.18.2-8.0.0+7.jar            Controlling-fabric-1.18.2-9.0+17.jar
entityculling-fabric-mc1.18-1.5.0.jar       fabric-api-0.48.0+1.18.2.jar
FPS-Monitor-1.18-1.1.0-Fabric.jar           InventoryHUD-fabric-[1.18] -3.4.2.jar
InventorySorter-1.8.6-1.18.jar              iris-mc1.18.2-1.2.2-build.32.jar.disabled
light-overlay-6.0.5.jar                     lithium-fabric-mc1.18.2-0.7.9.jar
malilib-fabric-1.18.2-0.12.1.jar            modmenu-3.1.0.jar
morevillagers-FABRIC-1.18.2-3.2.0.jar       MouseTweaks-fabric-mc1.18-2.22.jar
notenoughcrashes-4.1.4+1.18.2-fabric.jar    phosphor-fabric-mc1.18.x-0.8.1.jar
replaymod-1.18.2-2.6.4.jar                  RoughlyEnoughItems-8.0.442.jar
sodium-fabric-mc1.18.2-0.4.1+build.15.jar   tweakeroo-fabric-1.18.2-0.13.1.jar
wthit-fabric-4.7.3.jar                      XaerosWorldMap_1.20.4.1_Fabric_1.18.2.jar
Xaeros_Minimap_22.3.1.1_Fabric_1.18.2.jar

9000% sure this modlist wont help you though

## Exception, error or logs
Please find your `latest.log` or `debug.log` in this folder and attach it to the issue

The error appears as baritone appears to try to put a block down, but cant seem to. It violently disallows you to move anything, the player, the camera, the active toolbar slot, and the empty slot in your hand. 
I've seen a few patterns it does, One where it tries to place nothing below it, one where it backs up and moves closer and backs up and moves closer while crouching, or just stopping, and staring at the block it needs to place as hard as it can.

## How to reproduce
Add your steps to reproduce the issue/bug experienced here.

if minecraft:air gets added to throwaway items, it tries to PLACE AIR. It decides that you can place air by having an empty inventory slot and right clicking for scaffolding and building and getting places. 

## Modified settings
To get the modified settings run `#modified` in game
acceptableThrowawayItems
allowInventory
blocksToAvoidBreaking
replantNetherWart
useSwordToMine

## Suggested Solution
Disallow air from being selected as a viable block to place, or ignore it's presence in the acceptableThrowawayItems list. 

## Final checklist

wagyourtail:
fixes #3399

### DIFF
--- a/src/main/java/baritone/behavior/InventoryBehavior.java
+++ b/src/main/java/baritone/behavior/InventoryBehavior.java
@@ -25,6 +25,7 @@ import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.init.Blocks;
+import net.minecraft.init.Items;
 import net.minecraft.inventory.ClickType;
 import net.minecraft.item.*;
 import net.minecraft.util.EnumFacing;
@@ -97,7 +98,8 @@ public final class InventoryBehavior extends Behavior {
     private int firstValidThrowaway() { // TODO offhand idk
         NonNullList<ItemStack> invy = ctx.player().inventory.mainInventory;
         for (int i = 0; i < invy.size(); i++) {
-            if (Baritone.settings().acceptableThrowawayItems.value.contains(invy.get(i).getItem())) {
+            Item item = invy.get(i).getItem();
+            if (Baritone.settings().acceptableThrowawayItems.value.contains(item) && item != Items.AIR) {
                 return i;
             }
         }
@@ -129,6 +131,7 @@ public final class InventoryBehavior extends Behavior {
 
     public boolean hasGenericThrowaway() {
         for (Item item : Baritone.settings().acceptableThrowawayItems.value) {
+            if (item == Items.AIR) continue;
             if (throwaway(false, stack -> item.equals(stack.getItem()))) {
                 return true;
             }
@@ -145,6 +148,7 @@ public final class InventoryBehavior extends Behavior {
             return true;
         }
         for (Item item : Baritone.settings().acceptableThrowawayItems.value) {
+            if (item == Items.AIR) continue;
             if (throwaway(select, stack -> item.equals(stack.getItem()))) {
                 return true;
             }


scorbett123:
Is it possible to expand this to all items which can't be walked on?

XorOwl:
> Is it possible to expand this to all items which can't be walked on?

Looking around for some candidates, 
Spider web and string are two obvious ones.
I dont know how it handles things, but paintings, signs and item frames are worth considering,
Maybe carpet.

wagyourtail:
should I just check if the block it would place is canWalkOn somehow (should work by passing the usual args except blockstate, probably getting like default state or whatever via that function in builderprocess)

ZacSharp:
I'd say `canWalkOn` with a special case for `isWater` and `isLava` blocks (because those use the coordinates) is a good solution. If you find a way to do it without passing dummy arumgents that's even better.

wagyourtail:
I did a thing, does all now. does that look good?

XorOwl:
> I did a thing, does all now. does that look good?

Looks good, but question, 
You removed 
import baritone.api.BaritoneAPI;
from InventoryBehavior.java
was that intentional? looks like it might have been an oopsie, but i'm not familiar enough with baritone's api to know for sure.

wagyourtail:
my ide isn't showing any errors and it still builds, so the import was unused

src/main/java/baritone/behavior/InventoryBehavior.java
```diff
@@ -97,7 +97,8 @@ private void swapWithHotBar(int inInventory, int inHotbar) {
     private int firstValidThrowaway() { // TODO offhand idk
         NonNullList<ItemStack> invy = ctx.player().inventory.mainInventory;
         for (int i = 0; i < invy.size(); i++) {
-            if (Baritone.settings().acceptableThrowawayItems.value.contains(invy.get(i).getItem())) {
+            Item item = invy.get(i).getItem();
+            if (MovementHelper.filterThrowawayForPlacing(Baritone.settings().acceptableThrowawayItems.value).contains(item)) {
```
scorbett123, line 101: Probably better for this to be run once at the beginning of the function instead of for every item in the inventory.

scorbett123 (changes requested):
I think this is much better, just a small performance thing.

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -344,6 +347,25 @@ static boolean canWalkOn(BlockStateInterface bsi, int x, int y, int z, IBlockSta
...
+                    // plus magma, which is a normal cube but it hurts you
+                    return false;
+                }
+                IBlockState state = block.getDefaultState();
+                if (state != null && state.isBlockNormalCube()) {
+                    return true;
+                }
+                return block == Blocks.FARMLAND || block == Blocks.GRASS_PATH ||
+                       block == Blocks.ENDER_CHEST || block == Blocks.CHEST || block == Blocks.TRAPPED_CHEST ||
+                       block == Blocks.GLASS || block == Blocks.STAINED_GLASS;
```
ZacSharp, line 365: This should also return true for slabs if `allowWalkOnBottomSlab` is enabled.

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -344,6 +347,25 @@ static boolean canWalkOn(BlockStateInterface bsi, int x, int y, int z, IBlockSta
         return block instanceof BlockStairs;
     }
 
+    static List<Item> filterThrowawayForPlacing(List<Item> items) {
```
ZacSharp, line 350: I think a comment here (and probably on `MovementHelper::canWalkOn` as well) would be a good idea to make sure people know these two methods  should be roughly in sync (precisely: this method may only return items where `MovementHelper::canWalkOn` is guaranteed to return `true` for all blockstates the item can place).

ZacSharp, line 350: Maybe this should be split into `filterThrowawayForPlacing` and an overload of `canWalkOn` (taking only a blockstate) so if we need to check single blocks we don't face the same problem once again.

PR #3404, opened 2022-04-18
Don't crash when worldData is null, and get0 is called in 1.18

scorbett123:
Changes in d22a52976b2988a0daa7ec8324de32ab2c095d0c mean that BlockStateInterface.get0, crashes when world data is null.

World data is rarely null, however one case I have found it being null is when launching mc with the --server option.


wagyourtail:
I think #3255 already fixed this in a different / better way on 1.17 and just needs to be merged up...

scorbett123:
I forgot that 1.17 also had changeable block heights so I didn't check for a fix on that branch. My gradle is really messed up so I can't test if that fixed it, I'll close this for now and come back to this if I test it in the future.

PR #3479, opened 2022-05-31
Caching performance upgrade

scorbett123:
Precomputing data in pathfinding leads to performance improvements.

This takes on average 7 - 10 ms to run initially, however then means that there is faster pathing throughout.

I've mainly focussed on changing occurrences during pathfinding, as MovementHelper.canWalkThrough being run every tick doesn't have any performance hit. A couple things I don't like about this though is that canWalkThrough is implemented again, with slightly different return values, so any change made to one, has to be also made to the other.

This is really dodgy in places, and is just an initial draft, which I'll later improve. I'd also appreciate it if other people can test it and tell me if there is any crashes / bugs.


src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -456,6 +460,11 @@ public BetterBlockPos pathStart() { // TODO move to a helper or util class
         return feet;
     }
 
+    @Override
+    public void onSettingChanged(SettingChangedEvent event) {
+        this.precomputedData.refresh();
```
leijurv, line 465: I think we'd want this to be more like `this.precomputedData = refreshed()`. Because we don't want to modify the precomputed data being used by the pathing thread, I'd worry about concurrent modification exception.

src/main/java/baritone/pathing/movement/movements/MovementAscend.java
```diff
@@ -68,7 +68,7 @@ protected Set<BetterBlockPos> calculateValidPositions() {
     public static double cost(CalculationContext context, int x, int y, int z, int destX, int destZ) {
         IBlockState toPlace = context.get(destX, y, destZ);
         double additionalPlacementCost = 0;
-        if (!MovementHelper.canWalkOn(context.bsi, destX, y, destZ, toPlace)) {
+        if (!context.precomputedData.canWalkOn(context.bsi, x, y, z, toPlace)) {
```
leijurv, line 71: Changing this from `destX` to `x` doesn't look right

src/main/java/baritone/behavior/PathingBehavior.java
leijurv: Ok, actually all that's happening is modifying a boolean array, which can't crash due to concurrent modification, but still it's best practice to consistently separate objects modified on main game tick thread from objects used on pathing thread.

leijurv (commented):
Overall good stuff, other than a few nits I only have one issue: we definitely can and should have a single implementation of canWalkOn cached and canWalkOn special case. Those should probably be moved from anonymous lambda back into static methods of MovementHelper. Then, the non cached canWalkOn can call the two variants written for caching. It'll be a bit slower to construct an Optional and such, but no big deal since it's just a handful of calls per tick. This is really important to only have one implementation of what's canWalkOn, so I think it's definitely worth it.

And the overhead of constructing an optional could be removed actually, let's just do `private static final Optional<Boolean> TRUE = Optional.of(Boolean.TRUE);` and then return that each time, instead of `Optional.of(true)` which allocates a new optional each time.

src/main/java/baritone/pathing/movement/movements/MovementAscend.java
scorbett123: It's not right, thanks for pointing this out.

src/api/java/baritone/api/Settings.java
```diff
@@ -1302,6 +1303,13 @@ public final T get() {
             return value;
         }
 
+        public void set(T value) {
+            this.value = value;
+            if (BaritoneAPI.getProvider() != null) {
+                BaritoneAPI.getProvider().getAllBaritones().forEach(iBaritone -> iBaritone.getGameEventHandler().onSettingChanged(new SettingChangedEvent(this)));
+            }
+        }
+
```
ZacSharp, line 1312: Not sure about this one.
Simply relying on this method being called means silently breaking compatibility with every single Baritone GUI, causing wrong behavior (ignoring or partially ignoring the changed value) whenever such a GUI is used.
Since we already have an api where others may silently change the settings I think it would be better to periodically check for changes. We can still provide this method and deprecate not using it though (and we can drop the check when this was called since in that case the cache is invalidated anyway).

wagyourtail:
so, what's the status on this, I kinda want to include this (or #3375 until this is ready) for 1.19

scorbett123:
> so, what's the status on this, I kinda want to include this (or #3375 until this is ready) for 1.19

Currently I've got exams, so don't have the same amount of time I usually do, I should be able to get this finished by the end of the week. It should only be a couple small changes moving some functions to movement helper, and fixing @ZacSharp's suggestion.

scorbett123:
This should be Leijurv's requested changes. Does this look better to you?

src/main/java/baritone/behavior/PathingBehavior.java
```diff
@@ -99,6 +103,11 @@ public void onTick(TickEvent event) {
             baritone.getPathingControlManager().cancelEverything();
             return;
         }
+
+        if (ticksElapsedSoFar % 200 == 0) {
+            precomputedData = new PrecomputedData(); // This is here for now in case settings aren't changed in normal ways, should mean it is updated whatever once every 10 seconds
+        }
```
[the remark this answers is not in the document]
ZacSharp: Yes, that's about what I meant.
My main concern about this is settings being outdated for a long enough time for users to notice it and we also can't speed this up to every tick like I though when writing my previous comment because now that I looked at this pr again I noticed `PrecomputedDataForBlockState` eagerly computes all data on construction, meaning any approach like this results in a repeated waste of cpu time. The two solutions I see are
1. making `PrecomputedDataForBlockState` lazy.
This would probably require an additional `boolean[]` and an additional check, resulting in more memory and time consumption on the pathing thread.
2. Not reconstructing `PrecomputedData` on a regular schedule.
Since reconstructing it every 10 seconds already makes it very likely that it is not reused for a second path, why not stop trying to do so and reconstruct it once per path?
Alternatively we could cache the (relevant) setting values and only reconstruct `PrecomputedData` if there was an actual change. This is the option I had in mind but thought it's too much effort for no gain over the simpler solution you implemented. This can also be combined with 2.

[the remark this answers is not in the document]
scorbett123: I was initially going to do it whenever a new CalculationContext is created, however it seems a new one is created every tick, so where would you propose that we update this?

src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.init.Blocks;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.Optional;
+
+import static baritone.pathing.movement.MovementHelper.isFlowing;
+import static baritone.pathing.movement.MovementHelper.isWater;
+
+public class PrecomputedData { // TODO add isFullyPassable
```
leijurv, line 35: this is super cute and i like it a lot!!

src/main/java/baritone/process/MineProcess.java
```diff
@@ -111,7 +111,7 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
         int mineGoalUpdateInterval = Baritone.settings().mineGoalUpdateInterval.value;
         List<BlockPos> curr = new ArrayList<>(knownOreLocations);
         if (mineGoalUpdateInterval != 0 && tickCount++ % mineGoalUpdateInterval == 0) { // big brain
-            CalculationContext context = new CalculationContext(baritone, true);
+            CalculationContext context = new CalculationContext(baritone, true, null); // Precomputed data should never be used on this calculation context
```
leijurv, line 114: Actually `rescan` calls `searchWorld` which calls `prune` which calls `plausibleToBreak` which calls `getMiningDurationTicks`, which SHOULD call the precomputed `canWalkThrough` (but doesn't currently)

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -88,61 +88,105 @@ static boolean canWalkThrough(BlockStateInterface bsi, int x, int y, int z) {
...
+
+        if (block instanceof BlockCauldron) {
+            return Optional.of(false);
+        }
+
+        try { // A dodgy catch-all at the end, for most blocks with default behaviour this will work, however where blocks are special this will error out, and we can handle it when we have this information
+            return Optional.of(block.isPassable(null, null));
+        } catch (NullPointerException exception) {
+            System.out.println("Error");
+            return Optional.empty();
```
leijurv, line 155: It took me a second to realize why this is correct - let's make this comment a bit more explanatory. If I'm understanding right, the idea is that `block.isPassable(bsi.access, bsi.isPassableBlockPos.setPos(x, y, z));` can only be done in special case, not in precomputed, so if we default to `Optional.empty()` then we're good because the special case will call it correctly?

leijurv: Also just out of kindness to other mods that will lose their minds when passing null data into `block.isPassable`, perhaps this should be a general `catch Throwable`?

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -88,61 +88,105 @@ static boolean canWalkThrough(BlockStateInterface bsi, int x, int y, int z) {
...
+    }
+
+    static boolean canWalkThrough(CalculationContext context, int x, int y, int z) {
+        return context.precomputedData.canWalkThrough(context.bsi, x, y, z, context.get(x, y, z));
+    }
+
+
     static boolean canWalkThrough(BlockStateInterface bsi, int x, int y, int z, IBlockState state) {
+        Optional<Boolean> canWalkOn = canWalkThroughBlockState(state);
+        return canWalkOn.orElseGet(() -> canWalkThroughPosition(bsi, x, y, z, state));
```
leijurv, line 102: Let's avoid `orElseGet` in this case. It requires constructing that lambda `() -> canWalkThroughPosition(bsi, x, y, z, state)` which allocates on the heap because it's impure (meaning that it captures those variables bsi x y z state and allocates with them). Let's just do the old way (an if statement on `isPresent`).

leijurv:
Let's make sure we get all the cases that happen on the pathing thread.

https://github.com/cabaletta/baritone/blob/868c023dbdd91012a074002ead135eee78d8c6d3/src/main/java/baritone/pathing/movement/MovementHelper.java#L460

Here's one example ^, this would be dreadfully nonperformant because it allocates an `Optional<Boolean>` as well as the `orElseGet` lambda, every time anything calls `getMiningDurationTicks`, which happens all the time in movement cost calculation.

Perhaps this could be tested by temporarily adding a `if (!Minecraft.isCallingFromMainGameThread()) throw new IllegalStateException();` to the precomputed calculator (the one that makes optionals) and seeing if anything trips that?

src/main/java/baritone/behavior/PathingBehavior.java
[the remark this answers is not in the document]
leijurv: ZacSharp makes a good point, I didn't conisder that: https://github.com/cabaletta/baritone/pull/3479#discussion_r890597059 

Walking through all BlockStates already takes a nontrivial amount of time, there are thousands of them, scorbett mentioned in the OP that precomputing all the data takes 7 to 10ms. That's very nontrivial, if we did that on the main game thread it could cause noticable stutter. For example, at 60fps, ticks can't take longer than 16.666ms otherwise the framerate will drop. With just a few mods adding new blockstates at 60fps, or with a higher than 60fps monitor with vanilla, this could cause notable performance issues. This would definitely not be desirable on the main thread.

So I think that, for sure, we **should not** precompute all block states on the main tick thread, at any time interval (whether it's once a tick, or once every 200 ticks).

Precomputing every time a setting is modified would be nice, but, as ZacSharp points out, other mods and clients can mess with Baritone's settings at any time, and we don't necessarily get a callback.


There are a whole bunch of options here with a lot of tradeoffs, here are a few:

- Fully lazy `PrecomputedDataForBlockState`. This would be relatively simple, but it would add an extra branch to every lookup, which is exactly what this is trying to avoid. This is a perfectly reasonable option and it would still be much better than before. Note that `ToolSet` already does exactly this with its HashMap.
- Construct the precomputed data on the pathing thread, first thing before it starts A*. This would work and it wouldn't make the main game stutter. It would be less than ideal from a thread safety pov, but, this is what Baritone does today (accessing settings from pathing thread), so, nothing new. We have a 50ms step function for pathing, which is that the pathing thread's completed path can't be presented to the game until the beginning of the next tick when `PathingBehavior` checks for it. That means that it's more or less "fine" to add a short delay to its startup. For example, even for tiny paths such as when using `MineProcess`, this wouldn't change anything, it would still be a 1 tick delay, it's just that instead of 5ms of the 50ms budget being used, we now use 15ms or something. However, this could still pose a problem with modpacks that add more blockstates, or, future versions of minecraft that do the same. Also not everyone has a fast computer.
- Some kind of godawful reflect to check when a Setting value has changed. Some kind of horribleness with `System.identityHashCode` should work? This could be made a little better by only checking the settings that are actually referenced by `canWalkThrough` and `canWalkOn` maybe? For the `Setting<List<?>>` ones, maybe we could make the `List` be a custom `TattleTaleList` that intercepts calls to `add` and `remove` and tattletales by telling the precomputed data to recompute itself? This is probably more trouble than it's worth.
- Precompute the data on a regular schedule, but on a totally separate thread from both main game and pathing? e.g. just `while (true) { Baritone.precomputedData = compute(); Thread.sleep(10000); }`
- Some kind of hybrid? Perhaps we construct a lazy precomputeddata every tick for the Movements to use, but pathing uses a fully computed one? Perhaps we check how big `Block.BLOCK_STATE_IDS.size()` is, and go lazy if it's above 10k otherwise precompute? Perhaps we benchmark how long precomputing takes, and do lazy if it's above 15ms, and precomputed on pathing thread if it's below 15ms?

leijurv:
It also would just be really fun if instead of four boolean arrays for `canWalkOn -> specialCases`, `canWalkOn -> dataPerBlockState`, `canWalkThrough -> specialCases`, `canWalkThrough -> dataPerBlockState`, we instead just had a giant `int[]` (or even `long[]`) `allDataForAllBlockStates` and each bit represented one of those booleans. It would also probably be faster because it'd be way more cache friendly.

src/main/java/baritone/behavior/PathingBehavior.java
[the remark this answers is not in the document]
scorbett123: Which of these would you prefer I do. I would go for the simplest, just starting a separate thread, would this be alright, or do  you want me to go for one of these more complicated solutions?

src/main/java/baritone/pathing/movement/MovementHelper.java
scorbett123: Ok, I can switch it to a throwable. The idea is that for most common blocks it won't crash when doing this, most don't override the default `return !this.material.blocksMovement();` however some do, and in those cases they probably use context from the world, which means we have to calculate them as a special case.

src/main/java/baritone/pathing/movement/MovementHelper.java
scorbett123: Sorry yeah, I had it like that originally, then I received complaints from intellij so thought this was what you are meant to do.

scorbett123:
> It also would just be really fun if instead of four boolean arrays for `canWalkOn -> specialCases`, `canWalkOn -> dataPerBlockState`, `canWalkThrough -> specialCases`, `canWalkThrough -> dataPerBlockState`, we instead just had a giant `int[]` (or even `long[]`) `allDataForAllBlockStates` and each bit represented one of those booleans. It would also probably be faster because it'd be way more cache friendly.

If someone else wants to do this then fine, but I don't even know how I'd get started doing something like this in a way with decent performance.

wagyourtail:
```
NORMAL_CASE = 0xb1
SPECIAL_CASE_1 = 0xb10
SPECIAL_CASE _2 = 0xb100
```
assembling each
```
if (stuff for special case) value[blockstateid] |= SPECIAL_CASE_1
```
etc

and then I think java bytecode just does if statements on number that's not zero so
`value[blockstateid] & currentFlags != 0` would drop the neq 0 part.

src/main/java/baritone/behavior/PathingBehavior.java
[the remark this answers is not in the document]
ZacSharp: I'd really like detecting changes (by pointer equality, IMO setting values should be considered immutable) because it allows reusing data for multiple paths instead of recaching every time even though normally nothing changes. However I can understand if that's too much effort and in that case vote for a lazy cache using a byte/short/int/long per state because that should be comparable to an eager cache (also with a single value per state) in performance but without the setup overhead (literally just one extra branch on the special case path, no change on the fast path).
From my thoughts a hybrid lazy + non-lazy approach might ruin inlining and perform worse in all cases, so unless you show a benchmark proving that's wrong I'll vote against it.

ZacSharp:
The bitpacking also allows checking two properties at once so if you set bit 0 indicate uninitialized values, bit 1 to indicate special cases and bit 3 to indicate the return value, if initialized and not special, then `statemask & 0x11 == 0` is true only for nonspecial initialized blocks, meaning in that case you can return bit 3 without any further checks.

ZacSharp:
> I thought it was many blocks in a single int / long

Which would require 10 or 20 less memory for ints and longs respectively. Not sure about speed though.
Minecraft does something like this to store chunk contents.

src/main/java/baritone/behavior/PathingBehavior.java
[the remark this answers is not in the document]
leijurv: I think the first one is good for now. It avoids all issues of threading, precomputing, scheduling, all of that. Simply put in a `boolean[] blockStateHasBeenEvaluatedYet` or something. At the top of every query, check if it's false, if so we compute all the data (canWalkOn, canWalkThrough, etc) for this blockstate. It'll be fast because that check will almost always do nothing, so the branch predictor will have fun.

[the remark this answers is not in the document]
leijurv: It also means that `SettingChangedEvent` is not needed, but maybe we should leave it in, just in case @ZacSharp wants to write a setting changed detector? 😉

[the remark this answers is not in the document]
leijurv: Or I guess, for simplicity, it would be okay if the laziness was per-attribute. i.e. `canWalkOn(Blocks.STONE)` is cached unrelatedly to `canWalkThrough(Blocks.STONE)`. That way it could still be separate `PrecomputedDataForBlockState` objects.

leijurv:
Also let's do https://github.com/cabaletta/baritone/pull/3479#pullrequestreview-991101180 with the `Optional<Boolean>`, it bugs me to construct new `Optional`s for no reason. 

```
private static final Optional<Boolean> TRUE = Optional.of(true);
private static final Optional<Boolean> FALSE = Optional.of(false);
```

👍

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -293,55 +339,78 @@ static boolean avoidWalkingInto(Block block) {
...
                 return isWater(up) && !Baritone.settings().assumeWalkOnWater.value;
             }
             // if assumeWalkOnWater is on, we can only walk on water if there isn't water above it
             // if assumeWalkOnWater is off, we can only walk on water if there is water above it
             return isWater(up) ^ Baritone.settings().assumeWalkOnWater.value;
         }
-        if (Baritone.settings().assumeWalkOnLava.value && isLava(block) && !isFlowing(x, y, z, state, bsi)) {
-            return true;
-        }
-        if (block == Blocks.GLASS || block == Blocks.STAINED_GLASS) {
```
leijurv, line 332: This check for glass might have slipped through the cracks? Should be added back to the cached variant.

leijurv:
Oh God, will this work in 1.13+? Or will IFluidState mess this up??

ZacSharp:
Fluids are still tied to blocks pretty tightly with all waterloggable blocks having a `waterlogged` property so I think we can get away with at most a handful of new special cases (for those where the flowing state matters e.g. waterlogged open bottom trapdors in `canWalkThrough`).

src/main/java/baritone/pathing/movement/MovementHelper.java
scorbett123: It did slip through the cracks, thanks for pointing this out

scorbett123:
Is that last commit the sort of thing you're looking for @leijurv? I still need to clean it up, but no point doing that if I'm going to need to significantly change it.

src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public boolean canWalkThrough(BlockStateInterface bsi, int x, int y, int z, IBlockState state) {
+        int id = Block.BLOCK_STATE_IDS.get(state);
+
+        if ((data[id] & completedMask) == 0) { // we need to fill in the data
+            fillData(id, state);
+        }
+
+        if ((data[id] & canWalkThroughSpecialMask) != 0) {
+            return MovementHelper.canWalkOnPosition(bsi, x, y, z, state);
```
ZacSharp, line 95:
```suggestion
            return MovementHelper.canWalkThroughPosition(bsi, x, y, z, state);
```

ZacSharp (changes requested):
Concept wise this looks clean now and with the `canWalkThrough` fix applied everything seemed to work.

Codacy is complaining about a dozen unused imports.

leijurv:
Good stuff!

I think we shouldn't add precomputeddata as a constructor parameter to calcluationcontext anymore. It seems like there's essentially no downside to just making a new precomputeddata each time. Sharing precomputeddata across threads is a no-go, now that it's mutable. 

Broadly, does this mean we don't actually need the setting modified event anymore?

src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -0,0 +1,100 @@
...
+            if (canWalkThroughState.get()) {
+                data[id] = data[id] | canWalkThroughMask;
+            }
+        } else {
+            data[id] = data[id] | canWalkThroughSpecialMask;
+        }
+
+
+        data[id] = data[id] | completedMask;
+    }
```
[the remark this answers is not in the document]
leijurv: Actually I think it's okay because the bitwise OR is idempotent. If two threads race to complete this, it'll never be in an incoherent/invalid state.

But this is just a happy accident, we should not rely on this.

leijurv:
ok i pushed a commit to your branch (i didnt know this was possible until today)

leijurv:
```
        if (block instanceof BlockLiquid) {
            if (state.getValue(BlockLiquid.LEVEL) != 0) {
                return NO;
```

Why?

leijurv:
Ok this feels roughly about right.

I'd still like a 2nd/3rd/4th pair of eyes over MovementHelper's before/after to make sure it's about right. Also some testing would be good. Also benchmarking would be epic sauce.

scorbett123:
What else is needed to be done on this to make this ready for merging @leijurv?

leijurv:
@scorbett123 basically what I said here https://github.com/cabaletta/baritone/pull/3479#issuecomment-1185180741

i'd like if someone (maybe u could beg and plead in #dev-help lol) would closely look at the before/after in MovementHelper to make sure all the special cases were respected and none slipped through the cracks

a benchmark would be nice but not strictly needed

leijurv:
ok whatever ill just merge it

PR #3458, opened 2022-05-17
Merge 1.16.5 into 1.17.1

ZacSharp:
Please don't merge immediately before or after updating 1.16.5 as I'd like to re-merge in that case.

PR #3488, opened 2022-06-04
Merge 1.17.1 into 1.18.2

ZacSharp:
Note that this does not contain changes from #3458 
Please don't merge immediately before or after updating 1.17.1 as I'd like to re-merge in that case.

Do you want updates for 1.18.1?


ZacSharp:
Checks failed because I accidentally made this pr target master and then changed the base branch while they were running.

wagyourtail:
you could try like force pushing or something to get the checks to rerun

ZacSharp:
Okay that definitely did not work as expected.
Now I have a pr desynced from it's origin branch.

ZacSharp:
Now I can't even reopen it because there's no new commits, no matter what's actually pushed to my repo.

ZacSharp:
I don't get it.

ZacSharp:
Seems to be all fine now so I'll leave it as is.
I think the problem was that I pushed a commit that's already in 1.18.2.

PR #3533, opened 2022-06-27
Fix for a sprint bug that can cause death

Echocage:
In the case where we're descending stairs and need to stop at the bottom to mine a block to our left or right. If the next space is lava or a hole, we'll currently overshoot right into it even with default settings.

Currently 

```
        if (next instanceof MovementDescend && next.getDirection().equals(current.getDirection())) {
            return true;
        }
```
is checked first inside the canSprintFromDescendInto, which will return True before MovementHelper.canWalkOn is checked. 

Checking MovementHelper.canWalkOn first ensures we won't sprint onto a block where the next block is not walkable.

### DIFF
--- a/src/main/java/baritone/pathing/path/PathExecutor.java
+++ b/src/main/java/baritone/pathing/path/PathExecutor.java
@@ -73,8 +73,8 @@ public class PathExecutor implements IPathExecutor, Helper {
     private HashSet<BlockPos> toPlace = new HashSet<>();
     private HashSet<BlockPos> toWalkInto = new HashSet<>();
 
-    private PathingBehavior behavior;
-    private IPlayerContext ctx;
+    private final PathingBehavior behavior;
+    private final IPlayerContext ctx;
 
     private boolean sprintNextTick;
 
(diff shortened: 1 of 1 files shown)


Echocage:
Video example of bug: https://www.youtube.com/watch?v=0UFFM5a9Bh0

leijurv:
From my initial read, it seems like this change would disable sprinting through a long series of descends? It seems like the first check would now return false because if you have several descends in a row, the destination of the next one would be air, which isn't canWalkOn.

ZacSharp:
The check that was moved down can only return true when the next movement is a `MovementDescend` in the same direction so  it's impossible to overshoot without ending up in the next movement, which guarantees us a block to stand on (unless there's a bug in pathing code). The reason why the fix works nonetheless is that, as leijurv said, it almost completely breaks sprinting in `MovementDescend`s and if you don't sprint at all you can't spring too much.

Echocage:
@leijurv @ZacSharp You guys are totally right. 

So the issue I've found is that not sprinting 1 block before the block next to the dropoff is not enough, we still fall off due to the momentum, which is why the code in master doesn't work.

I've removed the if swap that I had done originally, and instead we now check both the next block, and the block after that if we can sprint onto it, only then do we allow descending sprinting. What do you guys think of this solution?

leijurv:
That definitely also doesn't feel right. The reason is that you're calling `canSprintFromDescendInto(ctx, next, ...)` without checking if `next` actually is a `MovementDescend` or not. I think it would be better to narrow down very specifically the case that fails, then add a new `if (blah) { return false; }` to `canSprintFromDescendInto` to cover that specific problem case.

PR #3539, opened 2022-07-03
Use immutable collections for settings

ZacSharp:
Using an `ImmutableSet` seems to be faster according to [this](https://github.com/ZacSharp/baritone/blob/benchmarks/immutableSettingValues/results.csv) benchmark, but that would either require a breaking API change or improperly making the settings `ImmutableSet`s while exposing only their list views (i.e. lists that behave as ordered sets, which currently works because we only use them as ordered sets).


leijurv:
Seems like this would really help with #3479. Are there any downsides?

Also I don't care about breaking API changes as long as impact client isn't affected 😈 (joke) (but only sorta)

ZacSharp:
Is any of those scenarios or something else holding this back?

PR #3540, opened 2022-07-05
1.19/1.19

BlueberryDuck:
hijack wagyoutail's 1.19 working build with my walkable magma block
I have no idea wether this works


BlueberryDuck:
didn't work as I thought
please ignore, sorry to bother

scorbett123:
uhm, no? This seems like it just means that baritone will walk on magma, doing so will hurt the player, which isn't wanted in baritone?

Also this pr is just a mess, you suggest you haven't even tested it. You need to make sure to test features before making a pull request, as otherwise it just wastes both of our times as we need to give feedback to you, and then you need to fix it, with possibly not the best description.

BlueberryDuck:
Sorry, I thought I could try and somehow use the Java CI with gradle Action for my own private build by opening this pr, similar to what happened here (https://github.com/cabaletta/baritone/actions/runs/2462884513)
Wasn't intended to actually be reviewed and merged

I only have very mediocre github skills :/

(And regarding the hurt player problem, could be avoided by having frost walker, that's why I was trying to implement it)

wagyourtail:
it should've started building on the actions of your fork before you even made the pr

PR #3562, opened 2022-07-13
Fix: set default minYLevelWhileMining to -64, add maxYLevelWhileMining

LayneTheEpic:
- Lowest Y level is now -64, so this is now the default value of `minYLevelWhileMining`
- Added maxYLevelWhileMining (this helps with mining things like ancient debris)


LayneTheEpic:
wtf did I do I'm so sorry

PR #3579, opened 2022-07-25
Stop backfill from filling while paused

lucasarden:
It's not perfect; while paused, it will still attempt to fill in any blocks that were added to blocksToReplace while running. So for example, if you pause it, then break a block that was previously backfilled, it will instantly replace it. For any other block however, it will not fill it.


ZacSharp:
If we remove blocks from `blocksToReplace` if they are not air (before calling `amIBreakingABlockHMMMMMMM` so the currently targeted block is added back) it will also stop messing around with backfilled blocks the user broke again.

lucasarden:
I've added that check!

PR #3613, opened 2022-08-11
1.19.1

Yanninay:
Add the ability to mine separately specified blocks using tools with a silk touch. There are servers where diamond ore is used as currency (for the purchase and sale of various goods), so the bot will need to mine it only using silk touch, otherwise it will get diamonds, and not the desired ore for the user


scorbett123:
Why do people keep doing this.

PR #3640, opened 2022-08-30
Fix for infinite loop in water

### LINKED ISSUE #3200: Baritone stuck in infinite loop while falling into water more than 1 block deep
## Some information
Operating system: Windows 10
Java version: 17
Minecraft version: 1.16.5
Baritone version: Impact 4.9.1
Other mods (if used): 

## How to reproduce
Make Baritone fall into water more than 1 block deep. You will notice that it constantly keeps trying to correct itself, but fails

https://user-images.githubusercontent.com/97650056/149560280-49585ba5-064e-4d5e-84f2-968971b4addd.mp4

## Modified settings
allowInventory, breakFromAbove, chatControl, desktopNotification. followRadius, freeLook, legitMine, maxFallHeightNoWater, randomLooking, randomLooking113, renderPath

## Final checklist

scorbett123:
This is a slightly dodgy solution, however it is the simplest one I could think of.

# Description of issue

When a fall into more than 2 block deep water occurs baritone enters an infinite loop as we swim up, to the surface, and end the movement when we "jump" above the water, this means that our position isn't in the valid positions of a traverse movement. And thus some of the lag protection jumps in and resets us to the fall (as this is a valid position for during the fall), entering an infinite loop.

Closes #3200.

If anyone can think of a better solution to this please let me know

### DIFF
--- a/src/main/java/baritone/pathing/movement/Movement.java
+++ b/src/main/java/baritone/pathing/movement/Movement.java
@@ -123,7 +123,7 @@ public abstract class Movement implements IMovement, MovementHelper {
     public MovementStatus update() {
         ctx.player().getAbilities().flying = false;
         currentState = updateState(currentState);
-        if (MovementHelper.isLiquid(ctx, ctx.playerFeet())) {
+        if (MovementHelper.isLiquid(ctx, ctx.playerFeet()) && ctx.player().position().y < dest.y + 0.6) {
             currentState.setInput(Input.JUMP, true);
         }
         if (ctx.player().isInWall()) {
--- a/src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
@@ -70,7 +70,7 @@ public class MovementTraverse extends Movement {
 
     @Override
     protected Set<BetterBlockPos> calculateValidPositions() {
-        return ImmutableSet.of(src, src.above(), dest); // src.above means that we don't get caught in an infinite loop in water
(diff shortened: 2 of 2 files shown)


wagyourtail (approved):

wagyourtail:
would some kind of water check make this less dodgy?

scorbett123:
> would some kind of isWater check make this less dodgy?

It would but that would complicate things and I can't really think of a situation in which this will really do any harm, if someone can think of one then then I'll add a check, but otherwise I think it is just unnecessary complication.

ZacSharp:
If Baritone mines straight down and then traverses (tunneling under a lava lake?) but the server lags right when the `MovementTraverse` starts, setting Baritone back to the beginning of the last `MovementDownward` before the block was mined, this will cause the path to fail after 5-10s instead of fixing it instantly.
Treating a position A as valid because the position B below it is valid should only be allowed if B is not `canWalkOn` (or even better, only if the player is actually unsupported).

scorbett123:
This now should instead make baritone only swim up to 0.6 of a way through a block, so instead of jumping out above the water we just swim along through it.

wagyourtail:
shouldn't this have targeted 1.16.5?

scorbett123:
> shouldn't this have targeted 1.16.5?

Do we still support 1.16.5? If so then probably, but thought we'd decided to deprecate it.

wagyourtail:
Iirc it got un deprecated

scorbett123:
Oh, ok, I missed that, this could probably just go on master so I'll make a pr for that later.

PR #3642, opened 2022-09-01
remove schematica command because it doesn't work anyway on 1.13+ 

wagyourtail:
remove schematica command because it doesn't work anyway on 1.13+ 
since sechematica doesn't exist past 1.12.2


ZacSharp (approved):
Can alwayays uncomment the line in case some mod implements schematica API or we repurpose the commant.

wagyourtail:
> Can alwayays uncomment the line in case some mod implements schematica API or we repurpose the commant.

yeah, that's why I didn't delete any code

PR #3669, opened 2022-09-24
Rework of @Emerson 1.12 litematica port for 1.18.2 fabric

rycbar0:
1.18.2 Update of Emersons 1.12.2 litematica support

no bug fixes just update
can build litematic schematics
fails with sertain blocks:
- wall_torche
- blocks that have multiple states like levers (powered on/off), repeaters(delay 1-4), etc. and are not in there base state
- waterloged blocks

strugles with blocks that have a orientation like stairs, chests, hoppers, etc
they can be placed but only if you  already standing at the right spot to place it

### DIFF
--- a/src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
+++ b/src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
@@ -19,6 +19,7 @@ package baritone.utils.schematic.format;
 
 import baritone.api.schematic.IStaticSchematic;
 import baritone.api.schematic.format.ISchematicFormat;
+import baritone.utils.schematic.format.defaults.LitematicaSchematic;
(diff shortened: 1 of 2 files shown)


ZacSharp:
Pullrequests should always target the lowest applicable branch, which is master in this case.
Also there is no value in duplicating a pr without applying any required changes to it.

ZacSharp:
Duplicate of #2544

rycbar0:
if his verison works why isnt it in merged? also as i understad it his version only supports 1.12 and mine targets 1.18.

ZacSharp:
It is not ready to be merged, just like your copy pasted pr.

Emersons pr is for 1.12.2 because new features for Baritone are always implemented on master first and then merged up into the other branches. Prs on branches for newer versions are only merged if it does not make to apply them to master (e.g. fixing bugs that were introduced in an mc update).

PR #3672, opened 2022-09-26
Litematica Support for bariton

rycbar0:
This implements Litematica support for Bariton and fixes the in #2544 mentioned issus.
i did not work on the previusly mentioned branch as it is abandoned for over 1.5 years now.


rycbar0:
@ZacSharp does this pull request meet the required standards?

src/launch/resources/mixins.baritone.json
```diff
@@ -31,6 +31,7 @@
     "MixinStateImplementation",
     "MixinTabCompleter",
     "MixinVboRenderList",
-    "MixinWorldClient"
+    "MixinWorldClient",
+    "MixinNBTTagLongArray"
```
leijurv, line 35: alphabetical order pls

src/launch/java/baritone/launch/mixins/MixinNBTTagLongArray.java
```diff
@@ -0,0 +1,37 @@
...
+ * @author rycbar
+ * @since 26.09.2022
+ */
+@Mixin(NBTTagLongArray.class)
+public abstract class MixinNBTTagLongArray implements INBTTagLongArray {
+
+    @Accessor("data")
+    @Override
+    public abstract long[] getLongArray();
+
```
leijurv, line 35: blank lines

src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
```diff
@@ -65,6 +66,26 @@ public IStaticSchematic parse(InputStream input) throws IOException {
                     throw new UnsupportedOperationException("Unsupported Version of a Sponge Schematic");
             }
         }
+    },
+
+    /**
+     * The Litematica schematic specification. Commonly denoted by the ".litematic" file extension.
+     */
+    Litematica("litematic") {
```
leijurv, line 74: I believe enum name should be in all caps?

src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
```diff
@@ -65,6 +66,26 @@ public IStaticSchematic parse(InputStream input) throws IOException {
...
+        @Override
+        public IStaticSchematic parse(InputStream input) throws IOException {
+            NBTTagCompound nbt = CompressedStreamTools.readCompressed(input);
+            int version = nbt.getInteger("Version");
+            switch (version) {
+                case 4:
+                    return new LitematicaSchematic(nbt);
+                case 5:
+                case 6:
+                    throw new UnsupportedOperationException("This Schematic Verion is to new.");
```
leijurv, line 84: too

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+    int minX=0;
+    int minY=0;
+    int minZ=0;
+    private static final String reg = "Regions";
+    private static final String meta = "Metadata";
+    private static final String schemSize = "EnclosingSize";
+    private static final String blSt = "BlockStates";
+    private static final String blStPl = "BlockStatePalette";
+    private static final String pos = "Position";
+    private static final String size = "Size";
```
leijurv, line 49: don't do this, instead just use the constant strings where used. if they have to be static final, name of field should be all caps with underscores between words

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+
+/**
+ * @author Emerson
+ * @since 12/27/2020
+ * @author rycbar
+ * @since 22.09.2022
+ *
+ */
+public final class LitematicaSchematic extends StaticSchematic {
+    int minX=0;
```
leijurv, line 40: run the intellij auto formatter, also remove the =0, also make final and ideally private final

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+     * @param s axis that should be read.
+     * @return the lower cord of the requested axis.
+     */
+    private static int getMinimumCord(NBTTagCompound nbt, String subReg, String s) {
+        int a = nbt.getCompoundTag(reg).getCompoundTag(subReg).getCompoundTag(pos).getInteger(s);
+        int b = nbt.getCompoundTag(reg).getCompoundTag(subReg).getCompoundTag(size).getInteger(s);
+        if (b < 0) {
+            b++;
+        }
+        return Math.min(a,a+b);
```
leijurv, line 102: what is going on here? can size be negative????

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+        return nbt.getCompoundTag(reg).getKeySet().toArray(new String[0]);
+    }
+
+    /**
+     * Calculates the minimum cords/origin of the schematic as litematica schematics
+     * can have a non-minimum origin.
+     */
+    private void getMinimumCorner(NBTTagCompound nbt) {
+        for (String subReg : getRegions(nbt)) {
+            this.minX = Math.min(this.minX, getMinimumCord(nbt, subReg,"x"));
```
leijurv, line 86: this feels gross. modifying minx like this is weird. can we instead calculate these mins "once and for all" in the constructor then save them in a final field?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+            this.minY = Math.min(this.minY, getMinimumCord(nbt, subReg,"y"));
+            this.minZ = Math.min(this.minZ, getMinimumCord(nbt, subReg,"z"));
+        }
+    }
+
+    /**
+     * @param s axis that should be read.
+     * @return the lower cord of the requested axis.
+     */
+    private static int getMinimumCord(NBTTagCompound nbt, String subReg, String s) {
```
leijurv, line 96: replace "cord" with "coord" everywhere

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+     */
+    private static int getBitsPerBlock(int amountOfBlockTypes) {
+        return  (int) Math.floor((Math.log(amountOfBlockTypes)) / Math.log(2))+1;
+    }
+
+    /**
+     * @return the volume of the subregion.
+     */
+    private static long getVolume(NBTTagCompound nbt, String subReg) {
+        return Math.abs(
```
leijurv, line 169: add comment explaining why abs is needed (because size is sometimes negative?)

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * @param x cord of the schematic.
+     * @param y cord of the schematic.
+     * @param z cord of the schematic.
+     * @return if the current block is inside the subregion.
+     */
+    private static boolean inSubregion(NBTTagCompound nbt, String subReg, int x, int y, int z) {
+        return
+                x < Math.abs(nbt.getCompoundTag(reg).getCompoundTag(subReg).getCompoundTag(size).getInteger("x")) &&
```
leijurv, line 211: how does this work if size can be negative? what if x is negative too?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,296 @@
...
+                y < Math.abs(nbt.getCompoundTag(reg).getCompoundTag(subReg).getCompoundTag(size).getInteger("y")) &&
+                z < Math.abs(nbt.getCompoundTag(reg).getCompoundTag(subReg).getCompoundTag(size).getInteger("z"));
+    }
+
+    /** LitematicaBitArray class from litematica */
+    private static class LitematicaBitArray
```
leijurv, line 217: this all looks copy pasted. is that legal given the license of wherever you pasted it from? is credit required? is it compatible with baritone's lgpl license?

leijurv (changes requested):

src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
rycbar0: what do you mean here? there are 3 schematic version used atm v4 for 1.12 v5 for 1.13-1.17 and v6 1.18+.
it is intentional that bouth case 5&6 point to the unsupportet version exception for beeing too new

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: litematicas schematics can point in all 8 cubical quadrants and therefor can have boxes with negative coordinat sizes.
schematic origin can be anywhere inside our outside of the schematic box. (only if manuay changed during creation of schematic else origin will be minimum corner)
attached excample schematic i used for testing (as txt because of github just remove the txt at the end)

[Main.litematic.txt](https://github.com/cabaletta/baritone/files/9677835/Main.litematic.txt)
[Main-cleartext.txt](https://github.com/cabaletta/baritone/files/9677847/Main-cleartext.txt)

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: exactly if a  uneven amount of axis are negative the volume would be negative, but there can only be positive volumes

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: size can be negative but we take the absulute value to compare it, makeing it allways positive.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: yes it is.
i will double check this  with the original author.

rycbar0: image

src/main/java/baritone/command/defaults/LitematicaCommand.java
```diff
@@ -0,0 +1,71 @@
...
+        super(baritone, "litematica");
+    }
+
+    @Override
+    public void execute(String label, IArgConsumer args) throws CommandException {
+        int schematic = 0;
+        if(args.hasAny()) {
+            args.requireMax(1);
+            if (args.is(Integer.class)) {
+                schematic = args.getAs(Integer.class)-1;
```
leijurv, line 41: run intellij autoformatter

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,345 @@
...
+    /**
+     * @param nbtTagCompound a decompressed file stream aka nbt data.
+     * @param rotated        if the schematic is rotated by 90° aka x and z size are switched.
+     */
+    public LitematicaSchematic(NBTTagCompound nbtTagCompound, boolean rotated) {
+        this.nbt = nbtTagCompound;
+        this.offsetMinCorner = new Vec3i(getMinOfSchematic("x"), getMinOfSchematic("y"), getMinOfSchematic("z"));
+        this.y = Math.abs(nbt.getCompoundTag("Metadata").getCompoundTag("EnclosingSize").getInteger("y"));
+
+        if (rotated) {
```
leijurv, line 53: this isn't rotation. if you swap x and z that's actually a mirror along a 45 degree axis. swapping x and z is not the same as a 90 degree rotation. in fact, swapping x and z is not the same as rotating by any number of degrees, it's a mirror not a rotate

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -0,0 +1,345 @@
...
+     * Subregion don't have to be the same size as the enclosing size of the schematic. If they are smaller we check here if the current block is part of the subregion.
+     *
+     * @param x coord of the block relative to the minimum corner.
+     * @param y coord of the block relative to the minimum corner.
+     * @param z coord of the block relative to the minimum corner.
+     * @return if the current block is part of the subregion.
+     */
+    private static boolean inSubregion(NBTTagCompound nbt, String subReg, int x, int y, int z) {
+        return
+                x < Math.abs(nbt.getCompoundTag("Regions").getCompoundTag(subReg).getCompoundTag("Size").getInteger("x")) &&
```
leijurv, line 171: does this also need to check if they are negative? should this have a `x >=0 && y >= 0 && z >=0 && ` at the front?

leijurv:
yeah this generally seems good at this point

how easy will it be to port this to newer versions of minecraft? does litematica still work a similar way, or is this code basically "1.12.2 only" and not useful for newer versions?

let's add an explicit comment at the top of every class that contains pasted code from emerson that says something like "copied from this file (link to exact file), the license is LGPLv3 which is compatible with baritone, additionally the author gave explicit permission" (it doesnt have to be phrased exactly like that but it should get that message across)

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: what i meant here was that the size in the x direction after a 90° rotation is the same as the size in the z direction befor the turn and vice versa

leijurv: ok plz add a comment explaining how rotation versus mirror works so that others arent confused in the same way that i am (because right now it looks wrong to the reader)

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rycbar0: no the xyz values come from a counter in a for loop that counts from 0 up to schemsize-1

leijurv: ok even if it doesn't need the check please add it for correctness sake (it is definitely incorrect that `inSubregion(-1000000, -1000000, -1000000)` would return true)

rycbar0:
i havnt checked what would have to be done but i think it wont be that much worke to push it to the newer versions.

there are only two classes that have code that is copied, the LitematicaSchematic code that has 1 method left from emmerson and the BitArray class at the end of the LitematicaSchematic class which is from the litematica mod. both should be marked as who is the original author and work under the LPGLv3.

Does someone else also have to review this because else i see it now fit to be merged.

PR #3688, opened 2022-10-12
Merge 1.17.1 into 1.18.2

ZacSharp:
As always, if you update 1.17.1 I'd like to remerge before this one is merged.


src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
```diff
@@ -65,6 +66,27 @@ public IStaticSchematic parse(InputStream input) throws IOException {
...
+            int version = nbt.getInt("Version");
+            switch (version) {
+                case 4: //1.12
+                    throw new UnsupportedOperationException("This litematic Version is too old.");
+                case 5: //1.13-1.17
+                    return new LitematicaSchematic(nbt, false);
+                case 6: //1.18+
+                    throw new UnsupportedOperationException("This litematic Version is too new.");
+                default:
+                    throw new UnsupportedOperationException("Unsuported Version of a Litematica Schematic");
```
rycbar0, line 87:
```
switch (version) {
 case 4: //1.12
 case 5: //1.13-1.17
  throw new UnsupportedOperationException("This litematic Version is too old.");
 case 6:
  return new LitematicaSchematic(nbt, false);
 default:
  throw new UnsupportedOperationException("Unsuported Version of a Litematica Schematic");
```

rycbar0 (changes requested):
the switch statement in the litematic version check needs to be adjusted

src/main/java/baritone/utils/schematic/format/DefaultSchematicFormats.java
wagyourtail: do you have to discard older versions? or is there a way to work with them?

rycbar0: there are blocks that will work fine like minecraft:glass over all versions.
other blocks like oak planks will be replaced with air blocks when they not match pre or post flattening.
there shouldnt be any issues between version 5 and 6 but litematica itself does deny loading a schematic from a version other then its own.
in short it wont break anything if we dont test the version but the shematic will maybe be built wrong/incompleet

ZacSharp: Are there any differences in parsing?

rycbar0: in the parsing itself no.
but a pre flattening 1.12 schematic with minecraft:stone{variant: garnite} cant be resolved in a post flattening verison and defaults to a minecraft:air block and vice versa

Issue #6, opened 2018-08-14
diagonal over magma


leijurv:
right now its cost_inf, it should instead be sqrt(2)*SNEAK_ONE_BLOCK_COST and during execution it should know to sneak over magma to avoid damage. 

this applies to traverse too actually

leijurv:
Could do the same for MovementTraverse lol

mmvanheusden:
leijurv stop playing with your toenails and close this issue already!

leijurv:
?? issue won't be closed until the issue is fixed

Amaroq-Clearwater:
Well, has it been fixed yet?

leijurv:
No. The issue is still open.

banana-man95:
Is it fixed yet?

ZacSharp:
No and **please** don't make this another of those issues spammed with "Is it fixed?", "Any update?", "Fix please", "I want this", etc.

leijurv:
its funny because you could do this with maybe 5 lines of code

leijurv:
if i tabbed to intellij right now i bet i could get this done within 2 minutes (if i timed myself) (speedrun)

mmvanheusden:
feet nail cut speedrun

PR #3690, opened 2022-10-13
resolves issue #6

rycbar0:
it took me more then 2 mins and also more then 5 lines of code.

### DIFF
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -280,27 +280,20 @@ public interface MovementHelper extends ActionCosts, Helper {
                 || block == Blocks.WEB;
     }
 
-    static boolean isOverMagma(IPlayerContext ctx) {
+    static boolean isOverMagma(IPlayerContext ctx, BetterBlockPos src, BetterBlockPos dest) {
         int x = ctx.playerFeet().x;
         int y = ctx.playerFeet().y;
         int z = ctx.playerFeet().z;
         BlockStateInterface bsi = new BlockStateInterface(ctx);
 
-        return isOverMagma(bsi, x, y, z);
+        boolean moveDiagonaly = src.x != dest.x && src.z != dest.z;
+        return  isOverMagma(bsi, x, y, z) ||                                  //we are on magma
+                isOverMagma(bsi,x-(src.x-dest.x),y,z-(src.z-dest.z)) ||       //next block is magma
+                isOverMagma(bsi,x-(src.x-dest.x),y,z) && moveDiagonaly ||     //we are cutting the corner of a magma block
+                isOverMagma(bsi, x, y, z-(src.z - dest.z)) && moveDiagonaly;  //we are cutting the corner of a magma block other side
     }
-    //todo its terrible but it does the job for the moment
     static boolean isOverMagma(BlockStateInterface bsi, int x, int y, int z) {
-        Block block0 = bsi.get0(x,y-1,z).getBlock();
-        Block block1 = bsi.get0(x+1,y-1,z).getBlock();
-        Block block2 = bsi.get0(x-1,y-1,z).getBlock();
-        Block block3 = bsi.get0(x,y-1,z+1).getBlock();
-        Block block4 = bsi.get0(x,y-1,z-1).getBlock();
-
-        return  block0 == Blocks.MAGMA ||
-                block1 == Blocks.MAGMA ||
-                block2 == Blocks.MAGMA ||
-                block3 == Blocks.MAGMA ||
-                block4 == Blocks.MAGMA;
+        return bsi.get0(x,y-1,z).getBlock() == Blocks.MAGMA;
     }
 
     /**
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -280,6 +280,29 @@ public interface MovementHelper extends ActionCosts, Helper {
                 || block == Blocks.WEB;
     }
 
+    static boolean isOverMagma(IPlayerContext ctx) {
+        int x = ctx.playerFeet().x;
+        int y = ctx.playerFeet().y;
+        int z = ctx.playerFeet().z;
+        BlockStateInterface bsi = new BlockStateInterface(ctx);
+
+        return isOverMagma(bsi, x, y, z);
+    }
+    //todo its terrible but it does the job for the moment
+    static boolean isOverMagma(BlockStateInterface bsi, int x, int y, int z) {
+        Block block0 = bsi.get0(x,y-1,z).getBlock();
+        Block block1 = bsi.get0(x+1,y-1,z).getBlock();
+        Block block2 = bsi.get0(x-1,y-1,z).getBlock();
+        Block block3 = bsi.get0(x,y-1,z+1).getBlock();
+        Block block4 = bsi.get0(x,y-1,z-1).getBlock();
+
+        return  block0 == Blocks.MAGMA ||
+                block1 == Blocks.MAGMA ||
+                block2 == Blocks.MAGMA ||
+                block3 == Blocks.MAGMA ||
+                block4 == Blocks.MAGMA;
+    }
+
     /**
      * Can I walk on this block without anything weird happening like me falling
      * through? Includes water because we know that we automatically jump on
@@ -294,7 +317,7 @@ public interface MovementHelper extends ActionCosts, Helper {
      */
     static boolean canWalkOn(BlockStateInterface bsi, int x, int y, int z, IBlockState state) {
         Block block = state.getBlock();
-        if (block == Blocks.AIR || block == Blocks.MAGMA) {
+        if (block == Blocks.AIR || (!Baritone.settings().allowSneakOverMagma.value && block == Blocks.MAGMA)) {
             // early return for most common case (air)
             // plus magma, which is a normal cube but it hurts you
             return false;
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -132,7 +132,19 @@ public class MovementDiagonal extends Movement {
                 }
             }
         }
-        double multiplier = WALK_ONE_BLOCK_COST;
+        Block cuttingOver1 = context.get(x, y - 1, destZ).getBlock(); //cutting over adjacent x-axis block
+        Block cuttingOver2 = context.get(destX, y - 1, z).getBlock(); //cutting over adjacent z-axis block
+        if (MovementHelper.isLava(cuttingOver1) || MovementHelper.isLava(cuttingOver2)) {
+            return;
+        } else if ((cuttingOver1 == Blocks.MAGMA || cuttingOver2 == Blocks.MAGMA) && !Baritone.settings().allowSneakOverMagma.value) {
+            return;
+        }
+        double multiplier;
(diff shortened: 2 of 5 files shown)


ZacSharp (changes requested):
As of now `MovementParkour`, `MovementAscend`, `MovementDescend` and `MovementFall` would walk on magma (because it's `canWalkOn` now) without pressing shift. Additionally `MovementParkour` might break when jumping from magma because the damage can prevent the jump. It shouldn't attempt to jump from magma, onto magma is fine as long as no damage is taken.

rycbar0:
should i check if the player wears boots with frost walker enchantment? those also prevent damage from magma blocks without sneaking. if so should i allow pathfinding over magma allways or only when `allowSneakOverMagma` is true?

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -280,6 +280,28 @@ static boolean avoidWalkingInto(Block block) {
...
+                         isOverMagma(bsi, dest.x, y, dest.z) ||                 //the block we are going too, at the same y level is magma
+                         (isOverMagma(bsi, dest.x, dest.y, dest.z)              //the block we are going too is magma and
+                                 && y < src.y) ||                               //we are already dropping
+                         ((src.x != dest.x && src.z != dest.z) &&               //when moving diagonaly and
+                                         (isOverMagma(bsi, src.x, y, dest.z) || //we are cutting over magma or
+                                          isOverMagma(bsi, dest.x, y, src.z) || //we are cutting over magma or
+                                          isOverMagma(bsi, dest.x, y, dest.z))); //we are ending up on magma
+
+        return result;
+    }
```
ZacSharp, line 300: If I understand this correctly this intends to check whether the player is touching magma *right now*, so why not use the same kind of check as in `MovementDiagonal.isSafeToCancel`?

src/main/java/baritone/pathing/movement/movements/MovementDescend.java
```diff
@@ -101,7 +101,7 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
         //C, D, etc determine the length of the fall
 
         IBlockState below = context.get(destX, y - 2, destZ);
-        if (!MovementHelper.canWalkOn(context.bsi, destX, y - 2, destZ, below)) {
+        if (!MovementHelper.canWalkOn(context.bsi, destX, y - 2, destZ, below) || (below.getBlock() == Blocks.MAGMA && !context.allowSneakOverMagma)) {
```
ZacSharp, line 104: Doesn't `canWalkOn` already return false for magma if `allowSneakOverMagma` is false?

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -91,6 +91,9 @@ public static void cost(CalculationContext context, int x, int y, int z, EnumFac
             return;
         }
         IBlockState standingOn = context.get(x, y - 1, z);
+        if (standingOn.getBlock() != Blocks.MAGMA) {
```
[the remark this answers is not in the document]
ZacSharp: Definitely looks like it should not be inverted.

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -72,28 +72,40 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
         IBlockState pb1 = context.get(destX, y, destZ);
         IBlockState destOn = context.get(destX, y - 1, destZ);
         Block srcDown = context.getBlock(x, y - 1, z);
+        if ((srcDown == Blocks.MAGMA || destOn.getBlock() == Blocks.MAGMA) && !context.allowSneakOverMagma) {
+            return COST_INF;
+        }
```
ZacSharp, line 77: Why disallow *leaving* magma?

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -72,28 +72,40 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
...
+            } else {
+                WC = WALK_ONE_BLOCK_COST / 2;
+            }
+            if (destOn.getBlock() == Blocks.MAGMA) {
+                WC += SNEAK_ONE_BLOCK_COST/2;
+            } else if (destOn.getBlock() == Blocks.SOUL_SAND) {
+                WC += WALK_ONE_OVER_SOUL_SAND_COST / 2;
+            } else {
+                WC += WALK_ONE_BLOCK_COST / 2;
+            }
```
ZacSharp, line 93: Why did you move this out of the `else` branch?
It's just useless calculations if there's water to swim in.

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -72,28 +72,40 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
...
-                    WC += (WALK_ONE_OVER_SOUL_SAND_COST - WALK_ONE_BLOCK_COST) / 2;
-                }
             }
             double hardness1 = MovementHelper.getMiningDurationTicks(context, destX, y, destZ, pb1, false);
             if (hardness1 >= COST_INF) {
                 return COST_INF;
             }
             double hardness2 = MovementHelper.getMiningDurationTicks(context, destX, y + 1, destZ, pb0, true); // only include falling on the upper block to break
-            if (hardness1 == 0 && hardness2 == 0) {
+            if (hardness1 == 0 && hardness2 == 0 && !MovementHelper.isOverMagma(context.bsi, x, y, z)) {
```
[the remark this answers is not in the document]
ZacSharp: It is necessary in the check one line below.
I.e. sneaking won't make the two blocks solid, but it will stop us from sprinting.

ZacSharp (changes requested):

ZacSharp:
> should i check if the player wears boots with frost walker enchantment? those also prevent damage from magma blocks without sneaking. if so should i allow pathfinding over magma allways or only when `allowSneakOverMagma` is true?

I'd say just treat magma as a harmless solid block without asking (or add yet another setting for it).

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -280,6 +281,45 @@ static boolean avoidWalkingInto(Block block) {
...
+    static boolean hasFrostWalker(CalculationContext context) {
+        return hasFrostWalker(context.getBaritone().getPlayerContext());
+    }
+
+    static boolean hasFrostWalker(IPlayerContext ctx) {
+        boolean frostwaker = false;
+        NBTTagList nbtTagList = ctx.player().inventory.player.inventory.player.inventory.player.inventory.player.inventory.armorInventory.get(0).getEnchantmentTagList(); //get all enchantments of the players boots
+        for (int i = 0; i<nbtTagList.tagCount();i++) {
+            if(nbtTagList.get(i).toString().contains("id:9s")) { //this is very hacky pls advice on how to test for frostwalker enchantment the proper way
+                frostwaker = true;
```
rycbar0, line 317: this could probably be made into a early return statment but since boots can hold at moste 9 enchantments includiung curses im not sure if its worth it.

rycbar0 (commented):
TODO-list:

- [x] javadoc for `isOverMagma()`
- [x] javadoc for `hasFrostwalker()`
- [ ] adjust jump timing when sneak ascending
- [x] sneak when breaking a block while standing on magma
- [x] evaluate optimization potential of costcalculation in `MovementDiagonal `

ZacSharp:
> i understand this that i should remove the allowSneakOverMagma setting and allways allow pathfinding over magma

Oh, sorry, that's my comment editing gone wrong. I originally intended to write something along the lines of "I'd say just use the boots without asking (or make it yet another setting)", but obviously I reworded that beyond repair.
As long as it is reliable I don't see a need for a setting (it complicates things after all) and frostwalker boots are 100% reliable.

As for "allowSneakOverMagma": I personally don't care.

src/main/java/baritone/pathing/movement/MovementHelper.java
ZacSharp: Is this what you are looking for?
https://github.com/ZacSharp/baritone/blob/fb814e912dadf3f6e34358354a86b54674103081/src/main/java/baritone/pathing/movement/CalculationContext.java#L104
or maybe `net.minecraft.enchantment.EnchantmentHelper.hasFrostWalkerEnchantment`?

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -132,27 +132,33 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
...
+            multiplier = SNEAK_ONE_BLOCK_COST;
+        } else {
+            multiplier = WALK_ONE_BLOCK_COST;
+        }
         // For either possible soul sand, that affects half of our walking
         if (destWalkOn.getBlock() == Blocks.SOUL_SAND) {
             multiplier += (WALK_ONE_OVER_SOUL_SAND_COST - WALK_ONE_BLOCK_COST) / 2;
         } else if (destWalkOn.getBlock() == Blocks.WATER) {
             multiplier += context.walkOnWaterOnePenalty * SQRT_2;
+        } else if (destWalkOn.getBlock() == Blocks.MAGMA && (cuttingOver1 != Blocks.MAGMA || cuttingOver2 != Blocks.MAGMA) && !MovementHelper.hasFrostWalker(context)) {
```
ZacSharp, line 151: I think it should be `cuttingOver1 != magma && cuttingOver2 != magma`.
As soon as one of the two is magma (and there's no frostwalker boots) the multiplier is already `SEAK_ONE_BLOCK_COST`. So only if both are not magma it may be increased.

ZacSharp (commented):
To sneak while breaking blocks, `Movement.prepared` might be the right place.

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -280,6 +281,45 @@ static boolean avoidWalkingInto(Block block) {
...
+        }
+        return false; //if we get this far there is no magma
+    }
+    static boolean isOverMagma(BlockStateInterface bsi, int x, int y, int z) {
+        return bsi.get0(x,y-1,z).getBlock() == Blocks.MAGMA;
+    }
+
+    static boolean hasFrostWalker(CalculationContext context) {
+        return hasFrostWalker(context.getBaritone().getPlayerContext());
+    }
```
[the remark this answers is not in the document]
ZacSharp: Perhaps I should have named reasons so you know why I called this a no-go.
The fact whether nor not the player has frost walker is constant during a pathing operation so the suggestion was to cache the boolean result of `EnchantmentHelper.hasFrostWalkerEnchantment` (what I had in mind is `public final boolean frostWalker = EnchantmentHelper.hasFrostWalkerEnchantment(player);`).
There's a rule about cost calculation doing no object allocation and `EnchantmentHelper` breaks that rule.
The pathing thread may not access game state because Minecraft is singlethreaded.
Recalculating a constant two and a half million times per second is a performance nightmare.

ZacSharp:
Autoformatting caused quite a bunch of unrelated changes but ok.
Code looks good to me now (apart from the todo comment).

It mostly sneaks way to much and not at all after parkour jumps onto magma.

https://user-images.githubusercontent.com/68165024/198906464-d94d19ee-8823-4b36-9593-d9dc5262f864.mp4

Do you think you could make it more like this (but without sneaking too little)?

https://user-images.githubusercontent.com/68165024/198906535-1ded1248-d6a1-48f1-9c1f-d5ad7fee77de.mp4

ZacSharp:
Implemented in #4742

Issue #3565, opened 2022-07-15
Nasty parts of Baritone


leijurv:
I am going to write down the nasty stuff. I won't include small issues, this is more of big structural problems. For example, InventoryBehavior could be better about figuring out what to put where on the hotbar, and it could be less hardcoded about the pickaxe and the throwaway blocks. This could be fixed without too much effort, just within that class. That's an example of something that isn't listed.

**MovementHelper checks.** Endlessly expanding nests of if statements, that get worse as we get into newer versions of MC. This is not good for performance. This is pretty much entirely fixed by #3479

**World height in chunk caching.** This is a big stinky, see #3375. The on disk format is also nasty, for example it uses strings for blocks, and it wasn't updated between 1.12 and 1.13. Some blocks have the same name between those versions, some don't.

**Basically all the movements executors.** For example, MovementFall has this absolutely nasty bit of code https://github.com/cabaletta/baritone/blob/9c64067d492e3fdd7d255bdf39092524046f7121/src/main/java/baritone/pathing/movement/movements/MovementFall.java#L134-L167 that governs how it strafes in midair while falling. It combines looking around, WASD, and sneaking. I tested this by making a 1x1 column of air through lava, and seeing if it could properly dodge ladders without ever touching any of the four lava walls. I have no idea if my testing was thorough. I have no idea if this works beyond 1.12.2. I have no idea how any of it works. The weird logic around when to start or stop sneaking is crazy. You may have noticed that baritone sneaks for a little bit at the beginning of every fall, this is why. Also `PathExecutor` just makes the whole thing much worse with all its bajillion special cases that override the individual movements.

**BuilderProcess.**

[...]

PR #3698, opened 2022-10-24
Reliably clear keys when paused

ZacSharp:

### DIFF
--- a/src/main/java/baritone/command/defaults/ExecutionControlCommands.java
+++ b/src/main/java/baritone/command/defaults/ExecutionControlCommands.java
@@ -56,6 +56,7 @@ public class ExecutionControlCommands {
 
                     @Override
                     public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
+                        baritone.getInputOverrideHandler().clearAllKeys();
                         return new PathingCommand(null, PathingCommandType.REQUEST_PAUSE);
                     }
 


wagyourtail:
Doesnt pausing remove the input override handler anyway?

ZacSharp:
`PathingBehavior` clears all keys after the first tick where a pause is requested (the first tick that is actually paused), which clears the keys *almost* always and reliably works for pure pathing processes.
However if you use something like `BuilderProcess` there are situation where the process presses down a key and requests a pause, so when control transfers to `#pause` then the pathing command is already `REQUEST_PAUSE` and `PathingBehavior` does *not* clear anything.
This is especially annoying if it happens with wasd or shift, because those keys are the trigger for Baritone blocking user input.

ZacSharp:
Actually, you can't rely on inputs *not* being cleared already so why not add this to `InputOverrideHandler.onTick` instead?
I don't know of anything using the input system that does not unconditionally clear keys while it is running and `BackfillProcess` even does it while idling.

ZacSharp:
> Actually, you can't rely on inputs not being cleared already so why not add this to `InputOverrideHandler.onTick` instead?

Because `InputOverrideHandler.onTick` is called before Minecraft reads the movement inputs so it would stop wasd, shift and space from working.

leijurv:
lol this is so #3565

PR #3703, opened 2022-10-26
formula error

rycbar0:
apparently i have a error in my formula that prevents schematics with only air or exactly a power of 2 diffrent blocks from being loaded.


ZacSharp (approved):

ZacSharp:
Actually this breaks if you somehow manage to get exactly `2^29` block states in a single subregion because of floating point imprecision (`Math.log(536870912) / Math.log(2) == 29.000000000000004` so rounding outputs 30 instead of 29).
That schematic would be roughly 14k by 256 by 14k blocks huge though and there aren't enough blockstates in the game for this to happen anyways.

PR #3735, opened 2022-11-22
Allow breaking next to some liquids

ZacSharp:

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -108,6 +108,13 @@ public final class Settings {
     public final Setting<Double> walkOnWaterOnePenalty = new Setting<>(3D);
 
     /**
+     * Don't allow breaking blocks next to liquids.
+     * <p>
+     * Enable if you have mods adding custom fluid physics.
+     */
+    public final Setting<Boolean> strictLiquidCheck = new Setting<>(false);
+
+    /**
      * Allow Baritone to fall arbitrary distances and place a water bucket beneath it.
      * Reliability: questionable.
      */
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -77,7 +77,18 @@ public interface MovementHelper extends ActionCosts, Helper {
                 && BlockFalling.canFallThrough(bsi.get0(x, y - 1, z))) { // and if it would fall (i.e. it's unsupported)
             return true; // dont break a block that is adjacent to unsupported gravel because it can cause really weird stuff
         }
-        return block instanceof BlockLiquid;
+        if (block instanceof BlockLiquid) {
+            if (directlyAbove || Baritone.settings().strictLiquidCheck.value) {
+                return true;
+            }
+            int level = state.getValue(BlockLiquid.LEVEL);
+            if (level == 0) {
+                return true; // source blocks like to flow horizontally
+            }
+            // everything else will prefer flowing down
+            return !(bsi.get0(x, y - 1, z).getBlock() instanceof BlockLiquid); // assume everything is in a static state
+        }
+        return false;
     }
 
     static boolean canWalkThrough(IPlayerContext ctx, BetterBlockPos pos) {


leijurv:
I'm a little worried about performance - `avoidAdjacentBreaking` is called a **lot**:

https://github.com/cabaletta/baritone/blob/eabd1150b0a619f7e2adada9dfbd837613bf9414/src/main/java/baritone/pathing/movement/MovementHelper.java#L60-L64

Nothing looks non-performant, to be clear. I can't think of anything to speed up. It's just about whether the slowdown is worth it at all, relative to how much of an improvement this is. 😬 sorry to say it but I think this is a very small improvement, right? For the use case of pathing through vanilla mc terrain gen, this probably affects something like one in a million `avoidBreaking` checks that will now return false instead of true... and what percent of those were the limiting factor in making a calculated path slower?

ZacSharp:
<details><summary>This (plus code to do the counting)</summary>

```java
    static boolean avoidAdjacentBreaking(BlockStateInterface bsi, int x, int y, int z, boolean directlyAbove) {
        // returns true if you should avoid breaking a block that's adjacent to this one (e.g. lava that will start flowing if you give it a path)
        // this is only called for north, south, east, west, and up. this is NOT called for down.
        // we assume that it's ALWAYS okay to break the block thats ABOVE liquid
        IBlockState state = bsi.get0(x, y, z);
        Block block = state.getBlock();
        if (!directlyAbove // it is fine to mine a block that has a falling block directly above, this (the cost of breaking the stacked fallings) is included in cost calculations
                // therefore if directlyAbove is true, we will actually ignore if this is falling
                && block instanceof BlockFalling // obviously, this check is only valid for falling blocks
                && Baritone.settings().avoidUpdatingFallingBlocks.value // and if the setting is enabled
                && BlockFalling.canFallThrough(bsi.get0(x, y - 1, z))) { // and if it would fall (i.e. it's unsupported)
            incrementCount("Unsupported falling");
            return true; // dont break a block that is adjacent to unsupported gravel because it can cause really weird stuff
        }
        if (block instanceof BlockLiquid) {
            if (directlyAbove || Baritone.settings().strictLiquidCheck.value) {
                incrementCount("Liquid directlyAbove");
                return true;
            }
            int level = state.getValue(BlockLiquid.LEVEL);
            if (level == 0) {
                incrementCount("Source liquid");
                return true; // source blocks like to flow horizontally
            }
            // everything else will prefer flowing down
            if (!(bsi.get0(x, y - 1, z).getBlock() instanceof BlockLiquid)) { // assume everything is in a static state
                incrementCount("Ground liquid");
                return true;
            } else {
                incrementCount("Waterfall");
                return false;
            }
        }
        incrementCount("Default");
        return false;
    }
```
</details>
<details><summary>Gave me these numbers</summary>

|         | Unsupported falling |Liquid directlyAbove | Source liquid | Ground liquid | Waterfall | Default |
| ------------- |-----------------------:| ----------------------------:|--------------------:|---------------------:|------------:|---:|
| walk 10k blocks without breaking                                    | 46 | 61330 | 11359 | 67 | 10 | 6000126 |
| walk 10k blocks                                                                    | 11810 | 471769 | 526772 | 34306 | 11436 | 1187766862 |
| walk 10k blocks underground                                           | 5763 | 1145443 | 2064633 | 462436 | 81403 | -1423917901* |
| mine diamonds until ~20k blocks were mined in total | 0 | 324575 | 556710 | 82647 | 7138 | 1457269714 |
| walk 10k blocks underground again                                | 0 | 586290 | 1102035 | 196261 | 46100 | 3182660109 |

\* I used ints at first so this value wrapped around twice
All cases except the first use diamond tools with efficiency 5 and the last two have `avoidUpdatingFallingBlocks` disabled (hence no returns from it).
For the last two I repurposed `BackfillProcess` to count what kind of block was actually broken, yielding

|      | Normal | Next to liquid | Next to unsupported falling blocks |
|---|---|---|---|
| mine diamonds until ~20k blocks were mined in total | 29240 | 0 | 0 |
| walk 10k blocks underground again                                | 6502 | 6 | 0 |
</details>
<details><summary>Screenshots from the third run (10k blocks underground)</summary>

I didn't watch the others





</details>

This seems to happen about as often as running into unsupported `BlockFalling`s and depending on the performance of `IblockState.getValue` should perform similarly, however floating sand is much more common in deserts and for liquids we are currently doing nothing at all, which is 100% faster than any extra logic. Given that the gain seen in the screenshots is small and there was not case where mining could take an extra ore the extra property lookup for >50% of all checked liquids might actually not be worth the effort.
Something like one of the traversable versions of 2b2t spawn with water and lava streams all over the place might be worth checking, but I won't do that today.

leijurv:
is this a correct interpretation of your analysis? between 0.01% and 2.4% of all liquids will now return `false` instead of `true`? if so, then I think this is worth merging probably

leijurv:
@ZacSharp if my interpretation is correct I'll merge, but please confirm that first

ZacSharp:
It's on my todo list, but I got other things before it on there.

ZacSharp:
It's, the percentage of measured liquids that are not next to an unsupported falling block and will now return `false` instead of `true`.
If you assume that liquids next to unsupported sand/gravel don't happen and you assume that my test is representative then your interpretation is correct and this is the percentage of all fluids now returning `false`.

PR #3743, opened 2022-11-28
Fix Registry.unregister

ZacSharp:
It's been more than three years and we finally got the first working version of Registry.unregister!


wagyourtail (approved):

leijurv:
lmfao

PR #3755, opened 2022-12-10
allowBreakAnyway should let mining happen

wagyourtail:
even if it will cause issues with ores beyond walls...
Peek 2022-12-10 16-32

idk if that's fixable, theoretically the allowbreaklist could have something common like stone on it so we can't just check if it's surrounded by not air

but I still think this change is worth making

### DIFF
--- a/src/main/java/baritone/process/MineProcess.java
+++ b/src/main/java/baritone/process/MineProcess.java
@@ -102,10 +102,17 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
                 return null;
             }
         }
+
         if (!Baritone.settings().allowBreak.value) {
-            logDirect("Unable to mine when allowBreak is false!");
-            cancel();
-            return null;
+            this.filter = new BlockOptionalMetaLookup(filter.blocks()
+                .stream()
+                .filter(e -> Baritone.settings().allowBreakAnyway.value.contains(e.getBlock()))
+                .toArray(BlockOptionalMeta[]::new));
+            if (this.filter.blocks().isEmpty()) {
+                logDirect("Unable to mine when allowBreak is false!");
+                cancel();
+                return null;
+            }
         }
         updateLoucaSystem();
         int mineGoalUpdateInterval = Baritone.settings().mineGoalUpdateInterval.value;
@@ -467,9 +474,15 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
     public void mine(int quantity, BlockOptionalMetaLookup filter) {
         this.filter = filter;
         if (filter != null && !Baritone.settings().allowBreak.value) {
-            logDirect("Unable to mine when allowBreak is false!");
-            this.mine(quantity, (BlockOptionalMetaLookup) null);
-            return;
+            this.filter = new BlockOptionalMetaLookup(filter.blocks()
+                .stream()
+                .filter(e -> Baritone.settings().allowBreakAnyway.value.contains(e.getBlock()))
+                .toArray(BlockOptionalMeta[]::new));
+            if (this.filter.blocks().isEmpty()) {
+                logDirect("Unable to mine when allowBreak is false!");
+                this.mine(quantity, (BlockOptionalMetaLookup) null);
+                return;
+            }
         }
         this.desiredQuantity = quantity;
         this.knownOreLocations = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/process/MineProcess.java
```diff
@@ -102,10 +102,17 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
...
-            return null;
+            this.filter = new BlockOptionalMetaLookup(filter.blocks()
+                .stream()
+                .filter(e -> Baritone.settings().allowBreakAnyway.value.contains(e.getBlock()))
+                .toArray(BlockOptionalMeta[]::new));
+            if (this.filter.blocks().isEmpty()) {
+                logDirect("Unable to mine when allowBreak is false!");
+                cancel();
+                return null;
+            }
```
ZacSharp, line 115: I don't think pruning every tick is a good idea because it causes inconsistent behavior when `allowBreakAnyway` is changed. Either we prune once at the start and ignore changes to `allowBreakAnyway` or we filter the original list every tick so both removals from and additions to `allowBreakAnyway` are picked up.
Since `MineProcess` also immediately responds to other setting changes I think filtering is the way to go. Not sure whether it is worth the effort to cache filtered list (and a copy of the setting value) to avoid refiltering.

ZacSharp (commented):
We could adjust the plausibility check to check whether an ore is "not surrounded by not allowBreakAnyway" (i.e. has blocks on `allowBreakAnyway` next to it)

src/main/java/baritone/process/MineProcess.java
wagyourtail: did you make some kind of setting change watcher in one of your pr's? but yeah, I guess I should store the original for refiltering each tick at least

wagyourtail:
> We could adjust the plausibility check to check whether an ore is "not surrounded by not allowBreakAnyway" (i.e. has blocks on `allowBreakAnyway` next to it)

usually ore will have one other of itself bordering it, so i don't think this will be effective enough to be worth doing, the only ore that would almost always benefit from this really is emerald...

src/main/java/baritone/process/MineProcess.java
ZacSharp: Regardless of whether we attempt to cache it (no, I have not yet implemented a setting watcher), the filtering should happen in `rescan`, not in `onTick`.

ZacSharp:
We'd get a slightly higher hit rate because the player is two blocks high but that's still not worth bothering so how about this more complicated one?
1. If an ore is next to a two high (or higher) column of blocks on `allowBreakAnyway`, keep it.
2. If an ore is next to a two high (or higher) column of kept ore or blocks on `allowBreakAnaway`, keep it

Not sure whether we want to do this on a near 20Hz basis. Maybe it is really best to let the blacklist "solve" the problem.

Also for consistency with other settings affecting ore selection there should also be a check in `prune`.

wagyourtail:
> We'd get a slightly higher hit rate because the player is two blocks high but that's still not worth bothering so how about this more complicated one?
> 
>     1. If an ore is next to a two high (or higher) column of blocks on `allowBreakAnyway`, keep it.
> 
>     2. If an ore is next to a two high (or higher) column of kept ore or blocks on `allowBreakAnaway`, keep it
> 
> 
> Not sure whether we want to do this on a near 20Hz basis. Maybe it is really best to let the blacklist "solve" the problem.

the first and second mean the same thing, because the ore has to be allowed to be mined... yeah I could add it and see how it does I guess...

> Also for consistency with other settings affecting ore selection there should also be a check in `prune`.

because searchWorld is also used in the get to block goals, and searchworld uses prune, I don't think so... it would block going to a block that you can't break

ZacSharp:
> because searchWorld is also used in the get to block goals, and searchworld uses prune, I don't think so...

Didn't notice it was public and takes the filter as a parameter already. Filtering the filter before passing it in like you are doing it now seems to be the solution then.

> the first and second mean the same 

Right. It should be "block in `allowBreakAnyway` but not in `filter`" so that only the boundary of a vein is relevant.

src/main/java/baritone/process/MineProcess.java
```diff
@@ -115,7 +111,7 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
             Baritone.getExecutor().execute(() -> rescan(curr, context));
         }
         if (Baritone.settings().legitMine.value) {
-            addNearby();
+            if (!addNearby()) return null;
```
ZacSharp, line 114: You forgot to cancel here so this crashes the game with an `IllegalStateException` from `PathingControlManager`

leijurv:
Cool idea. Seems like no performance impact since this simply computes a different BOM filter on each scan?

PR #3771, opened 2022-12-26
Merge master into 1.16.5

ZacSharp:
This pr also removes the obsolete `MixinLongArrayNBT`.

Draft until someone tells me they have double checked `MovementHelper`. That thing was a pain to merge.


ZacSharp:
Why is Baritone not allowed to walk on honey blocks?

Why do we pass `BlockPos.ORIGIN` at the end of `canWalkThrough` since `bsi.access` was merged to 1.13.2 in 1edd7126fda9b3b822c5dcbfecebcf58e7e9134b?

leijurv:
You can skip merging through 1.13, 1.14, and 1.15 ever since those branches were deprecated by the way. Or if it's easier to do it that way feel free to continue, just wanted to make sure you knew :)

leijurv:
I think honey was just annoying to code for? idk

ZacSharp:
Found a discreptancy between pre-precomuteddata and current behavior with waterlogged blocks:
Baritone now considers breaking them a valid option to make them traversable, which causes the path to cancel and recalculate for every broken waterlogged block (no loop though because the block is broken).

Also I know I don't need to to old branches, I just like keeping the differences between versions low. It also helps in situations like this one where you unexpectedly merge my pr while I'm working on it, because this way I only had to do 1.15.2 -> 1.16.5 from scratch. That's way easier than master -> 1.16.5 because the bulk of functionality changes is master -> 1.13.2 and the renaming is mostly 1.13.2 -> 1.14.4 and 1.14.4 -> 1.15.2.

leijurv:
hee hee

leijurv:
mentioned in #3792

leijurv:
makes sense about the old branches too!

Issue #1586, opened 2020-04-18
Reflexes


LoganDark:
when you falling down a wall and place block to avoid 500 fall damage.

when you jumping a jump and player places 2 blocks

![image](https://user-images.githubusercontent.com/4723091/79624514-df8ae580-80d6-11ea-9bbd-963ab9c1cc6d.png)

ToxicDevv:
I would love to see this added!!!

UltraBlackLinux:
same :D

dominikaaaa:
Commenting "i'd like to see this added" on issues is no more useful then just reacting to it, and it sends everyone a notification the someone commented instead, which isn't helpful and doesn't contribute to the issue itself.

PR #3775, opened 2022-12-31
Allow pathing to use frostwalker

### LINKED ISSUE #3571: Treat Water as walkable if FrostWalker
## Describe your suggestion
Baritone should check if boots with FrostWalker are equiped and if they are water should be treated as a block you can walk on.

## Settings
A setting to turn this on or off.

## Context
Right now Baritone has issues with FrostWalk. It stops constantly to recalculate the path and sometimes walks long paths on land rather than a straight path through water.

## Final checklist

ZacSharp:
Wanted to get the pr open this year :laughing:
Short term pathing is tested pretty well but for other cases I only have a random collection of manual tests.

Closes #3571

### DIFF
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -30,7 +30,6 @@ import baritone.utils.BlockStateInterface;
 import baritone.utils.pathing.MutableMoveResult;
 import com.google.common.collect.ImmutableSet;
 import net.minecraft.block.Block;
-import net.minecraft.block.BlockLiquid;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.init.Blocks;
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -135,6 +135,7 @@ public class MovementDiagonal extends Movement {
                     return;
                 }
             }
+            frostWalker &= !context.assumeWalkOnWater; // do this after checking for descends because jesus can't prevent the water from freezing, it just prevents us from relying on the water freezing
         }
         double multiplier = WALK_ONE_BLOCK_COST;
         // For either possible soul sand, that affects half of our walking
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -30,6 +30,7 @@ import baritone.utils.BlockStateInterface;
 import baritone.utils.pathing.MutableMoveResult;
 import com.google.common.collect.ImmutableSet;
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockLiquid;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.init.Blocks;
@@ -117,6 +118,7 @@ public class MovementDiagonal extends Movement {
         boolean ascend = false;
         IBlockState destWalkOn;
         boolean descend = false;
+        boolean frostWalker = false;
         if (!MovementHelper.canWalkThrough(context.bsi, destX, y, destZ, destInto)) {
             ascend = true;
             if (!context.allowDiagonalAscend || !MovementHelper.canWalkThrough(context.bsi, x, y + 2, z) || !MovementHelper.canWalkOn(context.bsi, destX, y, destZ, destInto) || !MovementHelper.canWalkThrough(context.bsi, destX, y + 2, destZ)) {
@@ -125,9 +127,8 @@ public class MovementDiagonal extends Movement {
             destWalkOn = destInto;
         } else {
             destWalkOn = context.get(destX, y - 1, destZ);
-            if (!MovementHelper.canWalkOn(context.bsi, destX, y - 1, destZ, destWalkOn)
-                    && !MovementHelper.canUseFrostWalker(context, destWalkOn)
-            ) {
+            frostWalker = MovementHelper.canUseFrostWalker(context, destWalkOn) && !(context.assumeWalkOnWater && context.getBlock(x, y - 1, z) instanceof BlockLiquid);
+            if (!MovementHelper.canWalkOn(context.bsi, destX, y - 1, destZ, destWalkOn) && !frostWalker) {
                 descend = true;
                 if (!context.allowDiagonalDescend || !MovementHelper.canWalkOn(context.bsi, destX, y - 2, destZ) || !MovementHelper.canWalkThrough(context.bsi, destX, y - 1, destZ, destWalkOn)) {
                     return;
@@ -138,8 +139,8 @@ public class MovementDiagonal extends Movement {
         // For either possible soul sand, that affects half of our walking
         if (destWalkOn.getBlock() == Blocks.SOUL_SAND) {
             multiplier += (WALK_ONE_OVER_SOUL_SAND_COST - WALK_ONE_BLOCK_COST) / 2;
-        } else if (!ascend && !descend && MovementHelper.canUseFrostWalker(context, destWalkOn)) {
-            // frostwalker lets us walk on water without the penalty, but only works if we don't ascend or descend
+        } else if (frostWalker) {
+            // frostwalker lets us walk on water without the penalty
         } else if (destWalkOn.getBlock() == Blocks.WATER) {
             multiplier += context.walkOnWaterOnePenalty * SQRT_2;
         }
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -127,7 +127,8 @@ public class MovementDiagonal extends Movement {
             destWalkOn = destInto;
         } else {
             destWalkOn = context.get(destX, y - 1, destZ);
-            frostWalker = MovementHelper.canUseFrostWalker(context, destWalkOn) && !(context.assumeWalkOnWater && context.getBlock(x, y - 1, z) instanceof BlockLiquid);
+            boolean standingOnABlock = !(context.getBlock(x, y - 1, z) instanceof BlockLiquid) || (!context.assumeWalkOnWater && !(context.getBlock(x, y, z) instanceof BlockLiquid));
+            frostWalker = standingOnABlock && MovementHelper.canUseFrostWalker(context, destWalkOn);
             if (!MovementHelper.canWalkOn(context.bsi, destX, y - 1, destZ, destWalkOn) && !frostWalker) {
                 descend = true;
                 if (!context.allowDiagonalDescend || !MovementHelper.canWalkOn(context.bsi, destX, y - 2, destZ) || !MovementHelper.canWalkThrough(context.bsi, destX, y - 1, destZ, destWalkOn)) {
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -127,7 +127,7 @@ public class MovementDiagonal extends Movement {
             destWalkOn = destInto;
         } else {
             destWalkOn = context.get(destX, y - 1, destZ);
-            boolean standingOnABlock = !(context.getBlock(x, y - 1, z) instanceof BlockLiquid) || (!context.assumeWalkOnWater && !(context.getBlock(x, y, z) instanceof BlockLiquid));
+            boolean standingOnABlock = MovementHelper.mustBeSolidToWalkOn(context, x, y-1, z, context.get(x, y-1, z));
             frostWalker = standingOnABlock && MovementHelper.canUseFrostWalker(context, destWalkOn);
             if (!MovementHelper.canWalkOn(context.bsi, destX, y - 1, destZ, destWalkOn) && !frostWalker) {
                 descend = true;
--- a/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
+++ b/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
@@ -114,6 +114,7 @@ public class MovementDiagonal extends Movement {
             return;
         }
         IBlockState destInto = context.get(destX, y, destZ);
+        IBlockState fromDown = context.get(x, y - 1, z);
         boolean ascend = false;
         IBlockState destWalkOn;
         boolean descend = false;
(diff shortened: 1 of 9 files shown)


src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -125,17 +126,22 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
             destWalkOn = destInto;
         } else {
             destWalkOn = context.get(destX, y - 1, destZ);
-            if (!MovementHelper.canWalkOn(context, destX, y - 1, destZ, destWalkOn)) {
+            boolean standingOnABlock = MovementHelper.mustBeSolidToWalkOn(context, x, y-1, z, context.get(x, y-1, z));
```
leijurv, line 129: run autoformatter

leijurv, line 129: let's save a BSI lookup call by caching the block at `(x, y - 1, z)`

https://github.com/cabaletta/baritone/blob/f76283ebfeb5bba7d53d76f727335de665131951/src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java#L148

it's saved as `fromDown` a little bit later on, let's move that up to something like `IBlockState fromDownState = context.get(x, y - 1, z);`, which could go maybe right after `boolean frostWalker = false;`, then use that in both places?

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -125,17 +126,22 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
...
+            frostWalker = standingOnABlock && MovementHelper.canUseFrostWalker(context, destWalkOn);
+            if (!MovementHelper.canWalkOn(context, destX, y - 1, destZ, destWalkOn) && !frostWalker) {
```
leijurv, line 131: move the `!frostWalker` to first so that it short circuits on the boolean

(this is not a big deal because we already have the IBlockState so this is just going to PrecomputedData which is no biggie but might as well)

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -265,7 +276,12 @@ public MovementState updateState(MovementState state) {
             }
         } else if (!ctx.playerFeet().equals(src)) {
             if (ctx.playerFeet().equals(src.offset(direction)) || ctx.player().posY - src.y > 0.0001) {
-                if (!MovementHelper.canWalkOn(ctx, dest.down()) && !ctx.player().onGround && MovementHelper.attemptToPlaceABlock(state, baritone, dest.down(), true, false) == PlaceResult.READY_TO_PLACE) {
+                if (Baritone.settings().allowPlace.value
```
leijurv, line 279: I don't know about this addition - if `allowPlace` is false we need to bail out MUCH earlier than here. At this point we're **in midair** and trying to parkour place. It would be unsafe and could kill the player to attempt a parkour place but then refuse to actually place it because of this setting having the wrong value.

src/main/java/baritone/pathing/movement/movements/MovementPillar.java
```diff
@@ -101,6 +101,10 @@ public static double cost(CalculationContext context, int x, int y, int z) {
             // if we're standing on water and assumeWalkOnWater is false, we must have ascended to here, or sneak backplaced, so it is possible to pillar again
             return COST_INF;
         }
+        if (fromDown.getBlock() instanceof BlockLiquid && (from == Blocks.WATERLILY || from == Blocks.CARPET)) {
```
leijurv, line 104: i guess let's move the `==` to be the first check because it's unlikely (lilypad and carpet are rare in the world, water is common), and because `==` is faster than `instanceof`

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -72,7 +72,9 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
         IBlockState pb1 = context.get(destX, y, destZ);
         IBlockState destOn = context.get(destX, y - 1, destZ);
         Block srcDown = context.getBlock(x, y - 1, z);
-        if (MovementHelper.canWalkOn(context, destX, y - 1, destZ, destOn)) {//this is a walk, not a bridge
+        boolean standingOnABlock = MovementHelper.mustBeSolidToWalkOn(context, x, y-1, z, context.get(x, y-1, z));
```
leijurv, line 75: let's use `srcDown` (modified to be an IBlockState) to save this extra lokup

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -72,7 +72,9 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
...
+        boolean frostWalker = standingOnABlock && !context.assumeWalkOnWater && MovementHelper.canUseFrostWalker(context, destOn);
+        if (MovementHelper.canWalkOn(context, destX, y - 1, destZ, destOn) || frostWalker) { //this is a walk, not a bridge
```
leijurv, line 77: move `frostWalker` to left side of `||`

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -142,9 +146,13 @@ public static double cost(CalculationContext context, int x, int y, int z, int d
                 if (srcDown == Blocks.SOUL_SAND || (srcDown instanceof BlockSlab && !((BlockSlab) srcDown).isDouble())) {
                     return COST_INF; // can't sneak and backplace against soul sand or half slabs (regardless of whether it's top half or bottom half) =/
                 }
-                if (srcDown == Blocks.FLOWING_WATER || srcDown == Blocks.WATER) {
+                if (!standingOnABlock) { // standing on water / swimming
                     return COST_INF; // this is obviously impossible
                 }
+                Block blockSrc = context.getBlock(x, y, z);
+                if (srcDown instanceof BlockLiquid && (blockSrc == Blocks.WATERLILY || blockSrc == Blocks.CARPET)) {
```
leijurv, line 153: swap order again (lilypad or carpet first)

src/main/java/baritone/pathing/path/PathExecutor.java
```diff
@@ -380,6 +380,21 @@ private boolean shouldSprintNextTick() {
...
 
+            if (pathPosition < path.length() - 2) {
+                // keep this out of onTick, even if that means a tick of delay before it has an effect
+                IMovement next = path.movements().get(pathPosition + 1);
+                if (MovementHelper.canUseFrostWalker(ctx, next.getDest().down())) {
+                    // frostwalker only works if you cross the edge of the block on ground so in some cases we may not overshoot
+                    // Since MovementDescend can't know the next movement we have to tell it
+                    if (next instanceof MovementTraverse || next instanceof MovementParkour) {
+                        boolean couldPlaceInstead = Baritone.settings().allowPlace.value && behavior.baritone.getInventoryBehavior().hasGenericThrowaway() && next instanceof MovementParkour; // traverse doesn't react fast enough
+                        boolean sameFlatDirection = current.getDirection().up().crossProduct(next.getDirection()).equals(BlockPos.ORIGIN); // here's why you learn maths in school
```
leijurv, line 391: so what is this cross product actually doing? `a.crossProduct(b).equals(ORIGIN)` does what exactly? checks if a and b are the same? checks if they are in the same plane but of different magnitude?

leijurv:
i assume this fixes the issue where if you have frostwalker boots on, baritone totally screws up all paths involving water?

ZacSharp:
Yep.
It's pretty much borderline on what A* in Z³ can handle though so there's degradations if `assumeWalkOnWater` is used alongside frostwalker (because when planning to stand water there's no way to tell which way of doing so will be used).

Will work on this tomorrow (and if I don't feel free to give my stupid brain an @ping reminder).
There's two more things I noticed while writing this (note to me: it's not backplacing against water `assumeWalkOnWater` and making sure `MovementFall` bails out early enough)

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
ZacSharp: We don't attempt to parkour place when `allowPlace` is false. This new check is there to reject placing when the block we are going to land on is not yet turned into water (but frostwalker can fail so I can't just say this is not a problem).

A straight traverse-traverse-descend-parkour chain has a 0.7% failure rate when checking `allowPlace` and a 19.9% chance of doing a forbidden placement without checking `allowPlace`. Doing straight descend-traverse-traverse-parkour instead yields 0.2% failure rate and 19.4% forbidden placements instead (test sizes are 1000-1200 runs each and I also have 50-70 runs each on recording if you want to see them, they are just too big to put them into this comment right away)

If you give me a nice name for it (maybe one fitting for #1586?) I will look into allowing it to place even with `allowPlace` disabled (delayed to reduce false positives).

ZacSharp:
Also forgot to mention that this has the same risk of self-intersecting paths as other movements modifying the world. It's just a little bit more likely due to the higher impact radius.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
leijurv: What's an example of such a forbidden placement? Like, when you're doing a parkour jump to ice, but the ice melts while you're in midair? How exactly does this happen?

leijurv: I can't visualize what you mean by `This new check is there to reject placing when the block we are going to land on is not yet turned into water`

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
ZacSharp: Maybe move it into the if-else part like `destWalkOn` so we don't look it up when taking the early return?

src/main/java/baritone/pathing/path/PathExecutor.java
ZacSharp: It's there to check whether the two movements are in the same horizontal direction and have a height difference of one (because descend always goes down and we are not interested in parkour ascend, only horizontal would be `.dot(UP) == 0` instead of `.equals(ORIGIN)`).
Theoretically there's also `a.crossProduct(a.scale(-n)).equals(ORIGIN)`, but you can't jump in the direction you came from (because you would jump through the `src` of the descend).

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
ZacSharp: You are @leijurv:leijurv.com on matrix right? Can I just message you an 105MB zip file (all four) or a 40MB mp4 (only first)?

ZacSharp: Forbidden placement it just placing even though it is neither planned nor allowed.

ZacSharp:
Last known problem besides review comments: When water is replaced by a solid block it becomes `minecraft:flowing_water` for a tick, which is enough for `PathExecutor` to cancel the path because a future traverse has become impossible (because it cannot walk on flowing water and bridging on water is not allowed when `assumeWalkOnWater` is enabled).

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
leijurv: Yeah, that's cool

src/main/java/baritone/pathing/path/PathExecutor.java
leijurv: Gotcha. `current.getDirection()` might be `(+n, -1, 0)` and `next.getDirection()` might be `(1, 0, 0)`, then `sameFlatDirection` would be true because `current.getDirection().up()` would be `(+n, 0, 0)`, and the cross product of `(+n, 0, 0)` and `(1, 0, 0)` will be `(0, 0, 0)` because both of those vectors are in the exact same direction and don't form a plane, thus the cross product is all zero?

Please add a comment explaining all that 😉

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
leijurv: Sure

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
leijurv: Although it might need to be in separate files since I have a vague memory that my shitty nginx has a 50mb upload limit lol

src/main/java/baritone/pathing/path/PathExecutor.java
ZacSharp: Actually traverse can go through `current.getSrc().down()` if `allowDownward` is or it has to place `current.getSrc().down(2)`. I've added a check that should make this work with whatever crazy movement the future might throw at it.

ZacSharp: Actually only Vec3d has the required method. Will force push a less universal check for `MovementTraverse` only.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
ZacSharp: Doesn't seem like I can message you. I can't even message myself on a different homeserver.

leijurv: Ok. Feel free to share the mp4 by any other means - youtube, dropbox, google drive, wetransfer.com. Actually now I remember you refuse to use discord so maybe you are one of that kind 😉 of person 😉, you might want a FOSS solution? Maybe try just an account on the matrix.org homeserver using app.element.io? I have one of those, it's `@leijurv:matrix.org`. Might work?

leijurv: Also maybe try again now, I just updated my synapse. I bet it was YEARS out of date I literally can't remember the last time I messed with it. Probably not updated since late 2021 lol

leijurv: i cant see any messages from you in any of the three rooms :(

ZacSharp: 3? Only saw two so far.
I can see your messages but mine just say "didn't send"

ZacSharp: Found the third one but:
image

leijurv: Hm, even the most recent where it's just both of us `@matrix.org`? cringe foss software. propietary software is superior lol

leijurv: maybe can you just find some place to upload the file that you're comfortable with, then paste a link here?

ZacSharp: Fallback would have been thrashing my uploads repo (which is probably already bigger than baritone)

ZacSharp: Of what (or whom?) exactly are you making fun here?

ZacSharp: Whoops. Just reread this conversation noticed that `This new check is there to reject placing when the block we are going to land on is not yet turned into water` should have been `This new check is there to reject placing when the block we are going to land on is not yet turned into ice` instead. Might have caused a good part of the conversation above.

PR #3785, opened 2023-01-09
buildonlyselection optimization

Entropy5:
buildinlayers ~ buildonlyselection lag fix for big schematics (skips layers more properly)
uwu


ZacSharp (commented):
Not merging this up yet because testing on 1.13.2 gave me this 9MB log file
[latest.log](https://github.com/cabaletta/baritone/files/10377994/latest.log)

Skipping up layers isn't guaranteed to hit the top layer.

Entropy5:
i tested something close to this exact case, but the issue seems to only trigger with a schematic that has an odd height, attempting to fix rn

PR #3786, opened 2023-01-10
bugfix for odd height schematics

Entropy5:
https://github.com/cabaletta/baritone/pull/3785#pullrequestreview-1241389460
@ZacSharp

### DIFF
--- a/src/main/java/baritone/process/BuilderProcess.java
+++ b/src/main/java/baritone/process/BuilderProcess.java
@@ -84,7 +84,7 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
     private int layer;
     private int numRepeats;
     private List<IBlockState> approxPlaceable;
-    public final int[] selectionYBounds = {0, 0};
+    public final int[] selectionLayerBounds = {0, 0};
 
     public BuilderProcess(Baritone baritone) {
         super(baritone);
@@ -123,10 +123,10 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
                 logDirect(String.format("Schematic starts at y=%s with height %s", y, schematic.heightY()));
                 logDirect(String.format("Selection starts at y=%s and ends at y=%s", minim.isPresent() ? minim.getAsInt() : 0, maxim.isPresent() ? maxim.getAsInt() : 0));
                 if (minim.isPresent() && maxim.isPresent()) {
-                    selectionYBounds[0] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - maxim.getAsInt() : minim.getAsInt() - y;
-                    selectionYBounds[1] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - minim.getAsInt() : maxim.getAsInt() - y;
+                    selectionLayerBounds[0] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - maxim.getAsInt() : minim.getAsInt() - y;
+                    selectionLayerBounds[1] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - minim.getAsInt() : maxim.getAsInt() - y;
                 }
-                this.layer = this.selectionYBounds[0] / Baritone.settings().layerHeight.value;
+                this.layer = this.selectionLayerBounds[0] / Baritone.settings().layerHeight.value;
                 logDirect("Skipped everything under layer " + this.layer);
             }
         }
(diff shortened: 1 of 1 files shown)


leijurv:
how does adding 1 after the division help? also why is this only a problem for odd height? wouldn't it be a problem for anything not evenly divisible by `layerHeight`? what if `layerHeight` isn't two?

leijurv:
(obviously i believe you that this fixes at least some combinations of schematic height and layerheight i'm just skeptical that it covers all cases)

ZacSharp:
I can confirm it doesn't freak out anymore, but I'm very much not convinced it is working as intended.

ZacSharp:
Adding one always works because `(floor(height / layerHeight) + 1) * layerHeight >= height`.

However there's two major concerns with the overall functionality
1. It applies even when `buildOnlySelection` doesn't apply. (because that is applier by methods calling the method you placed this optimization in)
2. If I build two schematics in a row, one with selections made and the second without, then the selection bounds from the first one are reused for the second one. You have to make sure to always set all state when building starts.

Also why not replace `int[] selectionLayerBounds` with a simple `int stopAtHeight`? We don't need the starting layer again and it's much easier to understand and easier to check against (you can probably just replace `layer * Baritone.settings().layerHeight.value < realSchematic.heightY()` with `layer * Baritone.settings().layerHeight.value < stopAtHeight` and be done without any helper methods)

Entropy5:
@ZacSharp I have implemented and tested some of the edits you suggested, but am not sure what you mean with your second point.
If someone switches between schematics then I do expect them to either redo their selections, or have it cycle through a couple extra layers. This modification is meant to only have a big impact when building huge schematics, where the game would previously freeze for several seconds before arriving at the first layer within a selection.

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -113,21 +113,24 @@ public void build(String name, ISchematic schematic, Vec3i origin) {
...
-                    selectionYBounds[0] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - maxim.getAsInt() : minim.getAsInt() - y;
-                    selectionYBounds[1] = Baritone.settings().layerOrder.value ? y + schematic.heightY() - minim.getAsInt() : maxim.getAsInt() - y;
+                    int startAtHeight = Baritone.settings().layerOrder.value ? y + schematic.heightY() - maxim.getAsInt() : minim.getAsInt() - y;
+                    this.stopAtHeight = (Baritone.settings().layerOrder.value ? y + schematic.heightY() - minim.getAsInt() : maxim.getAsInt() - y) + 1;
+                    this.layer = startAtHeight / Baritone.settings().layerHeight.value;
+                    if (Baritone.settings().chatDebug.value) {
+                        logDirect(String.format("Schematic starts at y=%s with height %s", y, schematic.heightY()));
+                        logDirect(String.format("Selection starts at y=%s and ends at y=%s", minim.getAsInt(), maxim.getAsInt()));
+                        logDirect(String.format("Considering relevant height %s - %s", startAtHeight, this.stopAtHeight));
+                    }
```
ZacSharp, line 132: There's also `logDebug`, which does the same without manually checking `chatDebug`.

ZacSharp (commented):
Point two is now fixed. The problem was that if both `buildInLayers` and `buildOnlySelection` were enabled but there were no selections is used whatever values were stored in `selectionYBounds`, e.g. those of a previously built schematic and that was leading to confusing behavior.

However point one is still valid: if I use this to e.g. paste a selection `buildOnlySelection` is not applied (and is not supposed to be applied) but this logic can still skip layers. You have to make sure the skipping only happens when the setting is actually applied (i.e. the schematic is wrapped in a `SelectionSchematic`) in one of the other build methods (i.e. `build(String,File,Vec3i)` and `buildOpenSchematic()`.

Also make sure to not break the `startAtLayer` setting. Currently if I were to set `startAtLayer` to something above the selection bounds this optimization will force the starting layer back down to the selection bounds.

Even though I don't really like it I think the easiest solution to both might be to move out the layer skipping logic into a helper method and call that after the call to `build(String,ISchematic,Vec3i)` if needed. In the helper method you can then increase `layer` and set `stopAtHeight` as needed (you still need to unconditionally set `stopAtHeight` in `build(String,ISchematic,Vec3i)` to prevent point 2) from coming back)

Entropy5:
I confirmed the unwanted behavior with `sel paste`
This is not the fix you suggested but it seems to work...
StartAtLayer is now considered too

ZacSharp (approved):
`SelectionSchematic` is not public API (i.e. we know if one is passed in it's us applying `buildOnlySelection`) so this is fine.

Haven't done a final test yet but I'll just approve already.

PR #3798, opened 2023-01-15
Fix canWalkThrough caching

ZacSharp:
Fixes https://github.com/cabaletta/baritone/pull/3771#issuecomment-1382950808

<details><summary>Before</summary>

</details>

<details><summary>After</summary>

</details>

(Screenshots are obviously not from 1.12.2)

### DIFF
--- a/src/main/java/baritone/pathing/precompute/PrecomputedData.java
+++ b/src/main/java/baritone/pathing/precompute/PrecomputedData.java
@@ -52,7 +52,7 @@ public class PrecomputedData {
         if (canWalkThroughState == YES) {
             blockData |= CAN_WALK_THROUGH_MASK;
         }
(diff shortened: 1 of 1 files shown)


leijurv:
Screenshots not from 1.12 but it's a problem in 1.12 too, yes?

ZacSharp:
I knew an example on newer versions and the code was obviously wrong so I just took what I had.

Took some time to dig up an example on 1.12.2 (vanilla) and it turns out that only low snow layers were affected.
Contrary to considering all waterlogged blocks instamineable (and thinking they would become air), this one is pretty subtle and merely increases the cost of walking through snow layers by a bit (break cost + breaking penalty).

Screenshots for completeness:
<details><summary>Before</summary>


</details>
<details><summary>After</summary>


</details>

PR #3801, opened 2023-01-17
Merge master into 1.16.5

ZacSharp:
Did a last second decision to try returning correct values from `mustBeSolidToStandOn` when `assumeWalkOnWater` is enabled (should only be relevant when combined with frostwalker,) so this took a bit longer than anticipated (whole bunch of trying changes and stuff versus 4 conflict free merges).


ZacSharp:
Whoops. Wrong branch.

PR #3836, opened 2023-02-12
allow to only follow certain items

rycbar0:


rycbar0:
the other one is better

PR #3855, opened 2023-02-27
Deobf build task to allow mods building against baritone

rfresh2:
Mods like Lambda (and previously kami blue) that bundle Baritone have been [stuck on 1.2.15 for years](https://github.com/lambda-client/lambda/blob/7b505a34362e31c91b6f53ce07e355091a918598/build.gradle#L111-L115). The baritone in `ImpactDevelopment/maven` [is also still on 1.2.15](https://github.com/ImpactDevelopment/maven/tree/master/cabaletta/baritone-api/1.2).

If mods build against [com.github.cabaletta.baritone:baritone-api:v1.2.17](https://jitpack.io/#cabaletta/baritone/v1.2.17), Baritone's API classes reference obfuscated MC classes which need to not be obf to reference against at compile steps of the build. 

This PR adds an explicit `deobf` jar that baritone builds. 

[Example of how this PR can be used by dependent mods](https://github.com/rfresh2/lambda/commit/5d9c9e11694c0eb8e0fdcd0f4507f7a74baff8aa)

Definitely open to alternate solutions, I opted not to touch existing baritone artifacts to not break consumers that depend on them as they are.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,23 +144,32 @@ task proguard(type: ProguardTask) {
 
 task createDist(type: CreateDistTask, dependsOn: proguard)
 
+task deobfJar(type: Jar) {
+    from sourceSets.api.output, sourceSets.main.output, sourceSets.launch.output
+    baseName += "-deobf"
+}
+
+build.dependsOn(deobfJar)
 build.finalizedBy(createDist)
 
 install {
     def jarApiName = String.format("%s-api-%s", rootProject.name, version.toString())
     def jarApiForgeName = String.format("%s-api-forge-%s", rootProject.name, version.toString())
+    def jarApiDeobfName = String.format("%s-deobf-%s", rootProject.name, version.toString())
     def jarSAName = String.format("%s-standalone-%s", rootProject.name, version.toString())
     def jarSAForgeName = String.format("%s-standalone-forge-%s", rootProject.name, version.toString())
 
     artifacts {
         archives file("$buildDir/libs/"+jarApiName+".jar")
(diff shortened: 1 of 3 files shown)


wagyourtail:
the real solution is to get a srg obf build (`baritone-forge-unoptimized`) instead of the notch ones. as `fg.deobf` can then work and won't have mapping issues if people use different mappings

ZacSharp:
In that case we probably have to rename `BetterBlockPos.x`, `BetterBlockPos.y` and `BetterBlockPos.z` to work around https://github.com/md-5/SpecialSource/issues/72 (or whatever else is incorrectly and incompletely remapping our `public final` fields).

leijurv:
so the idea is that you dev against deobf then when actually releasing your jar it would work against api?

what mappings would deobf use though? would this only solve the problem for certain mappings like mcp? like this wouldn't work for a fabric mod developer for new versions of Mc right? they'd need a -deobf-fabric? or could they just use -fabric?

leijurv:
i put this version up on the maven here https://github.com/ImpactDevelopment/maven/commit/155f71376555ee970f6a66adf5656e770262851d

ZacSharp:
> what mappings would deobf use though? would this only solve the problem for certain mappings like mcp? like this wouldn't work for a fabric mod developer for new versions of Mc right? they'd need a -deobf-fabric? or could they just use -fabric?

It would only solve the problem for projects using the same mappings as Baritone (loader independent but fabric mods usually don't use mcp mappings).
See #3856 for how to do it.

PR #3856, opened 2023-02-28
backport current buildscripts to 1.12.2

wagyourtail:
Backports current buildscripts to 1.12.2, and adds legacy fabric support... 
(legacy fabric doesn't launch in dev due to not being able to find minecraft when it's mcp mapped -_-)

So basically, no more `setupDecompWorkspace`, and stuff is under `unimined` instead of `forgegradle`, `genIntellijRuns` may require you to set the java version to 8 afterwards for launching in dev if you have 17 set as the sdk in idea... 

## TODO
- [ ] update setup.md


wagyourtail:
when porting up, until 1.17, no build files should change except gradle.properties, so if something changes one of these files it'll probably be wrong... also ci only needs gradle build not the ones for forge/fabric now.

at 1.17, there will be a mapping change to official mojang mappings...
this uses `mappings "net.minecraft:minecraft:${project.minecraft_version}:client-mappings"`

leijurv:
will this work easily with https://github.com/cabaletta/baritone/pull/3855? such as for https://github.com/ImpactDevelopment/maven/commit/155f71376555ee970f6a66adf5656e770262851d as used by https://github.com/lambda-client/lambda/commit/90d557540bb29e759785d484669a30d91d493cb3

currently i can just do `./gradlew jar` to get a mcp deobf jar, is that still the case?

leijurv:
> when porting up, until 1.17, no build files should change

sounds like a great use case for the git fake merge (example: `git reset --hard $(git commit-tree 1.13.2^{tree} -p 1.13.2 -p master -m fake merge of master into 1.13.2")`) :trollface:

wagyourtail:
> will this work easily with https://github.com/cabaletta/baritone/pull/3855? such as for https://github.com/ImpactDevelopment/maven/commit/155f71376555ee970f6a66adf5656e770262851d as used by https://github.com/lambda-client/lambda/commit/90d557540bb29e759785d484669a30d91d493cb3
> 

this is set up for maven publish to publish artifacts similarly to arch-loom's output... also srg is the proper target so they dont need the exact same mcp mappings and fg.deobf will work. maven publish local it and see the output... but yes they should be able to handle the outputs

fabric/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,31 @@
+
+{
+  "schemaVersion": 1,
+  "id": "baritone",
```
ZacSharp, line 4: Can we maybe either use "Baritone" or "Baritoe" for all platforms instead of Forge -> "Baritoe" and Fabric -> "Baritone"?

forge/build.gradle
```diff
@@ -0,0 +1,132 @@
...
+        mixinConfig = ["mixins.baritone.json"]
+    }
+
+    launches.client = {
+        args.addAll(["--tweakClass", "org.spongepowered.asm.launch.MixinTweaker"])
+        for (int i = 0; i < args.size(); i++) {
+            if (args[i] == "-mixin.config") {
+                args[i] = "--mixin"
+            }
+        }
```
ZacSharp, line 38: An explanatory comment for why we do this would be nice.

ZacSharp (commented):
Dist jars have the appropriate mappings for their platform, `tweaker:runClient` and `fabric:runClient` seem to work, `forge:runClient` is broken and `:runClient` surprisingly existed and started fabric.

What's the purpose of the "-dev" and "-shadow" jars? E.g. the fabric "-dev" jar doesn't contain any classes.

Is there any way to run proguard less often? Running it six times per build isn't quite ideal.

ZacSharp:
> sounds like a great use case for the git fake merge (example: git reset --hard $(git commit-tree 1.13.2^{tree} -p 1.13.2 -p master -m "fake merge of master into 1.13.2")) :trollface:

There's other changes so I'll do a normal merge and then checkout the old file.
Also thanks @wagyourtail for the detailed update instructions.

> also srg is the proper target so they dont need the exact same mcp mappings and fg.deobf will work

just gotta repeat that. Mcp/yarn mapped is almost as bad as notch mapped because it forces you to use the same mappings version as Baritone.

<details><summary>Could just drop this into my libs folder and it worked</summary>


(I was using a srg mapped Baritone build all the time though so not the same situation as lambda)
</details>

wagyourtail:
> Dist jars have the appropriate mappings for their platform, `tweaker:runClient` and `fabric:runClient` seem to work, `forge:runClient` is broken and `:runClient` surprisingly existed and started fabric.

hmm, weird forge:runClient didn't... genIntellijRuns might make a working forge run... I'll have to look into that at some point tho.

> What's the purpose of the "-dev" and "-shadow" jars? E.g. the fabric "-dev" jar doesn't contain any classes.

-dev is just jaring the results of the current project, see in tweaker it'll just have the 2 classes defined there, and in forge same thing with the BaritoneForgeModXD class.
-shadow then shades in the "common" classes before remapping
and then finally it remaps to the final jar


> Is there any way to run proguard less often? Running it six times per build isn't quite ideal.

I'm not a proguard expert. if there's a way to overlap it maybe... but there are technically 6 distinct builds. We could probably run proguard once, export the mappings and then do the others with tiny remapper, but the other optimiazations wouldn't be achieved that way

build.gradle
```diff
@@ -15,97 +15,130 @@
...
-group 'baritone'
-version '1.2.17'
+allprojects {
+    apply plugin: 'java'
+    apply plugin: "xyz.wagyourtail.unimined"
+    apply plugin: "maven-publish"
+
+    archivesBaseName = rootProject.archives_base_name
+    def vers = 'git describe --always --tags --dirty'.execute().text.trim()
+    if (!vers.startsWith("v")) {
```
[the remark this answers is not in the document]
wagyourtail: there was another change in #3486 iirc, but I pulled it in at the beginning of the 1.19 cycle so we could get better error reports

buildSrc/src/main/java/baritone/gradle/task/ProguardTask.java
```diff
@@ -192,197 +197,88 @@ public boolean accept(File dir, String name) {
...
+        //TODO: fix for j16
+//        final JavaVersion javaVersion = new DefaultJvmVersionDetector(new DefaultExecActionFactory(new IdentityFileResolver())).getJavaVersion(java);
+//
+//        if (!javaVersion.getMajorVersion().equals("8")) {
+//            System.out.println("Failed to validate Java version " + javaVersion.toString() + " [" + java + "] for ProGuard libraryjars");
+//            // throw new RuntimeException("Java version incorrect: " + javaVersion.getMajorVersion() + " for " + java);
+//            return false;
+//        }
+//
+//        System.out.println("Validated Java version " + javaVersion.toString() + " [" + java + "] for ProGuard libraryjars");
```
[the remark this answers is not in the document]
wagyourtail: ignore, it doesn't matter, actually that's probably been resolved now that we've upgraded proguard, I also put a little if statement in place to make java8 and java17 both work at https://github.com/cabaletta/baritone/blob/7aa46a3031731233a2756c537f7c84f3d5109a25/buildSrc/src/main/java/baritone/gradle/task/ProguardTask.java#L221-L229

forge/build.gradle
wagyourtail:
```suggestion
        // old mixin (0.7.11) uses a different flag for specifying configs
        // in the launch args
        for (int i = 0; i < args.size(); i++) {
            if (args[i] == "-mixin.config") {
                args[i] = "--mixin"
            }
        }
```

build.gradle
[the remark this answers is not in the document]
ZacSharp: #3486 adds some extra logic for pull requests and to make CI use "+" instead of "-" as a separator.
And as of now it is broken because I can't use "'" and '"' properly.

fabric/src/main/resources/fabric.mod.json
wagyourtail: @leijurv why'd you let me merge fabric with baritone id back in 1.16
https://github.com/cabaletta/baritone/blob/1e7e504650222dc24e82379f1e81fbbb3c793e3e/src/launch/resources/fabric.mod.json#L4

and which should we use now

leijurv: > why'd you let me
LOL gonna stop you there, you're assuming I read your PRs??? hahaha

go flip a coin, you pick. baritone or baritoe

wagyourtail:
```suggestion
  "id": "baritoe",
```

wagyourtail:
I rolled high on a d20 so "baritoe" it is

build.gradle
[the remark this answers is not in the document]
ZacSharp: As long as leijurv is fine with prs showing an artificial commit in their version (i.e. the reason why #3486 got bloated) I'm also fine with the current handling, except
1. We should use `--match "v*"` to filter version tags (like I suggested in #3486 but without the quotation mistake)
2. This doesn't work without git
```
* What went wrong:
A problem occurred evaluating root project 'baritone'.
> Cannot run program "gitnotinstalled": error=2
```

build.gradle
```diff
@@ -15,97 +15,130 @@
...
+        println "git doesn't appear to be installed!"
```
ZacSharp, line 26: Actually, why not try to suggest the thing I mean again? (And this time I applied it locally before commenting)
```suggestion
    def vers;
    try {
        vers = 'git describe --always --tags --match v* --dirty'.execute().text.trim()
    } catch (IOException ignored) {
        vers = ""
        println "git doesn't appear to be installed!"
    }
    if (vers.equals("")) {
        println "couldn't find version tag!"
```

For some reason trying to quote the glob pattern makes it not find anything. Seems like the quotes become part of the pattern.

build.gradle
```diff
@@ -15,97 +15,130 @@
...
+        println "using version number: " + rootProject.mod_version
+        version = rootProject.mod_version
```
ZacSharp, line 28: Here's another attempt (I deleted the previous comment, don't search for it), this time actually making use of the fact that we can still use the commit hash even if we didn't find a tag.
```suggestion
    def vers;
    try {
        vers = 'git describe --always --tags --match=v* --dirty'.execute().text.trim()
    } catch (IOException ignored) {
        vers = ""
        println "git doesn't appear to be installed!"
    }
    if (vers.equals("")) {
        println "Couldn't find version tag!"
        println "Using fallback version number: " + rootProject.mod_version
        version = rootProject.mod_version
    } else if (!vers.startsWith("v")) {
        println "Couldn't find version tag!"
        println "Using fallback version number: " + rootProject.mod_version
        version = rootProject.mod_version + "-g" + vers
```

ZacSharp:
<details><summary>The server thread crashes when joining a world on `tweaker:runClient`.</summary><pre>java.lang.IncompatibleClassChangeError: Class com.google.common.base.Functions$ForMapWithDefault does not implement the requested interface java.util.function.Function
        at net.minecraft.advancements.Advancement$Builder.resolveParent(Advancement.java:157) ~[Advancement$Builder.class:?]
        at net.minecraft.advancements.AdvancementList.loadAdvancements(AdvancementList.java:64) ~[AdvancementList.class:?]
        at net.minecraft.advancements.AdvancementManager.reload(AdvancementManager.java:74) ~[AdvancementManager.class:?]
        at net.minecraft.advancements.AdvancementManager.<init>(AdvancementManager.java:66) ~[AdvancementManager.class:?]
        at net.minecraft.world.WorldServer.init(WorldServer.java:164) ~[WorldServer.class:?]
        at net.minecraft.server.integrated.IntegratedServer.loadAllWorlds(IntegratedServer.java:98) ~[IntegratedServer.class:?]
        at net.minecraft.server.integrated.IntegratedServer.init(IntegratedServer.java:130) ~[IntegratedServer.class:?]
        at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:434) [MinecraftServer.class:?]
        at java.lang.Thread.run(Thread.java:750) [?:1.8.0_362]</pre></details>

wagyourtail:
damn, thought I'd fixed that bug.... I'll look into it

wagyourtail:
> The server thread crashes when joining a world on `tweaker:runClient`.

fixed

ZacSharp (approved):
<details><summary>Not sure anymore whether this was the error from <code>forge:runClient</code> all the time so I'll include it here just in case.</summary><pre>[23:35:26] [main/ERROR] [LaunchWrapper]: Unable to launch
java.lang.reflect.InvocationTargetException: null
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_362]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_362]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_362]
        at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_362]
        at net.minecraft.launchwrapper.Launch.launch(Launch.java:135) [launchwrapper-1.12.jar:?]
        at net.minecraft.launchwrapper.Launch.main(Launch.java:28) [launchwrapper-1.12.jar:?]
Caused by: joptsimple.MissingRequiredOptionsException: Missing required option(s) [accessToken, version]
        at joptsimple.OptionParser.ensureRequiredOptions(OptionParser.java:426) ~[OptionParser.class:?]
        at joptsimple.OptionParser.parse(OptionParser.java:400) ~[OptionParser.class:?]
        at net.minecraft.client.main.Main.main(Main.java:51) ~[Main.class:?]
        ... 6 more</pre></details>

`tweaker:runClient` works now so everything which used to work is working and I can't complain :smile:
(I did test a built jar as well)

wagyourtail:
ok, well that's also fixed now... (the task from genIntellijRuns worked before)

ZacSharp:
On the idea of running proguard only once:
If remapping is faster than proguard we could first build everything, put it into one big jar, run it through proguard together and then split the loader specific parts back out and remap the resulting 9 jars to notch/srg/intermediary.
Apart from this being a weird way of ordering things the only problem I see so far is Forge modifying the Minecraft code so we might get problems providing a minecraft library to proguard (because there's two) or convincing proguard everything is indeed fine.
The loader specific parts would need to be contained within their own package and proguard must not move code across those package boundaries, because otherwise we can't cleanly separate the code for the different loaders afterwards.

wagyourtail:
since we don't depend on any forge patches, the vanilla jar would be fine...

ZacSharp:
Since the loader specific parts depend on main and api we have to run them through proguard, but explicitly marking them as kept should have the required effect.
If we want to not run them through proguard at all (and hence not merge them) we need a new internal api package / source set which is never obfuscated and used as a dependency by both main and the loader specific parts.

PR #3904, opened 2023-04-10
Add Patrol Command and Goal

rycbar0:
The new goal type allows to have baritone path to multiple points of interest in succession. There is also a new command to use with the new goal type. The waypoint command got a update to add a waypoint to a patrol route.

### DIFF
--- a/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
@@ -67,7 +67,7 @@ public class GoalPatrol implements Goal{
     }
 
     private boolean getNextOrReturnTrue() {
-        if(mode == Mode.ONEWAY && index == waypoints.size()-1) {
+        if(mode == Mode.ONEWAY && index == waypoints.size()-1 || waypoints.size() == 1) { //if we move oneway and are at the last waypoint or if we only have 1 waypoint we stop
             return true;
         } else {
             switch (mode) {
--- a/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
@@ -20,15 +20,12 @@ package baritone.api.pathing.goals;
 import baritone.api.BaritoneAPI;
 import baritone.api.IBaritone;
 import baritone.api.process.ICustomGoalProcess;
-import baritone.api.utils.Helper;
+import baritone.api.utils.interfaces.IGoalRenderPos;
 import net.minecraft.util.math.BlockPos;
 
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 
-public class GoalPatrol implements Goal{
+public class GoalPatrol implements Goal, IGoalRenderPos {
 
     private List<Goal> waypoints;
     private Mode mode;
@@ -38,13 +35,6 @@ public class GoalPatrol implements Goal{
         this.waypoints = waypoints;
         this.mode = mode;
         this.index = index;
-    }
-    public GoalPatrol(List<Goal> waypoints, Mode mode) {
-        new GoalPatrol(waypoints, mode, 0);
-    }
-
-    public GoalPatrol(List<Goal> waypoints) {
-        new GoalPatrol(waypoints, Mode.ONEWAY, 0);
     }
 
     @Override
@@ -67,19 +57,14 @@ public class GoalPatrol implements Goal{
     }
 
     private boolean getNextOrReturnTrue() {
-        if(mode == Mode.ONEWAY && index == waypoints.size()-1 || waypoints.size() == 1) { //if we move oneway and are at the last waypoint or if we only have 1 waypoint we stop
+        if(waypoints.size() == 1) {
             return true;
         } else {
             switch (mode) {
                 case ONEWAY:
-                    index++;
+                    waypoints.remove(0);
                     break;
                 case CIRCLE:
-                    /*if (index >= waypoints.size() - 1) {
-                        index = 0;
-                    } else {
-                        index++;
-                    }/**/
                     index++;
                     index = index >= waypoints.size() ? 0 : index;
                     break;
@@ -103,35 +88,11 @@ public class GoalPatrol implements Goal{
             return false;
         }
     }
-    public void setIndex(int Index) {
-        this.index = index;
-    }
-    public void setMode(String name) {
-        mode = Mode.getByNameOrDefault(name);
-    }
-
-    public Mode getMode() {
-        return mode;
-    }
-
-    public void setWaypoints(List<Goal> waypoints) {
-        this.waypoints = waypoints;
-    }
-
-    /*@Override
-    public boolean isInGoal(BlockPos pos) {
-        return Goal.super.isInGoal(pos);
-    }
 
     @Override
-    public double heuristic(BlockPos pos) {
-        return Goal.super.heuristic(pos);
+    public BlockPos getGoalPos() {
+        return ((GoalBlock)waypoints.get(index)).getGoalPos();
     }
-
-    @Override
-    public double heuristic() {
-        return Goal.super.heuristic();
-    }/**/
 
     public enum Mode {
         ONEWAY("oneway"),
@@ -164,15 +125,6 @@ public class GoalPatrol implements Goal{
             } else {
                 return "";
             }
-        }
-
-        public static String[] getAllNames() {
-            Set<String> names = new HashSet<>();
-            for (Mode mode : Mode.values()) {
-                names.addAll(Arrays.asList(mode.names));
-            }
-            String[] adfs = names.toArray(new String[0]);
-            return adfs;
         }
     }
 }
--- a/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
+++ b/src/api/java/baritone/api/pathing/goals/GoalPatrol.java
@@ -1,0 +1,178 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.api.pathing.goals;
+
+import baritone.api.BaritoneAPI;
+import baritone.api.IBaritone;
+import baritone.api.process.ICustomGoalProcess;
+import baritone.api.utils.Helper;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class GoalPatrol implements Goal{
+
+    private List<Goal> waypoints;
+    private Mode mode;
+    private int index;
+
+    public GoalPatrol(List<Goal> waypoints, Mode mode, int index) {
+        this.waypoints = waypoints;
+        this.mode = mode;
+        this.index = index;
+    }
+    public GoalPatrol(List<Goal> waypoints, Mode mode) {
+        new GoalPatrol(waypoints, mode, 0);
+    }
+
+    public GoalPatrol(List<Goal> waypoints) {
+        new GoalPatrol(waypoints, Mode.ONEWAY, 0);
+    }
+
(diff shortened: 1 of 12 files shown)


src/api/java/baritone/api/pathing/goals/GoalPatrol.java
```diff
@@ -0,0 +1,130 @@
...
+        BlockPos playerPos = BaritoneAPI.getProvider().getPrimaryBaritone().getPlayerContext().playerFeet();
+        if (playerPos.getX() == x && playerPos.getY() == y && playerPos.getZ() == z) {
+            if (inGoal) {
+                return getNextOrReturnTrue();
+            }
+            return false;
+        } else {
+            return inGoal;
+        }
+    }
```
ZacSharp, line 52: Sorry, to say it this harshly but NO.
1. I have concerns about thread safety (`playerFeet()` accesses the world)
2. This method is performance critical
3. `Goal`s are assumed to be stateless (and definitely should not depend on game state)

If you want to add patrolling you'll have to make some process responsible for changing the goal.

ZacSharp (changes requested):

src/api/java/baritone/api/pathing/goals/GoalPatrol.java
ZacSharp: I know (at least now, 27 minutes after submitting my previous review. Please forgive me) this pr is a draft, but since this is a pretty fundamental structural problem with this approach to patrolling I'll write it anyway.
If you want a changing goal you need some process to run the logic, both because goals are assumed to be stateless and because making them depend on game state like this can cause threading and performance problems.

wagyourtail:
Fix completed event status for arriving with goto. Add a process that wraps a command queue to other processes that sends the next when it gets the completed event. That sound about right for what the structure should be @ZacSharp

src/api/java/baritone/api/pathing/goals/GoalPatrol.java
leijurv: ✅  zacsharp is correct. all logic relating to patrolling should be in `PatrolProcess` or something similar. There shouldn't be any need to add a new subclass of `Goal` I don't think

wagyourtail:
so i phrased it in a way to implement it wider with the long ask for tesky chair for this end that we would have to have a new term child process for this new task aggregator process in order to have it wrap the process that's performing the current task in order to know when it

leijurv:
I don't think this is the right approach either. Commands and processes aren't the right abstraction to create a state machine. It would need to be a new class I think. Processes aren't meant to contain each other and they don't reasonably have duration, and commands having duration is also weird. A command isn't a flexible enough interface to create a good state machine, you want to be able to pass in arbitrary classes (such as other states) rather than just strings. Setting a wait in CustomGoalProcess breaks down the abstraction in an icky way.

Let me give an example from an internal spawnmason project, which uses a state machine to implement gringotts.

Each state can return either a new state, or a `PathingCommand`.

The wait state would have this as equivalent:

```
    public static class DelayState implements State {

        private final State then;
        private int ticks;

        // 0 = do it immediately
        // 1 = do nothing this tick, but do it next tick
        // 2 = do nothing this tick nor next, do it after that
        public DelayState(State then, int ticks) {
            this.ticks = ticks;
            this.then = then;
        }

        @Override
        public StateOrPath update() {
            ticks--;
            return ticks < 0 ? new StateOrPath(then) : STATIONARY;
        }
    }
```

Here are some more examples of states:

```
private static final StateOrPath STATIONARY = new StateOrPath(new PathingCommand(null, PathingCommandType.REQUEST_PAUSE));

...

    public class AwaitingEnderPearlThrowConfirmation implements State {

        int ticks;
        boolean gotCooldown;
        boolean gotEntity;

        @Override
        public StateOrPath update() {
            if (getPackets(SPacketCooldown.class).anyMatch(packet -> packet.getItem() instanceof ItemEnderPearl)) {
                gotCooldown = true;
                logDirect("Got ender pearl SPacketCooldown");
            }
            if (getPackets(SPacketSpawnObject.class).anyMatch(packet -> packet.getType() == 65)) {
                gotEntity = true;
                logDirect("Got ender pearl SPacketSpawnObject");
            }
            if (gotCooldown && gotEntity) {
                logDirect("Throw confirmed");
                return new StateOrPath(new SortState(true));
            }
            if (++ticks > 600) {
                logDirect("Retrying throw");
                return new StateOrPath(new SetPearlState());
            }
            return STATIONARY;
        }
    }

...

    public class GotoEnderChestState implements State {

        final EnderChestInteractionState onceOpened;
        final List<BlockPos> echests = new CalculationContext(baritone).worldData.getCachedWorld().getLocationsOf("ender_chest", Integer.MAX_VALUE, ctx.playerFeet().x, ctx.playerFeet().z, 2);

        public GotoEnderChestState(EnderChestInteractionState onceOpened) {
            this.onceOpened = onceOpened;
        }

        @Override
        public StateOrPath update() {
            Goal goal = new GoalComposite(echests.stream().map(GoalGetToBlock::new).toArray(Goal[]::new));
            if (goal.isInGoal(ctx.playerFeet()) || goal.isInGoal(baritone.getPathingBehavior().pathStart())) {
                for (BlockPos maybe : echests) {
                    Optional<Rotation> rot = RotationUtils.reachable(ctx, maybe, false);
                    if (rot.isPresent()) {
                        baritone.getLookBehavior().updateTarget(rot.get(), true);
                        if (ctx.isLookingAt(maybe)) {
                            return rightClickThen(new AwaitingEnderChestGuiOpenConfirmation(onceOpened, this, maybe));
                        }
                        return STATIONARY;
                    }
                }
                logDirect("I'm standing somewhere where I think I should be able to open a chest, but it doesn't seem to be reachable :(");
            }
            return new StateOrPath(new PathingCommand(goal, PathingCommandType.REVALIDATE_GOAL_AND_PATH));
        }
    }
```

ZacSharp:
<details><summary>A comment that ended up being my pen and paper for sorting my thoughts. Might contain inconsistencies and whatever. Not all of it reflects my current opinion anymore, but it would be a shame to just delete it. You don't need to read it.</summary>
I think we can do two things here
<ol>
    <li>Go back to the original scope of this pr and do just a goal queue. There's already functionality to check whether a goal is reached so we can just make <code>CustomGoalProcess</code> have a queue of goals or create a new <code>PatrolProcess</code> for more complicated goal switching logic (like switching randomly, not "logic" as in "programmable") and get away with it.</li>
    <li> Do more general task chaining. Without major efforts I'd do that as follows<details><ol>
   <li> Add a temporary <b>low priority</b> process which ignores <code>onLostControl</code> and is notified by <code>#cancel</code> like <code>BuilderProcess</code> is notified by <code>#resume</code>.</li>
   <li> Whenever that process is ticked it pops an instruction from it's queue and runs it. An instruction would generally be a wrapper around a <code>String</code> which is executed as a normal command.</li>
   <li> The hacky wait command would be avoided by having certain subcommands of the <code>#queue add</code> command like <code>#queue add wait <time></code> which look like the normal <code>#queue add</code> command but in reality they add a special instruction token to the command queue. This would effectively create commands which can only be used inside the queue.</li>
   <del><li>This could easily be expanded to a <em>very imperative</em> DSL with subroutines etc. with <code>CommandQueueProcess</code> turning into a line wise interpreter</li>
   <li> We could also use this with arbitrary java code by making the instruction class/interface accessible</li></del>
<li>The above points are struck through because I can't imagine them being nice to use. More like programming in Brainfuck.</li></ol></details>
I personally don't having a higher priority process wait on lower priority processes to finish or abandon their current task (processes don't have duration, but the things they do do have a duration) would be that bad, however I see how it's not really optimal since it's very much on the line of trying to establish delegation without the appropriate structure.
<details><summary>TLDR of this spoiler: I don't like some things</summary>My approach to establishing the required structure would probably have been creating some sort of <code>Task</code> which is essentially a short lived process made with delegation in mind, but that feels very similar to AltoClef and I don't like the way their tasks look and it would probably still suffer much of the same problems as the current processes (i.e. passing control back and forth in the presence of unsafe actions as mentioned in #3565). I also would try my best to prevent FSMs and their variations like HFSMs, they are either powerless or spaghetti. Doing behavior trees would end up being the <code>Task</code> approach I already mentioned with some additional headache for making it a behavior tree and just slapping some <code>onFinish</code> at the current processes doesn't feel like a good idea.</details><details><summary>Two things I like</summary>We could use <em>threading</em> to turn long behaviors into synchronous functions, composing behavior just like normal code (I have a ready to use synchronizer class to be used for that exact purpose for whatever I want to automate next) and switching between concurrent tasks by switching between "parallel" threads. Waiting for things you started would become as simple as waiting for the call to return and even <code>isSafeToCancel</code> would become easier to do: things which aren't safe to cancel would just catch the <code>CancellationException</code> thrown by <code>waitTick</code> until they are done. Forceful cancellation without memory leak would be impossible though and shared control at least hard to do because the framework we'd have to build on is already restrictive.
<br>However what leijurv showed also looks simple to write and use and would allow for easy composition both in the "command queue" case and for the creation of general behavior. It is also easily combinable with some changes to input handling I'm currently working on (ask if you are interested, this wall of text is too big already) and would maybe also make the "cancel once safe" problem easier to solve by caching states which aren't safe to cancel so they can run instead of the normal logic in the next tick. The situation for combining inputs is about the same as it currently is. Given that leijurv also seems to have positive experience with it and I just have hypothetics or negative experience with other options I think it's the best option we have for a new execution scheme</details>
<details><summary>Doubts whether we even need any of the above</summary>For a working command queue we would still need a way to determine whether the action started by a command is done though and the current processes are already enough to do a goal queue, so unless we're going to offer a complete new script-esque interface to the functionality of the existing commands I don't see much use in any type of new execution system. Third parties can (and do) already roll their own execution schemes so unless it offers e.g. better interaction with other Baritone based features there's no reason for them to use the new state machine and unless some of the things in the previous sentence change there's no reason for Baritone to use the new system. Even for things like finally implementing branch mining it would be simpler to just use an in-process state machine.
<br>However none of the above actually helps anything for command queues because those inherently have to run commands and regardless of how we try to set up a new system to allow nesting and complex behavior, if it runs arbitrary commands it ends up controlling itself in some sense.</details><details><summary>And finally an insight</summary>What we actually need to change is not the execution of behavior, it's the execution of commands. If a command is run, it must not activate some process, instead it must create a state / subprocess which will be returned to the caller. In the case of the player running a command from chat `ExampleBaritoneControl` will add that subprocess as a temporary (as in "will eventually cease to exist") top level task or set the state as the current state of the state machine and in the case of some process / state running the command it will start ticking it until it's done / return it as the next state (possibly wrapped in an <code>AndThen(state, this)</code>.</details>
</ol>

Also I hate markdown for it's poor nestability. Why can't I just have a list in a spoiler in a list and still use `inline code`?
</details>

The short version is: I don't think we need a new execution system for command queues but rather a means of executing commands "within a context", meaning that commands shouldn't go and set some global state to make Baritone do some thing. Instead they should return an object that, when executed by some means, makes Baritone do the thing. That way there's the clear caller -> callee relationship needed to have the command queue take care of it's children properly.
<details><summary>Based on the current process system it could look like this</summary>

```java
// the object would be more than just a BooleanSupplier, but we don't need more functionality for this example
// also note that I ingored the fact that a command might not return a thing to execute (e.g. #help)
// in QueueCommand.java
    public BooleanSupplier run(String label, IArgsConsumer args) {
        Stack<BooleanSupplier> taskList = // create task list, e.g. with code like in createRunCommandTask
        return () -> {
                    if (taskList.peek().get()) {
                        taskList.pop();
                    }
                    return taskList.isEmpty();
            };
        }
    }
    private BooleanSupplier createRunCommandTask(String command) {
        Lazy<BooleanSupplier> action = new Lazy(() -> baritone.getCommandManager().execute(command));
        return () -> action.get().get();
    }
// in ExampleBaritoneControl.java
    public void onSendChatMessage(ChatEvent event) {
        // parse command
        BooleanSupplier action = command.execute(label, args);
        baritone.getExecutorProcess().setTask(action); // could also add, create a new process, whatever
    }
```
</details>
<details><summary>And with a CPS executor maybe like this</summary>

```java
// note that I ingored the fact that a command might not return a thing to execute (e.g. #help)
// in QueueCommand.java
    public State run(String label, IArgsConsumer args, State andThen) {
        State state = // create task list, e.g. with RunCommandStates
        // we don't really need to decompose the list into a chain of states, we could also have a State continuously return to itself and run the next action
        return state;
    }
    private static class RunCommandState extends State {
        private final String command;

        public RunCommandState(String command, State andThen) {
            super(andThen);
            this.command = command;
        }

        // override annotation omitted because GitHub can't handle it
        StateOrPath update() {
            return new PathOrState(baritone.getCommandManager().execute(command, andThen));
        }
    }
// in ExampleBaritoneControl.java
    public void onSendChatMessage(ChatEvent event) {
        // parse command
        State state = command.execute(label, args, State.DONE);
        baritone.getStateExecutor().setCurrentState(state);
    }
// actually, we might not go for full CPS and instead have a ExecuteAndThen(first, second) state which
// runs its own state machine for the first state so we don't have to pass andThen literally everywhere
// the example leijurv provided wasn't pure CPS either
```
</details>

I don't think the state machine is needed for this pr, but I'd be happy about a pr introducing it (I could also write it if you want that).

rycbar0:
Im not able to do this.

PR #3908, opened 2023-04-11
Update Setup.md

wagyourtail:
it causes users to have issues building newer versions as the instructions change with 1.13+ and moreso with 1.17+, people really shouldn't be building anyway and should already understand the basics of forge (/fabric) modding before contributing to baritone as the way we build is fairly standard.


leijurv:
ok sure

PR #3923, opened 2023-04-24
✨ Add maxYLevelWhileMining setting

### LINKED ISSUE #1883: Parameter to limit Y level for automine (Non legit)

ZacSharp:
Behavior on 1.12.2 is obvious, but I'm not sure what to do when updating.
The option I prefer is simply bumping the default value to be the maximum maximum build height (Baritone doesn't work in CubicChunks worlds, right?), but for consistency with `minYLevelWhileMining` we'd need to also offset it by the minimum build height.
@wagyourtail why is `minYLevelWhileMining` offset by the minimum y anyway? Just so the default can stay at 0 or is there another reason?

closes #1883

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -854,6 +854,11 @@ public final class Settings {
     public final Setting<Integer> minYLevelWhileMining = new Setting<>(0);
 
     /**
+     * Sets the maximum y level to mine ores at.
+     */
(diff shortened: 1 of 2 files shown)


wagyourtail:
> Just so the default can stay at 0

  yeah pretty much

PR #3924, opened 2023-04-24
Extend BlockOptionalMeta parsing to parse block properties

### LINKED ISSUE #1760: Mining/Following with NBT
labels: enhancement

## Describe your suggestion

So basically allows baritone to mine a block with a specific nbt tag, e.g. 

#mine furnace[lit=true]

## Settings

#togglenbtmode
Toggles between the vanilla nbt tag syntax and an easier one based on entity nbt ("block" = placeholder for mc block):
mine block[nbt]
What also would work with the entity based syntax:

mine block[r=5]

mine block[r=..5]

And custom syntax:

mine block[facing=north]..

(This ".." at the end allows the use of an expandable selector: not only blocks with this nbt will be mined, also blocks with this and some other.)

Examples:

mine log:3[r=20,x=~3,y=3..5,z=7]

mine iron_door[facing=north]..

## Context

This would add more funcionality to the mine feature and make it better for specific uses, e.g. u placed a ton of doors in the wrong direction  near u, then u do: 

#mine door[r=..20,facing=north]

This could also enhance the follow command (and maybe more)

## Final checklist

ZacSharp:
Examples:
*  `#mine log[variant=oak]`
*  `#sel fill log[axis=y]` (if you build on a flat surface it won't even fail) 
*  `#sel replace wheat[age=7] air`
* `#goto chest[facing=north]`

Basically you can now specify any blockstate and it will ignore any property you didn't specify.

The only "breaking api change" is that the return value of `getMeta` is even less useful now, but I don't see a problem in that so here we go.

closes #1760 (even though the author said "nbt" instead of "block properties")

### DIFF
--- a/src/api/java/baritone/api/utils/BlockOptionalMeta.java
+++ b/src/api/java/baritone/api/utils/BlockOptionalMeta.java
@@ -19,6 +19,7 @@ package baritone.api.utils;
 
 import baritone.api.utils.accessor.IItemStack;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
 import net.minecraft.block.*;
 import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.IBlockState;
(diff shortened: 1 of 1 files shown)


pronery:
Can I Get a list of the block states??

wagyourtail:
is this consistent with how block properties work in commands on 1.13+?

ZacSharp:
For the most part that's the intention and from a quick read through `BlockPredicateArgument` and `BlockStateParser` it should implement the same logic.
However this does **not** mean `#sel` subcommands will behave like their `/command` counterparts since for Baritone all blockstate arguments mean "anything matching the given properties/meta" while for Minecraft it can also mean "the default state of the block after changing the specified properties" (that's `BlockStateArgument`). This makes sense since for e.g. `/fill log` there is no reason to place different states of log, but for Baritone it *does* make sense to place different states depending what it can actually place.

wagyourtail:
Ok

PR #3944, opened 2023-05-10
Allow blocking main for scanning

### LINKED ISSUE #3917: ArrayIndexOutOfBoundsException in FasterWorldScanner
## Some information
Operating system: Linux 5.15.0-47-generic
Java version: 1.8.0_362
Minecraft version: 1.12.2
Baritone version: originally v1.2.17 branched from ff1732011 (commit is https://github.com/ZacSharp/baritone/commit/03955d9276ec6fb0077885806ec13ec3c6b155a9 in case I push it). Reproduced on ff1732011 with changes specifically to trigger this.
Other mods (if used): None

## Exception, error or logs
[original log](https://github.com/cabaletta/baritone/files/11299878/latest.log)
[reproduction log](https://github.com/cabaletta/baritone/files/11299940/latest.log)

### Method 1 (How I found this, did not work again)
1. Run a random combination of commands which involve placing/breaking blocks
2. Run `#farm`

### Method 2 (How it happened again)
Apply
```diff
diff --git a/src/main/java/baritone/cache/FasterWorldScanner.java b/src/main/java/baritone/cache/FasterWorldScanner.java
index fbdc238ce..6f94981e4 100644
--- a/src/main/java/baritone/cache/FasterWorldScanner.java
+++ b/src/main/java/baritone/cache/FasterWorldScanner.java
@@ -192,6 +192,9 @@ private void visitSection(BlockOptionalMetaLookup lookup, ExtendedBlockStorage s
         if (isInFilter.length == 0) {
             return;
         }
+        try {
+            Thread.sleep(1000);
+        } catch (InterruptedException ignored) {}
 
         BitArray array = ((IBlockStateContainer) section.getData()).getStorage();
         long[] longArray = array.getBackingLongArray();
```
Then generate a world using the void preset, place some farmland, plant some wheat, give yourself some bonemeal and run `#farm`.
With this patch applied it also worked in my original test world, but only after manually moving the player around a bit and thus changing the block Baritone bonemeals, leaving the previous one in a partially grown state.

## My guess why it happens
Since `inFilter` had the wrong size and contructing it is separate from extracting the palette array and scanning happens off thread I suspect this happens when the palette is resized between the line directly above the if statement and the line directly after it. This is consistent with a `Thread.sleep` between them making the exception far easier to encounter and also fits nicely with the fact that adding some partially grown crops helped triggering the exception. I haven't printed out the numbers yet, though I expect to see the palette size crossing a power of two every time AIOOBE is thrown (will test this tomorrow).

## Modified settings
None (for method 2)

## Final checklist
- [x] I have not added any OwO's or UwU's to this issue.

ZacSharp:
I'm not happy with this but there's no progress anymore form my side so no point in waiting.
What I'd like to improve:
1. [x] Don't catch "concurrency errors" when main is actually blocked. I can't condition the catch on the setting because `IWorldScanner` itself actually has a sync api and the async part is just Baritone calling it from another thread. I also can't test whether the code is run on main because it runs on a thread pool which may or may not use main. 
2. [x] Maybe have a "maybe async executor" to reduce the amount of `if (<setting>) {<line>} else {<nearly identical line>}`? Where would that go? 
3. [ ] Name/documentation. This is a minor problem

closes #3917

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -907,6 +907,13 @@ public final class Settings {
     public final Setting<Boolean> extendCacheOnThreshold = new Setting<>(false);
 
     /**
+     * When enabled the main thread will not be used for world scanning.
+     * This trades correctness for performance so do not enable this unless you
+     * really need it and do not report bugs if it causes problems.
+     */
+    public final Setting<Boolean> scanAsynchronously = new Setting<>(false);
+
+    /**
      * Don't consider the next layer in builder until the current one is done
      */
     public final Setting<Boolean> buildInLayers = new Setting<>(false);
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -907,11 +907,13 @@ public final class Settings {
     public final Setting<Boolean> extendCacheOnThreshold = new Setting<>(false);
 
     /**
-     * When enabled the main thread will not be used for world scanning.
-     * This trades correctness for performance so do not enable this unless you
-     * really need it and do not report bugs if it causes problems.
+     * When enabled the main thread will not participate in world scanning.
+     * <p>
+     * Rarely causes a race condition so if Baritone doesn't find blocks anymore
+     * and your log is full of {@code ArrayIndexOutOfBoundsException}s you might
+     * want to try turning this off.
      */
-    public final Setting<Boolean> scanAsynchronously = new Setting<>(false);
+    public final Setting<Boolean> scanAsynchronously = new Setting<>(true);
 
     /**
      * Don't consider the next layer in builder until the current one is done
--- a/src/main/java/baritone/Baritone.java
+++ b/src/main/java/baritone/Baritone.java
@@ -235,4 +235,12 @@ public class Baritone implements IBaritone {
     public static Executor getExecutor() {
         return threadPool;
     }
+
+    public static void executeScan(Runnable func) {
+        if (settings().scanAsynchronously.value) {
+            getExecutor().execute(func);
+        } else {
+            func.run();
+        }
+    }
 }
--- a/src/main/java/baritone/cache/FasterWorldScanner.java
+++ b/src/main/java/baritone/cache/FasterWorldScanner.java
@@ -29,6 +29,7 @@ import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.ints.IntList;
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.client.Minecraft;
 import net.minecraft.network.PacketBuffer;
 import net.minecraft.util.BitArray;
 import net.minecraft.util.ObjectIntIdentityMap;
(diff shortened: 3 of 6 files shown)


wagyourtail:
minecraft has a isRunningOnMain function iirc, tou could also submit it to main with mc.execute or whatever its called when not on main

ZacSharp:
I don't want to use `mc.execute` (`mc.addScheduledTask` on 1.12.2) inside the scan function because we'd have to block the calling thread on main and if someone calls from another thread they likely wouldn't expect it to do that. If they for some reason decide to wait for their async scan to complete that makes a deadlock.

`mc.isRunningOnMain` is was I was thinking about in 1. but dismissed because the point where I would query it is not the one that actually matters. Querying in the exception handler would be "is main the thread that got the error" (never true) while the question I'm interested in is "is main the thread that started the scan" / "is main in the thread pool".
What I didn't think about an hour ago (and the last 2 days) is that I could query the thread before starting the parallel part and then pass the result through to the error handler. Not particularly nice but it should work.

leijurv:
looking through the world on other threads is sorta the whole idea of baritone's pathfinding. does it really cause issues on 1.12.2? i don't think i've ever seen this arrayindexoutofboundsexception

ZacSharp:
It does happen randomly (that's how I found it) but to be fair, without the `Thread.sleep` it should be very unlikely to happen and if I understood things right it can also only happen when a blockstate is added to a chunk section for the first time since that section was loaded.
My understanding might be incomplete though, since I can't really explain why when I first encountered this exception it happened again after canceling and restarting and neither why https://github.com/cabaletta/baritone/issues/3917#issuecomment-1518463507 happened.
I never had it happen again after opening this pr (i.e. while testing `#farm cocoa` for the merge up) and so far nobody left a comment or reaction on the issue so I guess that was a near-one timer.

Having the new scanner run while main is mutating the world is probably riskier than pathfinding or the old scanner because the time frame for trouble is some 2k times longer (rough guess based on section size) but apart from that I still think it should be pretty safe to turn this setting on by default. As long as no users complain, who cares about artificial tests? The intention behind this pr was being completely safe so it is totally possible to just do the slightly faster but theoretically buggy thing.
Having the switch would still be nice though so *if* someone reports this bug again we can just tell them how to turn the bug off..

leijurv:
Yeah it would be safe, but wouldn't it be extremely laggy? The reason why MineProcess only scans the world like 4 times a second rather than 20 was because that was the most my old laptop could handle on normal render distances. So, I'd be at like 5fps max if this happened on main game thread

ZacSharp:
It scans on main *and* whatever the jdk thinks is a good number of threads. Also `FasterWorldScanner` really is faster.

This video is recorded with a second Baritone building a cylinder in the background so lag is a bit higher than normal. I'd say that's why fps went up after I turned the setting off again.

https://github.com/cabaletta/baritone/assets/68165024/f78d1a7a-5720-4791-8cca-dc648f4f9ab3

If you want a video without questionable scheduling due to high background load, look at https://github.com/cabaletta/baritone/issues/3917#issuecomment-1537253741

ZacSharp:
This should now be obsolete

Issue #881, opened 2019-08-30
Scriptatone™ - Fast development of complex behaviors for Baritone


SIGSTACKFAULT:
Hook up a scripting language to Baritone, to allow "normal people" to make baritone do complicated things -- no compiling in new commands.

I think ([Jython](https://www.jython.org)) is the best choice, but ([luaj](https://github.com/luaj/luaj)) is also available.

No idea if this is an even vaguely good idea. Maybe it should be a wrapper or something.

0-x-2-2:
Scripting is a even worse then what we have already because the game is obfuscated and Baritone is re-mapped as compile time. Processes in a scripting language might cause performance issues depending if they use reflection or not. What do you expect people to create with a scripting API?

FriedrichDuerr:
What I meant with making it possible to inputs scripts it that also people that dont know much about any programming language have the chance to let baritone do more complicated things then walking somewhere. I do agree that processes are probably better suited but not everyone has the possibilities. Another argument is that writing a script would be faster then coding a process.

IronException:
What I meant with making it possible to inputs scripts it that also people that dont know much about any programming language have the chance to let baritone do more complicated things then walking somewhere. I do agree that processes are probably better suited but not everyone has the possibilities. Another argument is that writing a script would be faster then coding a process.

IronException:
the way I imagine it is that you can call the baritone processes in a row. So what baritone needs to do is look at what is the next thing to do and then lets that process run. If it is just that it would already help a lot of people.

Maybe if we want to go more complicated we could also let run different code when the process fails eg. Or other minecraft specific events... 
loops / goto might also come in handy.

0-x-2-2:
> What I meant with making it possible to inputs scripts it that also people that dont know much about any programming language have the chance to let baritone do more complicated things then walking somewhere. I do agree that processes are probably better suited but not everyone has the possibilities. Another argument is that writing a script would be faster then coding a process.

You would be coding a process but in a scripting language like js or lua,

ZeroMemes:
![image](https://user-images.githubusercontent.com/24485393/64060096-ceb05280-cb8d-11e9-9529-bdbbef4c7525.png)
lol

ZeroMemes:
NOT GONNA LIE
![image](https://user-images.githubusercontent.com/24485393/64060101-d8d25100-cb8d-11e9-9ed8-e3ae5e131185.png)

0-x-2-2:
[...]

PR #3952, opened 2023-05-18
Add "craft" command

strubium:
Welcome to the draft PR! 

This PR is going to:

- Use my limited knowledge of Java and the baritone API
- Have baritone path a crafting table
- Then, "process" items it may have from a set list 

Help + feedback would be appreciated, I do not know what I am doing.


rycbar0:
at the moment this is only a wrapper for the `#goto crafting_table` command

- if you want to add crafting to baritone you should also check the players inventory for a crafting table and try to place it instead. (unless the player already is next to one of curse)
- if you placed a crafting table you maybe also want to break and collect it after you finished crafting.
- this may be still changing but i would expect the command syntax to incoulde what you want to craft example `#craft stick 64` or `#craft stone_pickaxe` and then actually do the craftign.
- checking if you have enough resources to craft befor pathing to a crafting table could be usefull.

while what you have at the moment is working and fine idk if it would bring anything to baritone.

RemainingToast:
This seems similar to https://github.com/Meloweh/altoclef - It would cool to see a similar implementation into a ICraftProcessor

strubium:
> at the moment this is only a wrapper for the `#goto crafting_table` command
> 
> * if you want to add crafting to baritone you should also check the players inventory for a crafting table and try to place it instead. (unless the player already is next to one of curse)

Gonna Add 
> * if you placed a crafting table you maybe also want to break and collect it after you finished crafting.

Gonna Add^2 
> * this may be still changing but i would expect the command syntax to incoulde what you want to craft example `#craft stick 64` or `#craft stone_pickaxe` and then actually do the crafting.

Gonna Add^3
> * checking if you have enough resources to craft befor pathing to a crafting table could be useful.

Yes, perhaps a "gather" route could be added.

strubium:
Does anyone know the syntax for checking what blocks are in the inventory?

rycbar0:
not sure if this is the best or even the correct solution but i would say try this:
```
for (ItemStack invSlot : ctx.player().inventory.mainInventory) {
    if (invSlot.getItem().equals(Item.getItemFromBlock(Blocks.CRAFTING_TABLE))) {
        //we have found a crafting table.
    }
}```

RemainingToast:
> not sure if this is the best or even the correct solution but i would say try this:
> 
> ```
> for (ItemStack invSlot : ctx.player().inventory.mainInventory) {
>     if (invSlot.getItem().equals(Item.getItemFromBlock(Blocks.CRAFTING_TABLE))) {
>         //we have found a crafting table.
>     }
> }```
> ```

pretty sure this ignores hotbar

ZacSharp:
> pretty sure this ignores hotbar

Hotbar should be included as slots 0-8 (both ends inclusive) of the main inventory. Not sure about the offhand (and armor) though. Crafting slots are probably excluded here.

I think [this line](https://github.com/cabaletta/baritone/pull/3952/files#diff-911749a14802e036b93c93d550f5750980c702f214b01a8bb49dd5af37888e6cR45) in craft command should be `Block table = Blocks.CRAFTING_TABLE` (no guarantees for the exact name).

RemainingToast:
I see, well how fleshed out this could this be while staying in the scope of baritone - because could add a gather/craft process similar to altoclefs tasks

ZacSharp:
I think in any form this will go beyond the current "walk and right/left click" feature set (did you ever notice basically every Baritone feature is walking somewhere and optionally clicking on the target position?) simply because clicking things in inventories isn't covered by that and neither is some sort of recipe detection / knowledge base. However I also think that's not a problem and as long as it's a short and simple task it will fit in with the rest. By "short and simple" I mean e.g. "craft this with the things you have in inventory" rather than "craft this and *if neccessary* craft a crafting table *and maybe* get some materials from storage *or* gather *if needed*".

If you want to do more automation I also don't think that's categorically ruled out, but you'll have to prepare some other things beforehand (execution part of #3565 would be good, definitely https://github.com/cabaletta/baritone/issues/881#issuecomment-535882984, also #3904 isn't quite a shining example, but you might be able to get something from the comments, specifically https://github.com/cabaletta/baritone/pull/3904#issuecomment-1505859304) .

Specifically concerning your suggestion of something similar to altoclefs tasks: I don't think that's a good idea. The goal here would be being able to craft any item, not to craft any item someone hardcoded into the system, which is what's done in AltoClef.

rycbar0:
@strubium you still working on this? i got a prototype working, but it still needs work. item parsing is a little buggy atm and it cant place crafting tables. but i can go to a crafting table and craft, as long as the item get parsed.

strubium:
> @strubium you still working on this? i got a prototype working, but it still needs work. item parsing is a little buggy atm and it cant place crafting tables. but i can go to a crafting table and craft, as long as the item get parsed.

No, I kinda lost the mojo. If you want help yeet me your fork and I'll see what I can do.

PR #3955, opened 2023-05-23
feat: dynamic setting registration

RemainingToast:
- Improves extensibility and ease of use for settings


RemainingToast:
Should allow something like this - then should natively work with Baritones command system already.

```
public class TestSettings {

    public final Settings.Setting<Boolean> anySetting = new Settings.Setting<>(false);

    TestSettings() {
        BaritoneAPI.getSettings().registerSettingsFromClass(TestSettings.class);
    }

}
```

src/api/java/baritone/api/Settings.java
```diff
@@ -1385,14 +1421,82 @@ public final Type getType() {
...
+                    String name = field.getName();
+                    setting.name = name;
+                    String lowercaseName = name.toLowerCase();
+                    superSecretRegisterSetting(setting, lowercaseName, ((ParameterizedType) field.getGenericType()).getActualTypeArguments()[0]);
+                } catch (IllegalAccessException e) {
+                    throw new IllegalStateException("Error accessing setting field", e);
+                }
+            }
+        }
+    }
```
ZacSharp, line 1460: This relies on the fields of the input class having the required type information, which is not guaranteed and can break other people's code in unexpected ways.
If there is a comment "here be dragons" it is probably not a good idea to expose exactly that code as public API.

This fails if the field is null.

ZacSharp (commented):
The main question I have with this is "why?".
I know hack clients like to put their settings into the corresponding module, but they also have a clear setting-module pairing in the rest of their UI and code, while Baritone has loads of settings which either apply to multiple commands or apply to no command at. In fact, those that "apply to commands" really apply to a *process* or *behavior* controlled by that command and potentially also via API.

In addition, this allows others to register arbitrary settings, essentially turning the addition of a new setting into a breaking change, because the new setting might conflict with an API added setting.

Also this is very dependent on load order. If someone triggers the first setting load before all settings are registered, the missing settings will be reset to their default value.

Does this break the Baritone discord bot?

ehylo:
> Does this break the Baritone discord bot?

The way the baritone bot does settings is pretty bad and things should not be changed in order to allow it to work, if something breaks the bot I will (if I'm not lazy) fix it.

RemainingToast:
The main reason is to make the Baritone API more extendible - make full use of the API. I was able to register commands through the api already but there is no way of registering settings and then using the subsequent #set settingName command.

> Does this break the Baritone discord bot?
how would this break this discord bot??


@ZacSharp how would you like me to move forward? 
I was thinking the Setting class should probably be in its own file and thats why its static now - to work out of that file.

RemainingToast:
Oh right the discord bot with all the settings - well it shouldn't break it. And the main API discord bot should obviously not show any registered by other mods and only baritone API settings.

🤷‍♂️ Idk how the dc bot works tho

RemainingToast:
Related Discord Conversation:
https://discord.com/channels/822011099561197579/833416754197168159/1110249432789815346

RemainingToast:
> Also this is very dependent on load order. If someone triggers the first setting load before all settings are registered, the missing settings will be reset to their default value.

Also RE this: It shouldn't be possible because the API settings are registered before you can BaritoneAPI#getSettings
image
*Unless I am severely misunderstanding something I should probably learn?*

ZacSharp:
Uhm, did you test save-load roundtripping a single dynamically added setting? From that screenshot I'd expect the worst case load order scenario where you *cannot possibly register a setting before it is too late and its saved value is already discarded*.
E.g. Lambda solves this by saving configs per module so they can always load the module so all settings are registered and only then they load the file from disk. Using a single file should also work so long as values for unknown settings are kept around in case they are registered later.

I asked about the discord bot because my knowledge on it is quite old and I simply wanted to know whether the thing I know (parses settings javadoc so extract help text) is also currently in use. In no way do I think changes should be help back for it. As ehylo said it can be fixed, even if that is by hardcoding the settings into it.
A point for which I would consider tweaking the design though is having the settings documentation available in-game.


> I was thinking the Setting class should probably be in its own file and thats why its static now - to work out of that file.

The `Setting` class can be static (but doesn't need to either). I wondered why you made the mapping from settings to their types static, since all other settings state is tied to the `Settings` instance.

> The main reason is to make the Baritone API more extendible - make full use of the API. I was able to register commands through the api already but there is no way of registering settings and then using the subsequent #set settingName command.

Imho there's other places which would benefit more from being more accessible but yes, that's a point. I'm still not happy about possible name collisions but commands have those as well so idk. 
Maybe you could make namespaced settings? Maybe so it's only needed on the user side if there would be ambiguity without it? Organizing settings in some better in the UI as well would be great.

> @ZacSharp how would you like me to move forward?

1. Get some other opinions. I'm not the person to ultimately decide on this so you also shouldn't treat me as the single source of decision.
2. Ignore the first paragraph of my review and fix the rest. That is
    * Don't make "internal" state publicly exchangeable. If `Collections.unmodifiableMap` allows mutating the underlying map you can use that, otherwise you'll probably have to create a new wrapper map for it.
    * Fix settings loading so dynamic settings can actually load from disk regardless of when they are registered.
    * Fix generics information loss. Using the current internal approach on API is bad because it relies on implicitly passed information which is not even available in a somewhat reliable way. Even a reiteration of the `TypeToken` trick would be better.
    * Hopefully fix name collisions? Might be asking a bit much here.

RemainingToast:
Alright thank you for the feedback - I will make those changes and make another push. Thanks again 👍

src/api/java/baritone/api/Settings.java
```diff
@@ -1385,14 +1421,82 @@ public final Type getType() {
...
+    /**
+     * Retrieves a list of settings of a specific type.
+     *
+     * @param <T>   the type of the settings to retrieve
+     * @param clazz the class representing the type of settings
+     * @return a list of settings of the specified type
+     */
     @SuppressWarnings("unchecked")
-    public <T> List<Setting<T>> getAllValuesByType(Class<T> cla$$) {
+    public <T> List<Setting<T>> getAllValuesByType(Class<T> clazz) {
```
ZeroMemes, line 1470: Please revert this (See https://github.com/cabaletta/baritone/commit/a7add6bc13df16cfbe42e6ae02ff1c6a0c4c1313)

ZeroMemes:
Imo this would be better implemented as part of a larger plugins system, which could be used to provide settings, commands, processes, etc.

To avoid collisions, namespaces could be denoted by the plugin ID, and explicitly specifying the namespace wouldn’t even be necessary for cases where there are no collisions.

PR #3966, opened 2023-06-04
Setting Field Metadata

ZeroMemes:
Replaces the `JAVA_ONLY_SETTINGS` array with a `@JavaOnly` annotation on the fields themselves.

I don't think there's anything else that should be addressed with annotations atm so this PR should be good to go.


leijurv:
this is....so gaming.....

PR #3972, opened 2023-06-06
bugfix

### LINKED ISSUE #3971: #litematica causes game to crash after a while (NullPointerException)
## Some information
Operating system: Windows 10
Java version: 17.0.3
Minecraft version: 1.19.2 
Baritone version: baritone-standalone-fabric-1.9.4
Other mods : malilib-fabric-1.19.2-0.13.0, litematica-fabric-1.19.2-0.12.7

## Exception, error or logs
[latest.log](https://github.com/cabaletta/baritone/files/11664605/latest.log)
[crash-2023-06-06_14.01.43-client.txt](https://github.com/cabaletta/baritone/files/11664606/crash-2023-06-06_14.01.43-client.txt)

## How to reproduce
Load a large litematic and build with the #litematica command (I built this litematic (added .txt so I can attach here))
[_MONUMENT1 - TILL LAMPS.litematic.txt](https://github.com/cabaletta/baritone/files/11664690/_MONUMENT1.-.TILL.LAMPS.litematic.txt)

After a while which can be anywhere from 5 seconds to 30 minutes the game crashes.

## Modified settings
allowInventory (Boolean)    (tried with enabled and with disabled)
allowSprint (Boolean)
blockBreakAdditionalPenalty (Double)
blockPlacementPenalty (Double)
blockReachDistance (Float)
breakCorrectBlockPenaltyMultiplier (Double)
breakFromAbove (Boolean)
buildInLayers (Boolean)

rycbar0:
I think this should fix #3971

### DIFF
--- a/src/main/java/baritone/utils/schematic/MapArtSchematic.java
+++ b/src/main/java/baritone/utils/schematic/MapArtSchematic.java
@@ -46,7 +46,7 @@ public class MapArtSchematic extends MaskSchematic {
             for (int z = 0; z < schematic.lengthZ(); z++) {
                 IBlockState[] column = schematic.getColumn(x, z);
 
-                OptionalInt lowestBlockY = lastIndexMatching(column, state -> !(state == null || state.getBlock() instanceof BlockAir));
+                OptionalInt lowestBlockY = lastIndexMatching(column, state -> !(state.getBlock() instanceof BlockAir));
                 if (lowestBlockY.isPresent()) {
                     heightMap[x][z] = lowestBlockY.getAsInt();
                 } else {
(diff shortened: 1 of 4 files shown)


ZeroMemes:
Maybe it would make more sense to return `false` from `inSchematic` if `states[x][y][z] == null`?

ZeroMemes:
Or as a more permanent solution, treat each entry in `Region` as its own `StaticSchematic`, and have `LitematicaSchematic` extend `CompositeSchematic`. The only issue with this that I could see is if the `.litematic` file format allows sub-regions to intersect.

ZacSharp:
Litematica also has to handle that situation, even before pasting, because all regions are loaded into the same `WorldSchematic` for rendering. I'd say if it handles overlap consistently we should do the same, otherwise there is no reason to care.

The problem with `CompositeSchematic` is that it's not an `IStaticSchematic` (which is needed for `MapArtSchematic`) and can easily become one because `getColumn` doesn't (currently?) allow holes.

BL1Z33 (approved):

rycbar0:
still needs fixing but this pr points now to the wrong branch

PR #3973, opened 2023-06-06
baritoneminecraft.com

BL1Z33:
Let everybody reading the README know that baritoneminecraft.com isn't an official Baritone website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,3 +132,6 @@ We thank them for granting Baritone an OSS license so that we can make our softw
 ## Why is it called Baritone?
 
 It's named for FitMC's deep sultry voice.
+
+## Important note
+baritoneminecraft.com is not an official Baritone website. If you find anybody telling you to download Baritone from this site, use the [official repo](https://github.com/cabaletta/baritone/) instead.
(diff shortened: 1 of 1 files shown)


UltraBlackLinux:
the website may not be really legitimate but it does link to the github repository instead of providing anything to download.

Also if you start with them you should also incluse 9minecraft and all the shitty other websites, because these actually distribute their own jars it seems

BL1Z33:
> the website may not be really legitimate but it does link to the github repository instead of providing anything to download.
> 
> Also if you start with them you should also incluse 9minecraft and all the shitty other websites, because these actually distribute their own jars it seems.
> 
> Also I don't think it makes much sense to warn somebody from illegitimate downloads if they are already in the github repo. Adding a warning here won't help the 9minecraft users...
> 
> I'd say this is not worth merging.

ill add that, i just saw an issue and no pr so i made one

UltraBlackLinux:
do you not understand? The 9minecraft people are not gonna pin this warning on their baritone download

this won't serve a purpose...

scorbett123:
I think possibly a "downloads from anywhere except the official repo are not official" owtte is fine, however specific name calling isn't as then you really need to call out everyone.

However I'm not entirely sure if this is necessary at all. I'm not actually going to close this as I'm not sure, but I'm doubtful if this is a good change.

0-x-2-2:
Don't link the websites as that just boosts SEO.

PR #3974, opened 2023-06-07
Add sphere and cylinder build commands to sel

ZeroMemes:
Could be better implemented through an advanced selection system so that replace/copy/paste/etc. could all work with masked selections, but that'd probably end up being API-breaking and is way beyond the scope of basic building commands.

Adds the commands:
- `sel sphere <block>`
- `sel hsphere <block>`
- `sel cylinder <block>`
- `sel hcylinder <block>`

Also fixes a bug where `sel shell` doesn't auto-complete the `block` argument.

Drafting this PR since the calculation for the masks is horribly inefficient and could just be pre-computed.


ZacSharp:
> Could be better implemented through an advanced selection system so that replace/copy/paste/etc. could all work with masked selections, but that'd probably end up being API-breaking and is way beyond the scope of basic building commands.

Maybe I'm completely misinterpreting your idea here, tough if it can be accomplished by simply allowing selections to have a mask (`#sel p1` `#sel p2` `#sel mask all sphere` `#sel fill stone` rather than `#sel p1` `#sel p2` `#sel sphere stone`) I think it can be done without breakages.
Won't argue against the scope concern though. `BuilderProcess` wouldn't be able to handle half of it anyway.

ZeroMemes:
> Maybe I'm completely misinterpreting your idea here, tough if it can be accomplished by simply allowing selections to have a mask

I was more thinking along the lines of selections fundamentally representing a shape, rather than the existing system of axis-aligned boxes, but I definitely worded it like "add masks to selections" which could be done as you said without API breakage.

ZeroMemes:
I WUV codacy

ZacSharp:
What is the intended behavior when `partOfMask` is called with a position outside the bounds of the mask? Di implementations have to return false in that case, do they have to return true or false or are they free to do whatever they want (like throwing exceptions)?

ZeroMemes:
Implementations can do whatever they want. The assumption is that the coordinates are within bounds and verified by the caller, which is the same as how `desiredState` behaves in schematic.

PR #3979, opened 2023-06-11
Allow free looking when breaking and placing blocks

ZeroMemes:


ZeroMemes:
Still requires some work. Doesn't work with `antiCheatCompatibility` disabled due to the player angles not being set at all.

ZeroMemes:
Did some cleanup, there's still some considerations though.

- Should `onSendPacket` be used to read in `LookBehavior.serverRotation`?
- Should `LookBehavior.pig()` be updated to respect the `Target.Mode`?

src/api/java/baritone/api/utils/RotationUtils.java
```diff
@@ -215,9 +216,10 @@ public static Optional<Rotation> reachable(EntityPlayerSP entity, BlockPos pos,
      * @param blockReachDistance The block reach distance of the entity
      * @return The optional rotation
      */
-    public static Optional<Rotation> reachableOffset(Entity entity, BlockPos pos, Vec3d offsetPos, double blockReachDistance, boolean wouldSneak) {
+    public static Optional<Rotation> reachableOffset(EntityPlayerSP entity, BlockPos pos, Vec3d offsetPos, double blockReachDistance, boolean wouldSneak) {
+        IPlayerContext ctx = BaritoneAPI.getProvider().getBaritoneForPlayer(entity).getPlayerContext();
```
ZacSharp, line 220: Technically this is api breaking, would be surprised if anyone calls it on arbitrary entities though.

Does this throw an NPE if I call it on a fake player?

ZeroMemes: > Technically this is api breaking, would be surprised if anyone calls it on arbitrary entities though.

Yeah I doubt that anybody is actually calling this. Either way I'm going to add back the one with the `Entity` parameter as `@Deprecated` and change the updated version to take an `IPlayerContext` instead of resolving the baritone instance from the player to get it.

> Does this throw an NPE if I call it on a fake player?

Yes. It'd only work correctly for `mc.player` on the main branch.

leijurv:
while you're looking at this stuff, i think there might be a missing mixin or missing hook. i don't know if it's in 1.12 or in a future version though.

jiggling the mouse while pathing or mining a block shouldn't send more player rotation packets to the server...... but it does (in 1.19.4 at least). mouse movement isn't being properly suppressed, but while freelook is on, it should be

src/main/java/baritone/behavior/LookBehavior.java
```diff
@@ -126,8 +128,12 @@ public void pig() {
         }
     }
 
-    public Rotation getEffectiveRotation() {
-        return this.serverRotation;
+    public Optional<Rotation> getEffectiveRotation() {
+        if (Baritone.settings().freeLook.value || Baritone.settings().blockFreeLook.value) {
+            return Optional.of(this.serverRotation);
```
ZacSharp, line 133: Doesn't `Optional.of` throw an NPE if you pass it `null` and this should be `Optional.ofNullable` to turn the `null` into an `Optional.empty()`?

ZeroMemes: fixed lol

src/main/java/baritone/behavior/LookBehavior.java
```diff
@@ -20,111 +20,209 @@
...
+        return current + mouseToAngle(deltaPx);
+    }
+
+    private static int angleToMouse(float angleDelta) {
+        final float minAngleChange = mouseToAngle(1);
+        return Math.round(angleDelta / minAngleChange);
+    }
+
+    private static float mouseToAngle(int mouseDelta) {
+        final float f = Helper.mc.gameSettings.mouseSensitivity * 0.6f + 0.2f;
```
ZacSharp, line 179: On newer versions (checked 1.13 and 1.16) this setting is a `double`. Should it just be cast into a `float` (what I've done so far) or should all calculations be changed to use `double` instead?

ZacSharp:
This causes unintuitive behavior when used with e.g. `#build` because pressing WASD will use the `LookBehavior` target rotations rather than the player rotations. Since I don't think we should block WASD as soon as Baritone does anything (manually overwriting Baritone inputs can be useful) I think it would be better to only set rotations during movement update if Baritone is actually pressing any of the movement keys (in which case player input is already blocked).

This also needs to set rotations while `InputOverrideHandler` is processing mouse inputs to prevent unintended desyncs when using items such as water buckets, because they read the player rotations:

https://github.com/cabaletta/baritone/assets/68165024/fbf73239-2f52-4b61-8a15-c62c87c79826

src/main/java/baritone/behavior/LookBehavior.java
ZeroMemes: I believe the player's angles are still a float, so however the game handles the setting in the mouse callback code should be replicated.

ZeroMemes: <details>
  <summary>Screenshots of the relevant game code from 1.19</summary>

image
image

</details>

PR #3982, opened 2023-06-12
Batch Rendering

ZeroMemes:
Reduces the number of `draw()` calls every frame from upwards of a few hundred to <10

I'm thinking about creating a separate `GoalRenderer` API though so that batching can be done a little smarter, especially with `GoalComposite`. (Would also help with cleaning up the current `drawGoal` mess)


leijurv:
WAOW!!!!!!!!!!!!!

PR #3984, opened 2023-06-12
Fix `censorCoordinates` recalc bug

ZeroMemes:
See #3565

From my own testing (by going through each usage of `FORCE_REVALIDATE_GOAL_AND_PATH` manually), everything should be correct

### DIFF
--- a/src/main/java/baritone/process/MineProcess.java
+++ b/src/main/java/baritone/process/MineProcess.java
@@ -319,6 +319,21 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
             int zDiff = z - this.z;
             return GoalBlock.calculate(xDiff, yDiff < -1 ? yDiff + 2 : yDiff == -1 ? 0 : yDiff, zDiff);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            return super.equals(o);
+        }
+
+        @Override
+        public String toString() {
+            return String.format(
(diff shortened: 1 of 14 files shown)


leijurv:
brady hates subclassing? he uses `getClass` instead of `instanceof`??

ZeroMemes:
IntelliJ automatic equals generation LOl

leijurv:
also le baritone style is `{}` around all `if`s 😤

ZeroMemes:
Can't have `new GoalBreak(BlockPos.ZERO).equals(new GoalGetToBlock(BlockPos.ZERO)) == true` 😤

ZacSharp (approved):

src/main/java/baritone/process/MineProcess.java
```diff
@@ -319,6 +319,21 @@ public double heuristic(int x, int y, int z) {
             int zDiff = z - this.z;
             return GoalBlock.calculate(xDiff, yDiff < -1 ? yDiff + 2 : yDiff == -1 ? 0 : yDiff, zDiff);
         }
+
+        @Override
+        public boolean equals(Object o) {
+            return super.equals(o);
+        }
```
ZacSharp, line 326: Why no early return for mismatched classes here?
A `GoalThreeBlocks` and a `GoalTwoBlocks` don't have the same heuristic or goal condition so they aren't equals.

PR #3985, opened 2023-06-13
Deprecate `Helper.mc`

ZeroMemes:
Also fixes some `Minecraft.getMinecraft()` references, reorganizes some code, and allows user creation of Baritone instances from a Minecraft instance.


ZeroMemes:
I didn't do anything about settings btw those still get stored in `Minecraft.getMinecraft().gameDir`

PR #3987, opened 2023-06-14
some modifications to buildsrc

wagyourtail:
checksum only new. make these work a bit more proper (still very jank)


ZacSharp:
Why only 1.20 and not 1.19?

wagyourtail:
Fair point... ill pull it back

PR #3988, opened 2023-06-14
swimming

leijurv:
someone else test this i am TOO TIRED and i hate swimming mechanics (this is why i have put this off even though swimming was added to minecraft before baritone even started)

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -27,12 +27,13 @@ import net.minecraft.network.chat.Component;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.Blocks;
+
 import java.awt.*;
 import java.lang.reflect.Field;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.*;
 import java.util.List;
+import java.util.*;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
@@ -729,6 +730,11 @@ public final class Settings {
      * Continue sprinting while in water
      */
     public final Setting<Boolean> sprintInWater = new Setting<>(true);
+
+    /**
+     * Attempt to enter the underwater swimming state
+     */
+    public final Setting<Boolean> swimInWater = new Setting<>(true);
 
     /**
      * When GetToBlockProcess or MineProcess fails to calculate a path, instead of just giving up, mark the closest instance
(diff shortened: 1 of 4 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -730,6 +731,11 @@ public final class Settings {
      */
     public final Setting<Boolean> sprintInWater = new Setting<>(true);
 
+    /**
+     * Attempt to enter the underwater swimming state
```
joerez, line 735: aka California after the big earthquake hits

ZacSharp:
The fact that this works is kinda stupid. Why doesn't looking up make you bob out of the water like pressing space does?

Anyway, `MovementDiagonal` doesn't actually sprint swim and this (explicitly) ignores `freeLook`.
I wonder whether moving the autoswim before the movement update would break anything. That way movements could explicitly unpress space instead of being a special case when pressing space.

leijurv:
Yeah I agree that this is stupid and badly written I just wanted to see if it would work.

I picked pitch -30 because in my testing, that pitch always gets the player's head out of the water so that they don't drown, but never ends swimming by going too high.

Once brady's (@ZeroMemes 's) recent PRs are merged up, this could use the movement freelook from there. But for now, I found that the pitch was not freelookable. Meaning, if I just set the pitch as a movement target with force=false, it actually did not affect swimming direction sadly. IDK if this is a 1.13+ thing or what.

Yes, the same logic should be added to MovementDiagonal to begin swimming.

ZeroMemes:
Perhaps the RotationMoveEvent change from `elytra-freelook` is needed?? https://github.com/cabaletta/baritone/commit/bde0c620ad6270cbd363140ab6be8aa49370644d

ZacSharp:
Actually, I didn't mean the code.
<details><summary>I meant this</summary>

https://github.com/cabaletta/baritone/assets/68165024/9a5f75a8-193f-4f4d-98f1-f0c1714a6d34

Even looking straight up won't stop it. The only thing that happens if you look too far up is that momentum can carry you out of the water so starting can take a second or third attempt.
</details>

> Once brady's recent PRs are merged up, this could use the movement freelook from there

Ok. In progress.

ZacSharp:
I tried this on out 1.16 with all of current master merged in and some changes to accommodate for changes in swimming mechanics (you need to press space to get high enough).
The player pitch is read in `PlayerEntity.travel`, which is above the elytra injection pair in the call hierarchy. To make swimming freelookable we'd either have to cover the entirety of that method with an injection pair or add one more targeted injection (pair) for just that span (didn't yet test the latter).
That `RotationMoveEvent` change will/did certainly come in handy.

PR #3992, opened 2023-06-17
Actually use stack hashes

ZacSharp:
@wagyourtail is there any specific reason why they weren't used?


0-x-2-2:
does this work with hash code conflicts if any exist?

ZacSharp:
We are relying on not getting random hash collisions (ehm...) with or without this so I figured it doesn't matter.

If you say that's a problem then the stack hashes set in `BlockOptionalMeta` should be removed as well.

PR #3996, opened 2023-06-18
firework inv management + pause/resume

rfresh2:
elytra flying epic style


src/main/java/baritone/behavior/ElytraBehavior.java
```diff
@@ -688,4 +766,5 @@ private static Chunk cachedChunk(final World world,
         }
         return world.getChunk(chunkX, chunkZ);
     }
-}
\ No newline at end of file
+}
+
```
ZacSharp: Useless noise.

src/main/java/baritone/behavior/ElytraBehavior.java
```diff
@@ -410,20 +439,69 @@ public void onTick(TickEvent event) {
 
         if (forceUseFirework || (!firework
                 && sinceFirework > 10
+                && Baritone.settings().elytraUseFireworks.value
```
ZacSharp, line 442: You also have to make sure it doesn't try to take off using fireworks if this setting is disabled (or document the fact that takeoff will always use fireworks).

src/main/java/baritone/behavior/ElytraBehavior.java
```diff
@@ -410,20 +439,69 @@ public void onTick(TickEvent event) {
...
+                final int firstFireworksInHotbar = firstFireworksInHotbar();
+                if (firstFireworksInHotbar == -1) {
+                    if (!swapToFireworksInInventory()) {
+                        logDirect("no fireworks");
+                        return;
+                    }
+                } else {
+                    ctx.player().inventory.currentItem = firstFireworksInHotbar;
+                }
+            }
```
ZacSharp, line 457: This looks *a lot* like `baritone.getInventoryBehavior().throwaway(true, ElytraBehavior::isBoostingFireworks)`

ZacSharp (commented):
I have to main concerns here
1. This is an early prototype and getting things to Just Work as early as possible is cool, but I don't think other mods should already depend on it (i.e. no api exposure for now). Currently `ElytraBehavior` is not much more than a raw executor with no proper control mechanisms in place and regardless of whether this is going to have its own kind of process system or the process system is adapted to control both `PathingBehavior` and `ElytraBehavior` (the latter sounds better tbh), `ElytraBehavior` will not be paused directly just like you can't pause `PathingBehavior` directly.
2. I'm not sure whether external firework controls will continue to make sense in the long run. There are plans to improve fireworks usage and once those are implemented an external client with less information about the path (and no way of changing the path to accommodate for future firework usage) might no longer be able to do better than Baritone itself. I don't know what those plans are so this as well is a maybe.

src/main/java/baritone/behavior/ElytraBehavior.java
leijurv: ^

leijurv:
what zacsharp said

leijurv:
idk how pathing control will work but pause/resume doesn't make much sense to me. recalculating the elytra path takes literally less than 1 millisecond to render distance. if you pause and fly off to the side, it **should** recalculate, rather than pausing and resuming on the same old path.

leijurv:
ok actual thoughts

@ZeroMemes feel free to merge this if you like

1. setting for baritone elytra to use fireworks or not - i don't know about this, because baritone elytra needs to know how its physics works. if another client is doing some kind of efly module that totally changes elytra behavior that just won't work at all, so i don't want to confuse people into thinking that could work. while under firework boost the physics are totally different and the anti collision simulations are also therefore totally different. so i don't think that's good. however if the setting is JUST for having another client deciding if/when to use fireworks, but it's normal physics with/without fireworks, then i guess there's nothing wrong with that, but i just don't want to focus on that at the moment.
2. for inventory, see zacsharp's comments. it should call InventoryBehavior rather than doing its own thing. perhaps, when #allowInventory is true, it should move fireworks to the hotbar only while actively elytraflying
3. IElytraBehavior: as i said in previous comment, i don't think that's what the api should be. isActive is fine, but instead of pause/resume it should be pathTo(blockpos) and cancel()
4. the mixin to see which entity is being boosted is cool and good, thanks

meow meow

ZeroMemes:
Thank you @rfresh2 very cool

I will be making the changes that @leijurv mentioned tho

rfresh2:
cool i was just about to start on those but i'll let you take it from here then.

PR #4000, opened 2023-06-18
Fix `CylinderMask` with horizontal alignment

ZacSharp:
This took far too much time to find.
The problem was that `getA` and `getB` used the `alignment` field to determine their return value, but when calculating `centerA` and `centerB` based on `getA` and `getB` in the constructor `alignment` is still null.
I could have just moved the initialization of `this.alignment` to the top of the constructor, but this way there is no calls to instance methods in the constructor at all and no risk of accidentally messing it up again.


ZeroMemes:
Whoops lol

leijurv:
naughty brady 🥺

leijurv:
thank you based and redpilled zacsharp

PR #4002, opened 2023-06-19
Add useMessageTag setting

jdbool:
...and make it function. Uses that 1.19.something tag bar system.

It could be any 24-bit colour but I just chose the same default light purple.

## Disabled
Screenshot 2023-06-18 at 10 44 53 PM

## Enabled
Screenshot 2023-06-18 at 11 19 54 PM
Screenshot 2023-06-18 at 11 19 58 PM


ZacSharp:
If this a 1.19 feature (didn't check) then the pr should be against the 1.19 branch (always against the lowest version which makes sense and is supported).

You changed a bunch of API method signatures which is not nice. Can you please add overloads instead? (or do this completely inside the default value of `logger`? Would that make sense?)

jdbool:
OK, closing this, will point it to 1.19 in a minute

PR #4014, opened 2023-06-21
Add useMessageTag setting

jdbool:
...and make it function. Uses that 1.19.1 tag bar system.

It could be any 24-bit colour but I just chose the same default light purple.

## Disabled
Screenshot 2023-06-18 at 10 44 53 PM

## Enabled
Screenshot 2023-06-18 at 11 19 54 PM
Screenshot 2023-06-18 at 11 19 58 PM

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -17,6 +17,7 @@
 
 package baritone.api;
 
+import baritone.api.utils.Helper;
 import baritone.api.utils.NotificationHelper;
 import baritone.api.utils.SettingsUtil;
 import baritone.api.utils.TypeUtils;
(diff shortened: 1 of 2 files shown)


ZacSharp:
Well I guess the setting just can't be backwards compatible.

One detail: In my opinion the message tag is just a different type of prefix so there shouldn't be a parameter to toggle it, like there isn't one for `shortBaritonePrefix` either. (i.e. I think it should decide based on the setting directly in `logDirect(logAsToast, components)`).
`logAsToast` is different because it redirects the message to a completely different output channel.

jdbool:
> Well I guess the setting just can't be backwards compatible.
> 
> One detail: In my opinion the message tag is just a different type of prefix so there shouldn't be a parameter to toggle it, like there isn't one for `shortBaritonePrefix` either. (i.e. I think it should decide based on the setting directly in `logDirect(logAsToast, components)`). `logAsToast` is different because it redirects the message to a completely different output channel.

Agreed, updated. Made the setting backward compatible too

ZacSharp (approved):
Relying on the logger to get the tag right is not ideal, but gotta pick your poison.
Passing the tag in would behave very similarly anyway (relies on the logger to use the tag instead of relying on the logger to add the tag).

leijurv:
that's cool

leijurv:
whoever does the merging up just needs to remember that this is in 1.19.2 and not 1.19.4

PR #4015, opened 2023-06-21
API for rotations produced by LookBehavior

ZeroMemes:
Fully deterministic with support for simulation of future ticks.

This PR does modify the way that RotationMoveEvent works in coordination with LookBehavior, meaning that randomLooking and simulated mouse input does have an affect on player movement. However, Baritone for 1.13+ already has affected movement because of randomLooking113.

### DIFF
--- a/src/api/java/baritone/api/behavior/look/IAimProcessor.java
+++ b/src/api/java/baritone/api/behavior/look/IAimProcessor.java
@@ -1,0 +1,44 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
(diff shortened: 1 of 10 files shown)


ZacSharp:
Does this mean `randomLooking113` can be moved into the normal rotation processing once this is merged?

src/api/java/baritone/api/behavior/look/IAimProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+public interface IAimProcessor {
+
+    /**
+     * Returns the actual rotation that will be used when the desired rotation is requested. This is not guaranteed to
+     * return the same value for a given input.
+     *
+     * @param desired The desired rotation to set
+     * @return The actual rotation
+     */
+    Rotation nextRotation(Rotation desired);
```
leijurv, line 34: I think this javadoc should explicitly say whether or not calling this function will advance the RNG by one tick, or not

leijurv (approved):

src/api/java/baritone/api/behavior/look/IAimProcessor.java
ZacSharp: I think using one method to advance by one tick *or* peek the next tick depending on the instance is not good.
Can `IAimProcessor` maybe be split into one interface which can only be peeked and forked and a second interface which extends the first one by adding methods to tick it and to get a rotation and advance one tick simultaneously? The root processor would implement just the first interface, the forked ones the second interface.

ZeroMemes: Agreed

ZacSharp (approved):

PR #4017, opened 2023-06-22
simplify null check

babbaj:


ZeroMemes:
thank you babbaj for fixing yet another critical vulnerability

Issue #3409, opened 2022-04-21
acceptableThrowawayItems is being moved into players hotbar, when baritone is inactive


Rumrobot:
Minecraft version: 1.18.2
Baritone version: 1.8.3

## How to reproduce
Set allowInventory to true. Get some acceptableThrowawayItems and they should be moved into your hotbar. 

The baritone is still doing this while inactive, so it's kind of annoying when playing.

ZacSharp:
This is a feature you explicitly enabled by setting `allowInventory` to `true`, why are you reporting it as a bug?
If you think the intended behaviour should be changed you should open a "suggestion issue instead of reporting the current behaviour as a bug.

I'll just change the label and it would be nice if you could edit your issue to contain a little more information (look at the [suggestion template](https://github.com/cabaletta/baritone/issues/new?assignees=&labels=enhancement&template=suggestion.md&title=Please+add+a+brief+but+descriptive+title) for what to put in it).

Rumrobot:
Thx i will do that, but i would call it a bug since baritone is inactive. Is it made to specifically do it like that? And then what would the point be to have it do it, while it has nothing to do?

ZacSharp:
I think that's intended. It's doing it here https://github.com/cabaletta/baritone/blob/cbef05838db84ec77f7a14234b44675a1bc9e5ab/src/main/java/baritone/behavior/InventoryBehavior.java#L57 since the inventory handling was added 3 years ago and I never saw a sign of someone considering it wrong behavior. There is checks to disable it in containers or while not ingame (and of course when `allowInventory` is disabled) but nothing to check whether we actually need the items.

liangforstudy:
Oh yeah, I find this annoying as well. Sometimes there's a cobblestone/dirt block on my 9th slot on my hotbar. And when I tried to keep that in my inventory, Baritone keep placing it back into my 9th slot on my hotbar until I turn @ allowinventory off.

Issue #3000, opened 2021-09-19
Baritone does not allow switch of pickaxe from hotbar


maxsupermanhd:
Operating system: any
Java version: any
Minecraft version: any
Baritone version: 1.2.15
Other mods (if used): does not matter

Baritone keeps swapping back pickaxe to first hotbar slot regardless of user actions. Slot can not be customized (aka I want it to always have pickaxe on slot 2 instead). No config for pickaxe keeping behavior (except AllowInventory that disables like, *everything*).

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [x] I have included logs, exceptions and / or steps to reproduce the issue.
- [x] I have not used any OwO's or UwU's in this issue.

ZacSharp:
Why do people think when they want something that is not implemented it's a bug? That definition of bug would only be applicable if the project explicitly stated "we have everything you want", but that's not the case for Baritone and following from that missing features are **not** a bug.

PR #4021, opened 2023-06-23
Suggestion

### LINKED ISSUE #4020: Baritone only moves pickaxes and throwaway blocks to the hot bar and does that even if no process is active
Inventory behavior only picks tools of the pickaxe type into the hotbar. If your shovel breaks and you dont have a second one in your hotbar baritone will use its fist to mine dirt blocks instead of using the backup shovel in your inventory regardless of your allowInventory setting.
Also it will always move pickaxes and throwAwayItems to the hotbar. maybe there is a good reason for that but i think if baritone doesnt have a process running it shouldnt move itmes in the inventory.
https://github.com/cabaletta/baritone/blob/1ea071de09631a41f0ea743df036db8000b7147c/src/main/java/baritone/behavior/InventoryBehavior.java#L60-L66

rycbar0:
a suggestion on how to fix #4020

### DIFF
--- a/src/main/java/baritone/behavior/InventoryBehavior.java
+++ b/src/main/java/baritone/behavior/InventoryBehavior.java
@@ -46,7 +46,7 @@ public final class InventoryBehavior extends Behavior implements Helper {
 
     @Override
     public void onTick(TickEvent event) {
-        if (!Baritone.settings().allowInventory.value || baritone.getPathingControlManager().mostRecentInControl().orElse(null) == null) {
+        if (!Baritone.settings().allowInventory.value || baritone.getPathingControlManager().mostRecentInControl().isPresent()) {
             return;
         }
         if (event.getType() == TickEvent.Type.OUT) {
--- a/src/main/java/baritone/behavior/InventoryBehavior.java
+++ b/src/main/java/baritone/behavior/InventoryBehavior.java
@@ -46,7 +46,7 @@ public final class InventoryBehavior extends Behavior implements Helper {
 
     @Override
     public void onTick(TickEvent event) {
-        if (!Baritone.settings().allowInventory.value) {
+        if (!Baritone.settings().allowInventory.value || baritone.getPathingControlManager().mostRecentInControl().orElse(null) == null) {
             return;
         }
         if (event.getType() == TickEvent.Type.OUT) {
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/behavior/InventoryBehavior.java
```diff
@@ -46,7 +46,7 @@ public InventoryBehavior(Baritone baritone) {
 
     @Override
     public void onTick(TickEvent event) {
-        if (!Baritone.settings().allowInventory.value) {
+        if (!Baritone.settings().allowInventory.value || baritone.getPathingControlManager().mostRecentInControl().orElse(null) == null) {
```
ZeroMemes, line 49: This can be replaced with `isPresent()`

ZacSharp:
This does certainly fix both issues (you can also add a "fix" or "fixes" for #3000 and #3409), but I don't like the way you fixed the "only pickaxes are moved" problem.

To fix the "don't move items you don't need" issue properly we have to move items only when we know we actually need them, because otherwise this fix is just going to result in follow up bug reports where people complain it insists on having a pickaxe even though `allowBreak` is off or it moves tools even though no blocks in `allowBreakAnyway` are in render distance or it moves stone with `allowPlace` disabled or it doesn't choose the correct super-hyper-spade from whatever-mod because that spade only works for blocks from whatever-mod or it doesn't move shears to hotbar to break leaves (same for hoes and swords).
What we need is a general purpose item selection method like `InventoryBehavior.throwaway` which can also select items by priority. The intention behind using general priority based selection rather than specialized tool selection is that e.g. `ElytraBehavior` also have use for this. It currently prioritizes boosting fireworks by trying to get some and if that fails it tries to get fireworks without boost. Prioritizing longer boosts would currently require two more queries (because there's 3 boost lengths) or duplicating the logic.
.

ZacSharp:
Just noticed I said that I didn't like one fix and then went on about the other, oops.
Guess that's because I really want on demand selection, which is a fix to both the hotbar space problem (3 tools + throwaway is 4 slots before it even started working on a task) and the various cases where it still won't move the tool or still would move the tool way too eagerly (i.e. if it can't possibly need it).

rycbar0:
no longer relevant

Issue #3945, opened 2023-05-11
Cannot invoke "net.minecraft.class_1922.method_8320(net.minecraft.class_2338)" because "world" is null


EvilEpicPanda:
## Some information
Operating system: Windows 10
Java version: 17
Minecraft version: 1.19.2
Baritone version: 1.9.4
Other mods (if used): [Prominence Fabric Modpack](https://www.curseforge.com/minecraft/modpacks/prominence-fabric)

## Exception, error or logs
Uncaught exception in thread "pool-9-thread-3"
java.lang.NullPointerException: Cannot invoke "net.minecraft.class_1922.method_8320(net.minecraft.class_2338)" because "world" is null
	at net.minecraft.class_4970$class_4971.handler$hlc000$interference$getHardness(class_4970.java:5653) ~[client-intermediary.jar:?]
	at net.minecraft.class_4970$class_4971.method_26214(class_4970.java) ~[client-intermediary.jar:?]
	at baritone.jp.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at baritone.jp.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at baritone.jp.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at java.util.function.Function.lambda$compose$0(Unknown Source) ~[?:?]
	at java.util.HashMap.computeIfAbsent(Unknown Source) ~[?:?]
	at baritone.jp.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at baritone.gp.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at baritone.iv.a(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at baritone.iv.d(Unknown Source) ~[baritone-standalone-fabric-1.9.4.jar:?]
	at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source) ~[?:?]
	at java.util.stream.DistinctOps$1$2.accept(Unknown Source) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(Unknown Source) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(Unknown Source) ~[?:?]
[...]

PR #4037, opened 2023-07-01
Improve InventoryBehavior

### LINKED ISSUE #4020: Baritone only moves pickaxes and throwaway blocks to the hot bar and does that even if no process is active

ZeroMemes:
Eventually aiming to fix #4020, #3409, #3000, the first point listed in #3565, and probably other issues.

### DIFF
--- a/src/main/java/baritone/PerformanceCritical.java
+++ b/src/main/java/baritone/PerformanceCritical.java
@@ -1,0 +1,33 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that should be used on methods which are performance critical (i.e. called millions of times per second
+ * by the pathfinder) and should be modified with care.
+ *
+ * @author Brady
+ */
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@Retention(RetentionPolicy.SOURCE)
+public @interface PerformanceCritical {}
--- a/src/main/java/baritone/behavior/InventoryBehavior.java
+++ b/src/main/java/baritone/behavior/InventoryBehavior.java
@@ -33,6 +33,7 @@ import net.minecraft.util.NonNullList;
 import java.util.ArrayList;
 import java.util.OptionalInt;
 import java.util.Random;
+import java.util.function.IntPredicate;
 import java.util.function.Predicate;
 
 public final class InventoryBehavior extends Behavior implements Helper {
@@ -70,7 +71,7 @@ public final class InventoryBehavior extends Behavior implements Helper {
         }
     }
 
-    public boolean attemptToPutOnHotbar(int inMainInvy, Predicate<Integer> disallowedHotbar) {
+    public boolean attemptToPutOnHotbar(int inMainInvy, IntPredicate disallowedHotbar) {
         OptionalInt destination = getTempHotbarSlot(disallowedHotbar);
         if (destination.isPresent()) {
             if (!requestSwapWithHotBar(inMainInvy, destination.getAsInt())) {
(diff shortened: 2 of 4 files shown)


src/main/java/baritone/PerformanceCritical.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that should be used on methods which are performance critical (i.e. called millions of times per second
+ * by the pathfinder) and should be modified with care.
+ *
+ * @author Brady
+ */
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
```
ZacSharp, line 31: Isn't a constructor marked as `@PerformanceCritical` already a hint you are doing something wrong?
I don't think this should be applicable to constructors.

src/main/java/baritone/behavior/InventoryBehavior.java
```diff
@@ -167,19 +163,18 @@ public boolean selectThrowawayForLocation(boolean select, int x, int y, int z) {
         if (maybe != null && throwaway(select, stack -> stack.getItem() instanceof ItemBlock && ((ItemBlock) stack.getItem()).getBlock().equals(maybe.getBlock()))) {
             return true;
         }
-        for (Item item : Baritone.settings().acceptableThrowawayItems.value) {
-            if (throwaway(select, stack -> item.equals(stack.getItem()))) {
-                return true;
-            }
-        }
-        return false;
+        return throwaway(select, this::isThrowawayItem);
```
ZeroMemes, line 166: This technically changes the behavior of the method by considering the slot before the item, but that's what `firstValidThrowaway` does so /shrug

src/main/java/baritone/PerformanceCritical.java
ZeroMemes: True

src/api/java/baritone/api/Settings.java
```diff
@@ -73,6 +73,12 @@ public final class Settings {
      */
     public final Setting<Boolean> allowInventory = new Setting<>(false);
 
+    /**
+     * Allow Baritone to automatically put useful items (such as tools and throwaway blocks) on the hotbar while
+     * pathing. Having this setting enabled implies {@link #allowInventory}.
+     */
+    public final Setting<Boolean> allowHotbarManagement = new Setting<>(false);
```
ZacSharp, line 80: Baritone settings style so far was settings depending on each other rather than implying each other (`allowParkour` -> `allowParkourPlace`, `avoidance` -> `mobAvoidance`, maybe more) and from testing `blockFreeLook` I found it very annoying and ended up always having `freeLook` disabled and controlling it via `blockFreeLook` because that's less typing.

ZeroMemes: Fair point, I did too encounter this issue while testing but never bothered to fix it. I've updated this setting accordingly and will go ahead with fixing `blockFreeLook` on `master` too.

ZeroMemes:
While I'm looking at ToolSet (since there's multiple code paths that almost do the same thing but not quite, for example `useSwordToMine` is only recognized by `getBestSlot`), it's probably worth fixing the following line to address issues like #3945.

https://github.com/cabaletta/baritone/blob/6654476da435203df9f5f42ed9f41310f5560c30/src/main/java/baritone/utils/ToolSet.java#L181

rycbar0:
will InventoryBehavior get a method to safely move items from, to, inside containers (like furnaces)?

ZeroMemes:
> will InventoryBehavior get a method to safely move items from, to, inside containers (like furnaces)?

That's not an objective of this PR, but I hope to get the code to a point where adding such functionality would be as simple as creating a custom inventory move operation that could be submitted in a similar way to regular inventory->hotbar moves.

ZacSharp:
Since I'll be away for the next couple days, will this be capable of doing what [this patch](https://github.com/ZacSharp/baritone/commit/42f5deae773c09f5ff67985bc75fc759a76b6a90) could do all the time (i.e. move items on demand)? So far you only seem to plan on cleaning up the old behavior.

To find a slot to move the item to, instead of using hardcoded slots `InventoryBehavior` could track when each item was last requested and then replace the least recently used one it is allowed to move away (a setting to lock items would be useful). Tracking interactions instead of requests would allow respecting the user needing an item and we could even give user interactions some more importance because the user can't swap items back as easily as Baritone can.

ZeroMemes:
> Since I'll be away for the next couple days, will this be capable of doing what [this patch](https://github.com/ZacSharp/baritone/commit/42f5deae773c09f5ff67985bc75fc759a76b6a90) could do all the time (i.e. move items on demand)? So far you only seem to plan on cleaning up the old behavior.

Yes. I intend on moving all inventory management code to use `InventoryBehavior`, which includes `MovementHelper.switchToBestToolFor`, and will support moving items from the inventory when `allowInventory` is on.

> To find a slot to move the item to, instead of using hardcoded slots `InventoryBehavior` could track when each item was last requested and then replace the least recently used one it is allowed to move away (a setting to lock items would be useful).

The `allowHotbarManagement` setting that I currently have is purely to preserve old behavior, as I haven't put in place a way to reserve/lock slots or [mark them as important/useful](https://github.com/cabaletta/baritone/blob/pr/inventoryImprovements/src/main/java/baritone/process/BuilderProcess.java#L545-L571). The hard-coded reserved slots will not be staying.

> Tracking interactions instead of requests would allow respecting the user needing an item and we could even give user interactions some more importance because the user can't swap items back as easily as Baritone can.

Good idea 👍

ZeroMemes:
Tools are automatically moved from the inventory as of bc18f0eabd4be03eea6984d333326d8646140505

PR #4040, opened 2023-07-06
Elytra

ZeroMemes:
😼

still wip, needs ncp testing especially

TODO
- [ ] send individual block updates to nether-pathfinder instead of entire chunks
- [ ] have chunk pack queue limit (like cached chunks so it doesn’t fall behind)
- [ ] either limit usage to nether only, or have better overworld support
- [ ] communicate isSafeToCancel
- [ ] verify state management in process
- [ ] automatically equip elytra (when almost broken)
- [ ] reset nether-pathfinder context
- [ ] recalculate segment when no point visible
- [ ] proper goal arrival routine
- [ ] safe landing (relating to previous point)


leijurv:
biggest todo is making it work on 2b2t in practice, in the face of 1. lag 2. ncp

ZeroMemes:
Only checked off "proper goal arrival routine" since #4054 doesn't avoid landing into lava

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
```diff
@@ -0,0 +1,57 @@
...
+    private int prevChunkX = Integer.MAX_VALUE;
+    private int prevChunkZ = Integer.MAX_VALUE;
+
+    public BlockStateOctreeInterface(final NetherPathfinderContext context) {
+        this.context = context;
+        this.contextPtr = context.context;
+    }
+
+    public boolean get0(final int x, final int y, final int z) {
+        if ((y | (128 - y)) < 0) {
```
leijurv, line 42: should this be 127 instead of 128

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
```diff
@@ -0,0 +1,57 @@
...
+        final int chunkZ = z >> 4;
+        long pointer = this.chunkPtr;
+        if (pointer == 0 | ((chunkX ^ this.prevChunkX) | (chunkZ ^ this.prevChunkZ)) != 0) {
+            this.prevChunkX = chunkX;
+            this.prevChunkZ = chunkZ;
+            synchronized (this.context.cacheLock) {
+                this.chunkPtr = pointer = NetherPathfinder.getOrCreateChunk(this.contextPtr, chunkX, chunkZ);
+            }
+        }
+        return Octree.getBlock(pointer, x & 0xF, y & 0x7F, z & 0xF);
```
leijurv, line 55: Keeping `pointer` as a local variable does guarantee that pointer can't be zero at this point. However, pointer can be invalid. Since `Octree.getBlock` is called outside the `synchronized` block, it's possible that the `NetherPathfinderContext` could `cullFarChunks` after the `synchronized` block but before the `Octree.getBlock` which could cause a segfault.

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -0,0 +1,239 @@
...
+    public void queueBlockUpdate(BlockChangeEvent event) {
+        this.executor.execute(() -> {
+            ChunkPos chunkPos = event.getChunkPos();
+            long ptr = NetherPathfinder.getChunkPointer(this.context, chunkPos.x, chunkPos.z);
+            if (ptr == 0) return; // this shouldn't ever happen
+            event.getBlocks().forEach(pair -> {
+                BlockPos pos = pair.first();
+                if (pos.getY() >= 128) return;
+                boolean isSolid = pair.second() != AIR_BLOCK_STATE;
+                Octree.setBlock(ptr, pos.getX() & 15, pos.getY(), pos.getZ() & 15, isSolid);
```
leijurv, line 93: Does this need to `synchronize` on `cacheLock` to ensure that `ptr` is still valid and not culled?

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
ZeroMemes: Yes

src/main/java/baritone/utils/PathRenderer.java
```diff
@@ -127,12 +129,12 @@ private static void drawPath(IPath path, int startIndex, Color color, boolean fa
...
             int dirZ = end.z - start.z;
 
-            while (next + 1 < positions.size() && (!fadeOut || next + 1 < fadeStart) &&
+            /*while (next + 1 < positions.size() && (!fadeOut || next + 1 < fadeStart) &&
                     (dirX == positions.get(next + 1).x - end.x &&
                             dirY == positions.get(next + 1).y - end.y &&
                             dirZ == positions.get(next + 1).z - end.z)) {
                 end = positions.get(++next);
-            }
+            }*/
```
leijurv, line 137: this can probably be added back, or made conditional. i only commented it out so that the elytra segments would be clearly visible.

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
babbaj: probably yes but synchronizing every call might be expensive

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
babbaj: not inside the executor

ZeroMemes (commented):
good to go 😎

PR #4050, opened 2023-07-11
Merge 1.16.5 into 1.18.2

ZacSharp:
`MixinPalettedContainer` is a mess. Ideas for improvement are welcome. 
Having the shadow classes mentioned in https://github.com/SpongePowered/Mixin/issues/284 would be nice...

Draft because I haven't yet verified this works after running proguard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,16 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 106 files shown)


ZacSharp:
Nope, this line still references `baritone.utils.accessor.IPalettedContainer` after Proguard renamed the interface.
https://github.com/cabaletta/baritone/blob/71291d1c554ed88097e349113e7b1568b0a33e3b/src/launch/java/baritone/launch/mixins/MixinPalettedContainer.java#L88

Issue #4049, opened 2023-07-10
Support Vulkan


CormacZ:
## Describe your suggestion
There is a mod called Vulkan and I wonder if you can support Vulkan Rendering.

## Context
More support on more Rendering.

ZeroMemes:
Should be possible if all OpenGL references are replaced with the game's renderer.

ZeroMemes:
Nevermind, `DEBUG_LINES` usage needs to be replaced with `LINES`.

ZeroMemes:
Resolved by #4059

PR #4059, opened 2023-07-17
VulkanMod Compatibility

ZeroMemes:
Also fixes line width settings not working since the removal of legacy OpenGL


ZeroMemes:
Fixed #4049

ZacSharp (approved):
Looks good.

Why do we need to set normals now?

ZeroMemes:
> Why do we need to set normals now?

Because `RenderType.LINES` is implemented using triangle primitives, the normal is used by Mojang's shader to orient the line to face the camera.

PR #4062, opened 2023-07-21
Merge 1.18.2 into 1.19.4

ZacSharp:
The fake merge is to stop the revert commits from going into 1.19.4.

`MixinPalettedContainer` managed to annoy me even more. For the classloading problems I have a suspicion that mixin yelling something about classloader exclusions every startup might be connected to the cause of the problem, but I couldn't find the code adding the exclusion though so I couldn't test whether that's really the case.

@wagyourtail I got [this](https://github.com/ZacSharp/baritone/actions/runs/5594230806/job/15153024454) CI fail for two attempts in a row but not on any other commit and it seems to happen in unimined code so maybe you want to have a look at it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,16 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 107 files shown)


leijurv:
❤️

PR #4063, opened 2023-07-22
make elytra play nice with CustomGoalProcess

babbaj:


src/main/java/baritone/utils/PathingControlManager.java
```diff
@@ -109,10 +111,6 @@ public void preTick() {
                 p.cancelSegmentIfSafe();
                 break;
             case FORCE_REVALIDATE_GOAL_AND_PATH:
-                if (!p.isPathing() && !p.getInProgress().isPresent()) {
```
babbaj, line 112: the case below it is the same so i just got rid of this duplicate code and made it rely on fall through

PR #4064, opened 2023-07-23
find safe landing spot for elytra

babbaj:

### DIFF
--- a/src/main/java/baritone/process/ElytraProcess.java
+++ b/src/main/java/baritone/process/ElytraProcess.java
@@ -57,6 +57,7 @@ public class ElytraProcess extends BaritoneProcessHelper implements IBaritonePro
 
     public State state;
     private boolean goingToLandingSpot;
+    private BetterBlockPos landingSpot;
     private Goal goal;
     private LegacyElytraBehavior behavior;
 
@@ -102,7 +103,7 @@ public class ElytraProcess extends BaritoneProcessHelper implements IBaritonePro
 
         if (ctx.player().isElytraFlying() && this.state != State.LANDING) {
             final BetterBlockPos last = this.behavior.pathManager.path.getLast();
-            if (last != null && ctx.player().getDistanceSqToCenter(last) < (5 * 5)) {
+            if (last != null && ctx.player().getDistanceSqToCenter(last) < 1) {
                 if (Baritone.settings().notificationOnPathComplete.value) {
                     logNotification("Pathing complete", false);
                 }
@@ -113,22 +114,25 @@ public class ElytraProcess extends BaritoneProcessHelper implements IBaritonePro
                     return new PathingCommand(null, PathingCommandType.CANCEL_AND_SET_GOAL);
                 }
                 if (!goingToLandingSpot) {
-                    BlockPos landingSpot = findSafeLandingSpot();
+                    BetterBlockPos landingSpot = findSafeLandingSpot();
                     if (landingSpot != null) {
                         this.pathTo(landingSpot);
+                        this.landingSpot = landingSpot;
                         this.goingToLandingSpot = true;
                         return this.onTick(calcFailed, isSafeToCancel);
                     }
+                    // don't spam call findLandingSpot if it somehow fails (it's slow)
+                    this.goingToLandingSpot = true;
                 }
                 this.state = State.LANDING;
             }
         }
 
         if (this.state == State.LANDING) {
-            final BetterBlockPos endPos = behavior.pathManager.path.getLast();
+            final BetterBlockPos endPos = this.landingSpot != null ? this.landingSpot : behavior.pathManager.path.getLast();
             if (ctx.player().isElytraFlying() && endPos != null) {
                 Vec3d from = ctx.player().getPositionVector();
-                Vec3d to = new Vec3d(endPos.x, from.y, endPos.z);
+                Vec3d to = new Vec3d(((double) endPos.x) + 0.5, from.y, ((double) endPos.z) + 0.5);
                 Rotation rotation = RotationUtils.calcRotationFromVec3d(from, to, ctx.playerRotations());
                 baritone.getLookBehavior().updateTarget(rotation, false);
             } else {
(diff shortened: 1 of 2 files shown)


src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +349,58 @@ public double placeBucketCost() {
...
+    }
+
+    private boolean isSafeLandingSpot(BlockPos pos) {
+        BlockPos.MutableBlockPos mut = new BlockPos.MutableBlockPos(pos);
+        while (mut.getY() >= 0) {
+            IBlockState state = ctx.world().getBlockState(mut);
+            if (state.getMaterial().isLiquid()) { // lava
+                return false;
+            }
+            if (state.getMaterial().blocksMovement() && state.getBlock() != Blocks.MAGMA) {
```
leijurv, line 376: not quite right i think. magma should be in previous check. the way it's written it will skip magma entirely, while really it should treat magma like lava.

leijurv: actually nvm this entire thing is cringe

for example it doesn't consider fire properly

instead make it like:
if air { continue onward }
if netherrack { this is a valid landing spot }
if anything else { exit and return false }

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +349,58 @@ public double placeBucketCost() {
...
+        Set<BetterBlockPos> visited = new HashSet<>();
+        queue.add(start);
+
+        while (!queue.isEmpty()) {
+            BetterBlockPos pos = queue.poll();
+            if (ctx.world().isBlockLoaded(pos) && isInBounds(pos) && ctx.world().getBlockState(pos).getBlock() == Blocks.AIR) {
+                if (isSafeLandingSpot(pos)) {
+                    return pos;
+                }
+                if (visited.add(pos.north())) queue.add(pos.north());
```
leijurv, line 399: 3d floodfill is fine, but, for the actual isSafeLandingSpot check, it should be slightly changed. rn it will uselessly check `(0,64,0)`, `(0,65,0)`, etc while they all are the same. there should be a new Set<BetterBlockPos> alreadyChecked. it's passed into isSafeLandingSpot and if its mutableblockpos is within that set, it exits early. the input pos is added to the set at the top of isSafeLandingSpot (it adds only the top of the column, NOT the entire column). this will work efficiently because pos.up is added to the queue before pos.down, therefore it will tend towards checking the tops of columns first.

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +349,58 @@ public double placeBucketCost() {
...
+        final BetterBlockPos start = ctx.playerFeet();
+        Queue<BetterBlockPos> queue = new LinkedList<>();
+        Set<BetterBlockPos> visited = new HashSet<>();
+        queue.add(start);
+
+        while (!queue.isEmpty()) {
+            BetterBlockPos pos = queue.poll();
+            if (ctx.world().isBlockLoaded(pos) && isInBounds(pos) && ctx.world().getBlockState(pos).getBlock() == Blocks.AIR) {
+                if (isSafeLandingSpot(pos)) {
+                    return pos;
```
leijurv, line 397: shouldn't this return `mut`, the actual safe spot that isSafeLandingSpot located?

babbaj: pos will be in the air (this is lazy and i should find a specific spot above it)

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +351,59 @@ public double placeBucketCost() {
...
+                || ctx.world().isAirBlock(pos.east())
+                || ctx.world().isAirBlock(pos.west())
+                // corners
+                || ctx.world().isAirBlock(pos.north().west())
+                || ctx.world().isAirBlock(pos.north().east())
+                || ctx.world().isAirBlock(pos.south().west())
+                || ctx.world().isAirBlock(pos.south().east());
+    }
+
+    private boolean isSafeLandingSpot(BlockPos pos) {
```
leijurv, line 371:
```suggestion
    private boolean isSafeLandingSpot(BlockPos pos, Set<BlockPos> landingSpotsChecked) {
        landingSpotsChecked.add(pos);
```

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +351,59 @@ public double placeBucketCost() {
...
+        while (mut.getY() >= 0) {
+            IBlockState state = ctx.world().getBlockState(mut);
+            Block block = state.getBlock();
+
+            if (block == Blocks.NETHERRACK || block == Blocks.GRAVEL || block == Blocks.NETHER_BRICK) {
+                return true;
+            } else if (block != Blocks.AIR) {
+                return false;
+            }
+            mut.setPos(mut.getX(), mut.getY() - 1, mut.getZ());
```
leijurv, line 382:
```suggestion
            mut.setPos(mut.getX(), mut.getY() - 1, mut.getZ());
            if (landingSpotsChecked.contains(mut)) return false;
```

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +351,59 @@ public double placeBucketCost() {
...
+        }
+        return false; // void
+    }
+
+    private BetterBlockPos findSafeLandingSpot() {
+        final BetterBlockPos start = ctx.playerFeet();
+        Queue<BetterBlockPos> queue = new LinkedList<>();
+        Set<BetterBlockPos> visited = new HashSet<>();
```
leijurv, line 390:
```suggestion
        Set<BetterBlockPos> visited = new HashSet<>();
        Set<BetterBlockPos> landingSpotsChecked = new HashSet<>();
```

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +351,59 @@ public double placeBucketCost() {
...
+        queue.add(start);
+
+        while (!queue.isEmpty()) {
+            BetterBlockPos pos = queue.poll();
+            if (ctx.world().isBlockLoaded(pos) && isInBounds(pos) && ctx.world().getBlockState(pos).getBlock() == Blocks.AIR) {
+                if (isSafeLandingSpot(pos)) {
```
leijurv, line 396:
```suggestion
                if (isSafeLandingSpot(pos, landingSpotsChecked)) {
```

src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -329,4 +351,59 @@ public double placeBucketCost() {
...
+                return false;
+            }
+            mut.setPos(mut.getX(), mut.getY() - 1, mut.getZ());
+        }
+        return false; // void
+    }
+
+    private BetterBlockPos findSafeLandingSpot() {
+        final BetterBlockPos start = ctx.playerFeet();
+        Queue<BetterBlockPos> queue = new LinkedList<>();
```
leijurv, line 395:
```suggestion
        Queue<BetterBlockPos> queue = new PriorityQueue(Comparator.comparingInt(pos -> pos.getY()).reversed());
```

maybe. solid maybe. definite maybe. definitely test this before doing it.

Issue #3599, opened 2022-08-08
`goto` command should accept multiple arguments like `mine`


Picklejar05:
## Describe your suggestion
`goto` should allow you to enter multiple target blocks, similarly to `mine`. It would go to the closest valid target in the same way. (also add the feature to this command that I describe in #3598)

## Context
This would be helpful if you want to reach any of multiple types of blocks, such as for example `goto crafting_table furnace bed chest shulker_box` to find any bases or villiages.

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

PR #4104, opened 2023-08-17
Add partial support for multiple #goto targets

Cyclotome:
Goto will work with multiple block arguments.
Special actions (ex. right click on arrival) will only trigger if they apply to all arguments.
(enhancement requested in #3599 )

### DIFF
--- a/src/api/java/baritone/api/process/IGetToBlockProcess.java
+++ b/src/api/java/baritone/api/process/IGetToBlockProcess.java
@@ -19,17 +19,21 @@ package baritone.api.process;
 
 import baritone.api.utils.BlockOptionalMeta;
 import net.minecraft.block.Block;
+import java.util.List;
+import java.util.Arrays;
 
 /**
  * but it rescans the world every once in a while so it doesn't get fooled by its cache
  */
 public interface IGetToBlockProcess extends IBaritoneProcess {
 
-    void getToBlock(BlockOptionalMeta block);
-
+    default void getToBlock(BlockOptionalMeta block) {
+        getToBlock(Arrays.asList(block));
+    }
     default void getToBlock(Block block) {
         getToBlock(new BlockOptionalMeta(block));
     }
+    void getToBlock(List<BlockOptionalMeta> blocks);
 
     boolean blacklistClosest();
 }
--- a/src/main/java/baritone/command/defaults/GotoCommand.java
+++ b/src/main/java/baritone/command/defaults/GotoCommand.java
@@ -29,6 +29,7 @@ import baritone.api.utils.BetterBlockPos;
 import baritone.api.utils.BlockOptionalMeta;
 
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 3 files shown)


src/api/java/baritone/api/process/IGetToBlockProcess.java
```diff
@@ -19,17 +19,21 @@
...
 
-    void getToBlock(BlockOptionalMeta block);
-
+    default void getToBlock(BlockOptionalMeta block) {
+        getToBlock(Arrays.asList(block));
+    }
     default void getToBlock(Block block) {
         getToBlock(new BlockOptionalMeta(block));
     }
+    void getToBlock(List<BlockOptionalMeta> blocks);
```
ZacSharp, line 36: The `List<BlockOptionalMeta>` should be a `BlockOptionalMetaLookup`

src/main/java/baritone/process/GetToBlockProcess.java
```diff
@@ -187,22 +193,23 @@ public synchronized void onLostControl() {
...
     private synchronized void rescan(List<BlockPos> known, CalculationContext context) {
-        List<BlockPos> positions = MineProcess.searchWorld(context, new BlockOptionalMetaLookup(gettingTo), 64, known, blacklist, Collections.emptyList());
+        List<BlockPos> positions = MineProcess.searchWorld(context, new BlockOptionalMetaLookup(gettingTo.toArray(new BlockOptionalMeta[0])), 64, known, blacklist, Collections.emptyList());
         positions.removeIf(blacklist::contains);
         knownLocations = positions;
     }
 
     private Goal createGoal(BlockPos pos) {
-        if (walkIntoInsteadOfAdjacent(gettingTo.getBlock())) {
+        // determines which actions to take (action done iff should be done for all target blocks))
```
ZacSharp, line 208: I think it's better to assume unclickable for unknown positions and decide per block for known positions.

src/main/java/baritone/command/defaults/GotoCommand.java
ZacSharp, line 1: You'll also need to update the tab completion and help text.

src/main/java/baritone/process/GetToBlockProcess.java
```diff
@@ -36,14 +36,19 @@
...
-    private BlockOptionalMeta gettingTo;
+    private List<BlockOptionalMeta> gettingTo;
     private List<BlockPos> knownLocations;
     private List<BlockPos> blacklist; // locations we failed to calc to
     private BlockPos start;
 
+    private String strGettingTo() {
+        return gettingTo.stream().map(bom -> bom.toString()).collect(Collectors.joining("/"));
+    }
+
```
ZacSharp, line 51: Just use the string representation of the `BlockOptionalMetaLookup`.

src/main/java/baritone/process/GetToBlockProcess.java
```diff
@@ -113,7 +118,8 @@ public double heuristic() {
         }
         if (goal.isInGoal(ctx.playerFeet()) && goal.isInGoal(baritone.getPathingBehavior().pathStart()) && isSafeToCancel) {
             // we're there
-            if (rightClickOnArrival(gettingTo.getBlock())) {
+            // currently don't know matched position/BOM, so will only right click if all BOM's in gettingTo are right clickable
```
ZacSharp, line 121: While this does certainly make things easier it will almost certainly result in false bug reports.
Maybe try clicking on any target location for which `createGoal(pos).isInGoal(playerPos)` is true and if there is no such position succeed right away.

ZacSharp (changes requested):

PR #4107, opened 2023-08-19
1.19 buildscript update

wagyourtail:
use proguard-gradle, only run proguard once for api/standalone and remap+shade the other targets in.


ZacSharp:
Holy shit...
I'm happy it works (no tests done yet, trusting your testing so far), even though I'll have to read it a dozen more times to understand it. You definitely got me feel bad about you spending so much effort on an idea that took me only a couple minutes of thinking + writing.

Issue #4098, opened 2023-08-15
ascend vines


leijurv:
vines can now be ascended by holding space. this should be supported

ZacSharp:
Scaffolding as well

Nekiplay:
I had support for scaffolding in 1.20.2 but I lost the sursor.

PR #4111, opened 2023-08-22
Add support for nether vines

scorbett123:
Make baritone be able to climb warped and twisting vines


leijurv:
Nice! Per #4098 can it also climb normal vines by holding space?

scorbett123:
Lol, only took 3 years to get merged

PR #4136, opened 2023-09-23
1.20.2

rfresh2:
To be put on a new branch


rfresh2:
Applied forge crash patch that also affected 1.20.1 in https://github.com/cabaletta/baritone/pull/4138

Issue #4133, opened 2023-09-17
How can I turn off the elytra logs. it clogs up my chat


ghost:
## What do you need help with?
Turning off the "debug" chat when I use #elytra

## Final checklist
- [x] I know how to properly use check boxes 
- [X ] I have not used any OwO's or UwU's in this issue.

ghost:
i dont know how to use checklists :(

ZacSharp:
You can't because for some reason those messages aren't debug messages. Otherwise you could toggle them with `chatDebug` which is disabled by default.

Relabeling to bug because "next segment completed" and similar messages clearly do not fit "[should only be used for critically important messages, or as a direct response to a chat command](https://github.com/cabaletta/baritone/blob/81932069d98eebb6e0cf81efe9aeb279594c1c40/src/api/java/baritone/api/utils/Helper.java#L192)".

ZacSharp:
Whoops, didn't mean to close this.

ghost:
Thanks. Because when all of those messages pop up in my chat it floods it and I can barely read chat if someone messages me or someone trys to talk to me but I cant even see them

ZacSharp:
Fixed in #4137

PR #4137, opened 2023-09-23
silence elytra chat spam

babbaj:


ZacSharp:
There's `logDebug`?
Having some more fine grained control over which debug messages you get would be nice (ever tried spotting the non-goal lines when `BuilderProcess` has 100+ goals?) but I think a single `Set` type setting would be less clutter.

Also closes #4133 either way

PR #4153, opened 2023-10-12
Adds git to the Dockerfile apt install (Windows build functionality)

xiaodown:
I saw in setup.md "(if you have docker on Windows, I'd be grateful if you could let me know if it works there too)".  So I went to build it on Windows.

I use Windows Subsystem for Linux, which is mostly Ubuntu but not really (there's no init).  But Docker Desktop has an integration for WSL2.

For me, gradle wouldn't build the image without forcing git to install in the container.  I'll attach screenshots to the pull request.  It just looks like whatever ubuntu:focal my docker system grabbed doesn't have git installed by default.

Anyway, this fixes it, and shouldn't screw anything even if git is installed by default.  I tested it on my M1 Mac pro and it still works fine with this change.

And you can (somewhat) say building works on Windows, or at least WSL2.

1.19.4 trying to build:

this branch building:
mybranch

Docker desktop on Windows showing the image:
image-on-win

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN apt update -y
 
 RUN apt install \
           openjdk-17-jdk \
+	  git \
           --assume-yes
 
 COPY . /code


xiaodown:
https://learn.microsoft.com/en-us/windows/wsl/install and https://docs.docker.com/desktop/wsl/ btw if interested.   Also the new Windows Terminal is ... actually really good (https://aka.ms/terminal or https://github.com/microsoft/terminal if you don't want to use the windows app store).

xiaodown:
image

^^^ container running + terminal showing build in progress (in windows)

leijurv:
Cool!

PR #4168, opened 2023-10-26
Improve version detection

ZacSharp:
1. No longer uses tags from merged-in branches. E.g. the head commit of this pr is "v1.9.2-22-g9accb2bfb" instead of "v1.2.19-572-g9accb2bfb". (Note: CI builds use "v1.9.3-something" because CI runs on a temporary merge commit)
2. No longer crashes if git is not installed.


leijurv:
Ooh, thank you so much! This was bugging me lol

PR #4178, opened 2023-11-05
Support for building selected litematic

thojo0:
Adds the buildOpenLitematic method without index to check for the currently selected litematic

I think this is more intuitive to use the `@litematic` command instead of just using the first loaded placement if no argument is passed

### DIFF
--- a/src/main/java/baritone/command/defaults/LitematicaCommand.java
+++ b/src/main/java/baritone/command/defaults/LitematicaCommand.java
@@ -40,11 +40,13 @@ public class LitematicaCommand extends Command {
             if (args.is(Integer.class)) {
                 schematic = args.getAs(Integer.class) - 1;
             }
-        }
-        try {
-            baritone.getBuilderProcess().buildOpenLitematic(schematic);
-        } catch (IndexOutOfBoundsException e) {
-            logDirect("Pleas provide a valid index.");
+            try {
+                baritone.getBuilderProcess().buildOpenLitematic(schematic);
+            } catch (IndexOutOfBoundsException e) {
+                logDirect("Pleas provide a valid index.");
+            }
+        } else {
+            baritone.getBuilderProcess().buildOpenLitematic();
         }
     }
 
--- a/src/main/java/baritone/process/BuilderProcess.java
+++ b/src/main/java/baritone/process/BuilderProcess.java
@@ -216,6 +216,20 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
     }
 
     @Override
+    public void buildOpenLitematic() {
+        if (LitematicaHelper.isLitematicaPresent()) {
+            Integer selectedIndex = LitematicaHelper.getSelectedIndex();
+            if (selectedIndex != -1) {
+                buildOpenLitematic(selectedIndex);
+            } else {
+                logDirect("No schematic currently selected");
+            }
+        } else {
+            logDirect("Litematica is not present");
+        }
+    }
+
+    @Override
     public void buildOpenLitematic(int i) {
         if (LitematicaHelper.isLitematicaPresent()) {
             //if java.lang.NoSuchMethodError is thrown see comment in SchematicPlacementManager
@@ -532,7 +546,7 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
         }
 
         Optional<Tuple<BetterBlockPos, Rotation>> toBreak = toBreakNearPlayer(bcc);
-        if (toBreak.isPresent() && isSafeToCancel && ctx.player().isOnGround()) {
+        if (toBreak.isPresent() && isSafeToCancel && ctx.player().onGround()) {
             // we'd like to pause to break this block
             // only change look direction if it's safe (don't want to fuck up an in progress parkour for example
             Rotation rot = toBreak.get().getB();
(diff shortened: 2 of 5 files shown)


src/main/java/baritone/command/defaults/LitematicaCommand.java
```diff
@@ -40,11 +40,13 @@ public void execute(String label, IArgConsumer args) throws CommandException {
...
-            baritone.getBuilderProcess().buildOpenLitematic(schematic);
-        } catch (IndexOutOfBoundsException e) {
-            logDirect("Pleas provide a valid index.");
+            try {
+                baritone.getBuilderProcess().buildOpenLitematic(schematic);
+            } catch (IndexOutOfBoundsException e) {
+                logDirect("Pleas provide a valid index.");
+            }
+        } else {
+            baritone.getBuilderProcess().buildOpenLitematic();
```
rycbar0, line 49: `baritone.getBuilderProcess().buildOpenLitematic(LitematicaHelper.getSelectedIndex());`
this would remove the need of a second buildOpenLitematic method in BuilderProcess however it also wouldnt be wrong to remove the manual selection entirely and just keep the one that builds the schematic litematica has selected.

thojo0: jeah that's right, i added the second one with the idea that you can easily use it through the api, but of course you can just send the command directly to the execute function, what works better anyways often.

generally adding options i think is good but for intuition maybe only using the selected one makes sense.

ZacSharp: What about building placements by name? It's intuitive and with proper tab completion is't easy to use.

For easier API usage I think the best option would be a method taking the desired `SchematicPlacement` (Why are we even using indices in the first place?). Adding `buildOpenLitematic()` is good for consistency with `buildOpenSchematic()` though.

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -532,7 +546,7 @@ public int lengthZ() {
         }
 
         Optional<Tuple<BetterBlockPos, Rotation>> toBreak = toBreakNearPlayer(bcc);
-        if (toBreak.isPresent() && isSafeToCancel && ctx.player().isOnGround()) {
+        if (toBreak.isPresent() && isSafeToCancel && ctx.player().onGround()) {
```
ZacSharp: Is there a reason for this change? It looks unrelated.

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -552,7 +566,7 @@ public int lengthZ() {
         }
         List<BlockState> desirableOnHotbar = new ArrayList<>();
         Optional<Placement> toPlace = searchForPlacables(bcc, desirableOnHotbar);
-        if (toPlace.isPresent() && isSafeToCancel && ctx.player().isOnGround() && ticks <= 0) {
+        if (toPlace.isPresent() && isSafeToCancel && ctx.player().onGround() && ticks <= 0) {
```
ZacSharp: Same here.

src/main/java/baritone/process/BuilderProcess.java
```diff
@@ -1063,9 +1077,6 @@ private boolean valid(BlockState current, BlockState desired, boolean itemVerify
         if (desired.getBlock() instanceof AirBlock && Baritone.settings().buildIgnoreBlocks.value.contains(current.getBlock())) {
             return true;
         }
-        if (!(current.getBlock() instanceof AirBlock) && Baritone.settings().buildIgnoreExisting.value && !itemVerify) {
-            return true;
-        }
```
ZacSharp: Please don't break `buildIgnoreExisting` again.

ZacSharp (changes requested):

src/main/java/baritone/process/BuilderProcess.java
thojo0: oh ups no maybe this happend because i tested it in 1.20.1 build

src/main/java/baritone/process/BuilderProcess.java
thojo0: ah sorry this all happened because i tested it it with 1.20.1 build will fix that

src/main/java/baritone/command/defaults/LitematicaCommand.java
rycbar0: names wont work well because you can have multiple schematics with the same name open. how would you decide which schematic is meant if you have 2 "myHouse.litematic"
as to why integers are used instead of SchematicPlacements thats because SchematicPlacement is a class from litematica and i didnt want to add a dependency to a optional mod to the BuilderProcess this way the only Class with a direct dependency on litematica should be LitematicaHelper and methods of that class should only be called after the LitematicaHelper.isLitematicaPresent() method returned true

thojo0: i added an branch with only the selected placement [thojo0/baritone/only_selected_litematic](/thojo0/baritone/tree/only_selected_litematic)
i will add a pull request if requested for that.

ZacSharp: > names wont work well because you can have multiple schematics with the same name open

We are talking about placements here and I though litematica enforces unique names on those, like it does for selection and subregion names. But apparently you can have indistinguishable placements. Reporting an error for duplicates would be an option, though it's certainly not as nice anymore.

PR #4172, opened 2023-10-29
Merge 1.19.4 into 1.20.2

ZacSharp:
#4112 is the sole reason for this pr. Otherwise I'd wait until there's more things to merge.
EDIT: totally forgot about 91588ca even though I looked at it's diff 30 seconds ago...

PR #4195, opened 2023-11-16
Fix scanning chunks using a SingleValuePalette

ZacSharp:
Completely special cased because there is no data to iterate over.

Reported in https://github.com/cabaletta/baritone/pull/4172#issuecomment-1787598593.

### DIFF
--- a/src/main/java/baritone/cache/FasterWorldScanner.java
+++ b/src/main/java/baritone/cache/FasterWorldScanner.java
@@ -38,6 +38,7 @@ import net.minecraft.world.level.chunk.LevelChunk;
 import net.minecraft.world.level.chunk.LevelChunkSection;
 import net.minecraft.world.level.chunk.Palette;
 import net.minecraft.world.level.chunk.PalettedContainer;
+import net.minecraft.world.level.chunk.SingleValuePalette;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


0-x-2-2:
Can confirm this was an issue, we switched from FasterWorldScanner to the legacy one downstream.

src/main/java/baritone/cache/FasterWorldScanner.java
```diff
@@ -191,7 +192,29 @@ private void visitSection(BlockOptionalMetaLookup lookup, LevelChunkSection sect
...
+        Palette<BlockState> palette = ((IPalettedContainer<BlockState>) sectionContainer).getPalette();
+
+        if (palette instanceof SingleValuePalette) {
+            // single value palette doesn't have any data
+            if (lookup.has(palette.valueFor(0))) {
+                // TODO this is 4k hits, maybe don't return all of them?
+                for (int x = 0; x < 16; ++x) {
+                    for (int y = 0; y < 16; ++y) {
+                        for (int z = 0; z < 16; ++z) {
+                            blocks.add(new BlockPos(
```
0-x-2-2, line 205: Does it use SingleValuePalette if the section is all entirely the same block?

ZacSharp: Yes

PR #4205, opened 2023-11-23
Add issues link and link to Discord for ModMenu

Jatc252:
As title says. I had baritone load with [ModMenu](https://modrinth.com/mod/modmenu) and noticed that there was some extra information that could be added.

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -11,7 +11,8 @@
   ],
   "contact": {
     "homepage": "https://github.com/cabaletta/baritone",
-    "sources": "https://github.com/cabaletta/baritone"
+    "sources": "https://github.com/cabaletta/baritone",
+    "issues": "https://github.com/cabaletta/baritone/issues" 
   },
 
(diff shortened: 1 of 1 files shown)


Julienraptor01:
one other thing to do could be adding an icon so ModMenu stop throwing a warning and an error about it and actually display smth instead of the default missing icon
maybe the discord server icon could be used as mod icon
webp version stored on discord servers :
[icon](https://cdn.discordapp.com/icons/822011099561197579/6a3bf431a25c34057696668763dd48fb.webp?size=512)

Jatc252:
> one other thing to do could be adding an icon so ModMenu stop throwing a warning and an error about it and actually display smth instead of the default missing icon
> 
> maybe the discord server icon could be used as mod icon
> 
> webp version stored on discord servers :
> 
> [icon](https://cdn.discordapp.com/icons/822011099561197579/6a3bf431a25c34057696668763dd48fb.webp?size=512)
> 
> 

Probably best to leave for the devs to pick their own (higher quality) image to use if they ever do it, but good idea

PR #4244, opened 2024-01-20
fix 🪨🧠 preventing null elytra process from working

wagyourtail:


ZacSharp:
https://github.com/wagyourtail/baritone/blob/340d2a1d54e58df8e6d6c61c9eb1ff1ae3091642/src/main/java/baritone/process/ElytraProcess.java#L87

Is there any reason for this cast?

aflolfin:
what does that mean and how do i change it

PR #4251, opened 2024-01-22
Fix feather crashing by catching error

wagyourtail:
because for some reason, feather waits until a little bit later than injecting it to make their ChatHud replacement actually work.


leijurv:
i don't love this but whatever

PR #4291, opened 2024-03-10
Block break delay setting

rfresh2:
Fixes baritone flagging [FastBreak](https://github.com/GrimAnticheat/Grim/blob/2.0/src/main/java/ac/grim/grimac/checks/impl/misc/FastBreak.java#L80-L92), causing it to get stuck in an endless loop sending stop break packets.

Due to how baritone calls start break slightly different (more direct) from vanilla, there is only 2 ticks from stop break to start break instead of vanilla's 6 ticks.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -385,6 +385,13 @@ public final class Settings {
      */
     public final Setting<Float> blockReachDistance = new Setting<>(4.5f);
 
+
+    /**
+     * Delay between breaking a block and starting to break the next block. The vanilla delay is 6 ticks.
+     * Baritone waits an additional 2 ticks on top of this setting value.
+     */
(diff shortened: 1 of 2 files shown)


ZacSharp:
Feels a bit odd how the actual delay is 2 more than the setting value. Is it like this for the place delay as well?
If not I think it would be a good idea to have the setting be the actual delay and document that values lower than 2 will be clamped.

rfresh2:
> Is it like this for the place delay as well

With rightClickSpeed = 4, i'm measuring a 5 tick delay between place packets

> have the setting be the actual delay

yeah should be more clear for users, i'll update it. I think changing the rightClickSpeed setting similarly would make sense too.

ZacSharp (approved):
> I think changing the rightClickSpeed setting similarly would make sense too.

Agreed

PR #4292, opened 2024-03-11
align rightClickSpeed setting with actual delay

rfresh2:
Aligns ticks between place packets with the setting value. The default rightClickSpeed value of 4 ticks is the correct vanilla number as tested.

referenced in: https://github.com/cabaletta/baritone/pull/4291#issuecomment-1987408744


ZacSharp (approved):

ZacSharp:
Actually, there is no tick logic causing delay. It's just that `rightClickSpeed` is implemented as "ticks between packets" instead of "ticks per packet".

PR #4320, opened 2024-04-01
don't apply block break delay to insta-breaks

rfresh2:
A block break delay setting was added in https://github.com/cabaletta/baritone/pull/4291

This 6 tick delay is correct - but only for blocks that are broken in 1 + N ticks 

* 1 tick breaks send one START_DESTROY_BLOCK action packet per block which acts as both the start and end of the break. Any following START_DESTROY_BLOCK should not be delayed
* 1 + N tick breaks send a START_DESTROY_BLOCK packet followed by STOP_DESTROY_BLOCK when the break completes. Any following START_DESTROY_BLOCK should not be sent until the delay is completed

I've also updated the logic and method names here to fix some issues and align more with the 1.19+ mojmap names. 

The renames aren't required but imo helps for understanding the vanilla mc methods that are being called. If these need to be kept for backwards compatibility they can be reverted.

### DIFF
--- a/src/main/java/baritone/utils/BlockBreakHelper.java
+++ b/src/main/java/baritone/utils/BlockBreakHelper.java
@@ -29,10 +29,10 @@ import net.minecraft.world.phys.HitResult;
  */
 public final class BlockBreakHelper {
     // base ticks between block breaks caused by tick logic
-    private static final int BASE_BREAK_DELAY = 2;
+    private static final int BASE_BREAK_DELAY = 1;
 
     private final IPlayerContext ctx;
-    private boolean didBreakLastTick;
+    private boolean wasDestroyingBlockLastTick;
     private int breakDelayTimer = 0;
 
     BlockBreakHelper(IPlayerContext ctx) {
@@ -41,13 +41,10 @@ public final class BlockBreakHelper {
 
     public void stopBreakingBlock() {
         // The player controller will never be null, but the player can be
-        if (ctx.player() != null && didBreakLastTick) {
-            if (!ctx.playerController().hasBrokenBlock()) {
-                // insane bypass to check breaking succeeded
-                ctx.playerController().setHittingBlock(true);
-            }
+        if (ctx.player() != null && wasDestroyingBlockLastTick) {
+            ctx.playerController().setDestroyingBlock(false);
             ctx.playerController().resetBlockRemoving();
-            didBreakLastTick = false;
+            wasDestroyingBlockLastTick = false;
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/baritone/utils/BlockBreakHelper.java
```diff
@@ -41,13 +41,10 @@ public final class BlockBreakHelper {
...
     public void stopBreakingBlock() {
         // The player controller will never be null, but the player can be
-        if (ctx.player() != null && didBreakLastTick) {
-            if (!ctx.playerController().hasBrokenBlock()) {
-                // insane bypass to check breaking succeeded
-                ctx.playerController().setHittingBlock(true);
-            }
+        if (ctx.player() != null && wasDestroyingBlockLastTick) {
+            ctx.playerController().setDestroyingBlock(false);
             ctx.playerController().resetBlockRemoving();
```
ZacSharp: This changed from conditionally restoring the breaking flag and then cancelling the breaking process to unconditionally clearing the flag before cancelling. Is this intended?

src/main/java/baritone/utils/BlockBreakHelper.java
```diff
@@ -29,10 +29,10 @@
  */
 public final class BlockBreakHelper {
     // base ticks between block breaks caused by tick logic
-    private static final int BASE_BREAK_DELAY = 2;
+    private static final int BASE_BREAK_DELAY = 1;
 
     private final IPlayerContext ctx;
-    private boolean didBreakLastTick;
+    private boolean wasDestroyingBlockLastTick;
```
ZacSharp: To be honest I think the new name primarily is more verbose.

src/main/java/baritone/utils/BlockBreakHelper.java
```diff
@@ -60,24 +57,30 @@ public void tick(boolean isLeftClick) {
         boolean isBlockTrace = trace != null && trace.getType() == HitResult.Type.BLOCK;
 
         if (isLeftClick && isBlockTrace) {
-            if (!didBreakLastTick) {
+            ctx.playerController().setDestroyingBlock(wasDestroyingBlockLastTick);
+            if (!ctx.playerController().isDestroyingBlock()) {
                 ctx.playerController().syncHeldItem();
                 ctx.playerController().clickBlock(((BlockHitResult) trace).getBlockPos(), ((BlockHitResult) trace).getDirection());
                 ctx.player().swing(InteractionHand.MAIN_HAND);
+            } else {
```
ZacSharp: This changed from immediately calling `onPlayerDamageBlock` to calling it the first time in the next tick. This is the packet change you mentioned?

ZacSharp (commented):
Does this consider what happens if the user presses LMB?

I'm not sure about the renames. Some of the old names are indeed weird and Baritone seems to not have a stable API, but then again, why have an API package if other mods can't rely on it more than on the internal packages?
Also I don't think those methods should be API so maybe at least not add a a new accessor there?

src/main/java/baritone/utils/BlockBreakHelper.java
rfresh2: The `hasBrokenBlock` method didn't work correctly before so there's not really any change in behavior. 

Previous Baritone code for reference:

```
   @Override
    public boolean hasBrokenBlock() {
        return ((IPlayerControllerMP) mc.gameMode).getCurrentBlock().getY() == -1;
    }
```

Couple issues:
1. Modern MC allows Y level < 0 now so a -1 Y is a valid break position
2. The gameMode class this variable is in never resets the BlockPos variable when a block is broken. So what this is checking has no bearing on if a block was broken, is being broken, or really anything.

The `wasDestroyingBlockLastTick` stores all the state we we need to determine if we actually have an ongoing baritone break.

src/main/java/baritone/utils/BlockBreakHelper.java
rfresh2: I guess. My only intention here was to align this with the mojmap variable name in the gameMode class

src/main/java/baritone/utils/BlockBreakHelper.java
rfresh2: Yeah we don't need to call `onPlayerDamageBlock` for 1-tick breaks. The block is already broken after `clickBlock`. And we don't need to send more break state changes in `onPlayerDamageBlock` until the 2nd tick and onward if its not a 1-tick break.

rfresh2:
> Does this consider what happens if the user presses LMB?

Same behavior as it was previously, users can press LMB and cause more breaks or interrupt ongoing breaks. Baritone doesn't intercept click events or cancel the internal MC methods those call.

> I'm not sure about the renames. Some of the old names are indeed weird and Baritone seems to not have a stable API, but then again, why have an API package if other mods can't rely on it more than on the internal packages?

I found it difficult to trace what was being called as the API has drifted from the mojmap method/variable names since created in 2019. But if we want to preserve the API stability, fine. I don't feel strongly either way.

> Also I don't think those methods should be API so maybe at least not add a a new accessor there?

sure

src/main/java/baritone/utils/BlockBreakHelper.java
ZacSharp: what about "wasDestroying"? It's two words, not longer and uses the clearer terminology.

ZacSharp:
Ok, looks fine as far as I can tell (but I don't understand the logic in detail).

Last question: How much of this applies to older versions (1.12.2 specifically)?

rfresh2:
> How much of this applies to older versions (1.12.2 specifically)?

Ignoring all the variable and method names I don't think there's much difference in the MC logic. But that's just based on me briefly looking through the 1.12 src

src/main/java/baritone/utils/BlockBreakHelper.java
rfresh2: If we're reverting the API name changes maybe `wasBreaking` or `wasHitting`

Nekiplay:
On some servers these changes have started to cause teleports back from anti-cheat especially when the baritone digs netherrack to the bottom

rfresh2:
This is the vanilla break speed limit. And this has been tested on Grim/2b2t and vanilla. 

you need to provide more information about what server and anticheat you're having issues with if you want anyone to help you

PR #4363, opened 2024-05-16
[WIP] handle multi subregion litematica schematics returning null blocks

tolland:
If the litematica schematic has multiple subregions that don't correspond to the full extent of the schematic, then "state" is null so the build crashes.


ZacSharp (changes requested):
`placeAt` only queries the block after making sure it is in bounds so a null state at this point is a bug in `LitematicaSchematic`.

Also `logDirect` should be used for direct feedback and important things like error messages, not for borderline spam which can trigger 20 times per second.

tolland:
### Background and reproduction

Try to build this litematica schematic (which has 2 subregions) using baritone 1.10.2 under minecraft 1.20.4:

[subregion test1.litematic.zip](https://github.com/cabaletta/baritone/files/15445825/subregion.test1.litematic.zip)

It looks like this:


loaded schematic:
image

try and build with baritone

```
# litematca
```

It crashes with null state for block error...

```
[03:10:13] [Render thread/INFO]: [CHAT] [Baritone] > litematica
[03:10:13] [Render thread/INFO]: [CHAT] [Baritone] Starting layer 0
[03:10:13] [Render thread/ERROR]: Unreported exception thrown!
java.lang.NullPointerException: Cannot invoke "net.minecraft.class_2680.method_26204()" because "desired" is null
	at baritone.api.schematic.SubstituteSchematic.desiredState(SubstituteSchematic.java:51) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess$2.desiredState(BuilderProcess.java:475) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess$BuilderCalculationContext.getSchematic(BuilderProcess.java:1104) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess.recalcNearby(BuilderProcess.java:644) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess.recalc(BuilderProcess.java:620) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess.onTick(BuilderProcess.java:505) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess.onTick(BuilderProcess.java:509) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.process.BuilderProcess.onTick(BuilderProcess.java:439) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.utils.PathingControlManager.executeProcesses(PathingControlManager.java:199) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.utils.PathingControlManager.preTick(PathingControlManager.java:90) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
	at baritone.behavior.PathingBehavior.onTick(PathingBehavior.java:104) ~[baritone-unoptimized-fabric-1.10.2.jar:?]
```

tolland:
So it was a while back when I looked at this, but the underlying issue is that baritone is returning true if the block is inside the extends of the schematic (assuming it is a rectangular cuboid) but the litematica schematic only defines the blocks for those subregions... and you can see from the image, that there are gaps between the minX and maxX etc

ZacSharp:
I wasn't saying that this doesn't happen, I was saying that you are fixing the wrong file. Just like you said yourself, `LitematicaSchematic.inSchematic` mistakenly returns `true` even outside subregions. `BuilderProcess` relying on a bounds checked position being in bounds is fine.

tolland:
Ah, I justed tested it... and yeah it doesn't fix the problem. So I will revisit.

tolland:
Any suggestions on how to get litematica mod running under the gradle debug build in intellij? dropping litematica.jar into the mods folder is not working

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -156,10 +156,25 @@ private static long getVolume(CompoundTag nbt, String subReg) {
...
+            Vec3i offsetSubregionMin = new Vec3i(getMinOfSubregion(nbt, subReg, "x"), getMinOfSubregion(nbt, subReg, "y"), getMinOfSubregion(nbt, subReg, "z"));
+
+            if (inSubregion(nbt, subReg, x + this.offsetMinCorner.getX(), y + offsetMinCorner.getY(), z + offsetMinCorner.getZ())) {
+                //System.out.println("found " + x + "," + y + "," + z + " in subregion: " + subReg + " state: " + currentState);
+                return true;
+            }
+        }
+        return false;
+    }
+
```
rycbar0, line 177: I agree that the default method needs a override however i have a few questions.
What is offsetSubregionMin doing? What happens if the schematic is rotated 90 degree? If im not mistaken the inSubregion() methods checks against the file in which the schematic isnt rotated.
Why not use
```
@Override
public boolean inSchematic(int x, int y, int z, BlockState currentState) {
    return getDirect(x, y, z) != null
}
```

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -39,7 +39,7 @@
  * @author rycbar
  * @since 22.09.2022
  */
-public final class LitematicaSchematic extends StaticSchematic {
+public class LitematicaSchematic extends StaticSchematic {
```
rycbar0, line 42: Why do you change all the method signatures? is it for your mock test?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -211,14 +250,14 @@ private void fillInSchematic() {
...
+    void writeSubregionIntoSchematic(CompoundTag nbt, String subReg, BlockState[] blockList, LitematicaBitArray bitArray) {
+        Vec3i offsetSubregionMin = new Vec3i(getMinOfSubregion(nbt, subReg, "x"), getMinOfSubregion(nbt, subReg, "y"), getMinOfSubregion(nbt, subReg, "z"));
         int index = 0;
         for (int y = 0; y < this.y; y++) {
             for (int z = 0; z < this.z; z++) {
                 for (int x = 0; x < this.x; x++) {
-                    if (inSubregion(nbt, subReg, x, y, z)) {
-                        this.states[x - (offsetMinCorner.getX() - offsetSubregion.getX())][z - (offsetMinCorner.getZ() - offsetSubregion.getZ())][y - (offsetMinCorner.getY() - offsetSubregion.getY())] = blockList[bitArray.getAt(index)];
+                    if (inSubregion(nbt, subReg, x + this.offsetMinCorner.getX(), y + offsetMinCorner.getY(), z + offsetMinCorner.getZ())) {
+                        this.states[x][z][y] = blockList[bitArray.getAt(index)];
```
rycbar0, line 260: I dont remember why exactly i did it this way but im sure there was a reason for it.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -211,14 +250,14 @@ private void fillInSchematic() {
...
-        Vec3i offsetSubregion = new Vec3i(getMinOfSubregion(nbt, subReg, "x"), getMinOfSubregion(nbt, subReg, "y"), getMinOfSubregion(nbt, subReg, "z"));
+    void writeSubregionIntoSchematic(CompoundTag nbt, String subReg, BlockState[] blockList, LitematicaBitArray bitArray) {
+        Vec3i offsetSubregionMin = new Vec3i(getMinOfSubregion(nbt, subReg, "x"), getMinOfSubregion(nbt, subReg, "y"), getMinOfSubregion(nbt, subReg, "z"));
         int index = 0;
         for (int y = 0; y < this.y; y++) {
             for (int z = 0; z < this.z; z++) {
                 for (int x = 0; x < this.x; x++) {
-                    if (inSubregion(nbt, subReg, x, y, z)) {
-                        this.states[x - (offsetMinCorner.getX() - offsetSubregion.getX())][z - (offsetMinCorner.getZ() - offsetSubregion.getZ())][y - (offsetMinCorner.getY() - offsetSubregion.getY())] = blockList[bitArray.getAt(index)];
+                    if (inSubregion(nbt, subReg, x + this.offsetMinCorner.getX(), y + offsetMinCorner.getY(), z + offsetMinCorner.getZ())) {
```
rycbar0, line 259: I assume this counteracts the changes in the next line in some way but i dont know how

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -169,10 +184,32 @@ private static long[] getBlockStates(CompoundTag nbt, String subReg) {
...
+        int yMax = yMin + Math.abs(ySize);
+        int zMax = zMin + Math.abs(zSize);
+
+        boolean withinX = xSize != 0 && x >= xMin && x < xMax;
+        boolean withinY = ySize != 0 && y >= yMin && y < yMax;
+        boolean withinZ = zSize != 0 && z >= zMin && z < zMax;
+
+        boolean found = withinX && withinY && withinZ;
+
+        return found;
```
rycbar0, line 212: a bit lengthy but not wrong as far as i can tell.
the xSize != 0 check is redundant (if size is 0, xMax = xMin  and x cant be >= and < to the same number)

tolland:
Thanks for the comments! I will address them.

I opened the PR because I thought it was a simple fix, but it seems quite involved. (e.g. it seems to break rotated and mirrored multiple subregions...)

I think I will close the PR until I am more confident about how this might interact with the other stuff in LitematicaHelper

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
tolland: yeah. to mock the class to avoid having to create the BuiltInRegistries stuff

PR #4379, opened 2024-06-05
[WIP] attempt 2 at fixing multiple subregions in litematica schematic printer

tolland:
This PR makes the following changes

1. adds a stracktrace if baritone fails to load the litematica schematic file
2. overrides the inSchematic method
3. sizes the BlockState array to allow x and z BlockState dimension to be swapped in case of rotation

### DIFF
--- a/src/main/java/baritone/process/BuilderProcess.java
+++ b/src/main/java/baritone/process/BuilderProcess.java
@@ -240,6 +240,7 @@ public final class BuilderProcess extends BaritoneProcessHelper implements IBuil
                     build(name, schematic2, correctedOrigin);
                 } catch (Exception e) {
                     logDirect("Schematic File could not be loaded.");
+                    e.printStackTrace();
                 }
             } else {
                 logDirect("No schematic currently loaded");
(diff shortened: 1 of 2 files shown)


ZacSharp:
Printing a stacktrace is nice and checking for null instead of searching for an enclosing subregion should do the trick (and is faster), but forcing a square footprint hurts.
Imagine loading a wall schematic, so maybe some 5 blocks wide, 10 blocks high and easily 50 blocks long. The schematic has a volume of 2.5k blocks but your change makes Baritone allocate space for 25k blocks. That's 90% wasted memory.
This example might be extreme, but looking at some schematics I have on disk there are actually two with 90% wasted space and 30% is common.

ZacSharp:
I should note that a lot of my schematics are too small to care about. Nobody cares about 10 more pointers.
Looking at a bulk storage system with size 34x56x128 that's 2.6Mb unused, and that's about the biggest non-square thing I could find.

ZacSharp:
Just noticed this is on the wrong branch. Please change the base branch to 1.19.4

PR #4389, opened 2024-06-14
bump versions in 1.20.4

tolland:
This PR

- bumps fabric loader version to 0.15.9 (to support running alongside malilib)
- update grade wrapper to 8.7 (from [unimined example](https://github.com/unimined/unimined/blob/main/gradle/wrapper/gradle-wrapper.propertiesl) )
- unimined to 1.2.3
- makes asm and sponge packages compileOnly (similar [to example ](https://github.com/cabaletta/baritone/compare/1.20.4...1.20.5#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R81-R82)in 1.20.5 branch)


ZacSharp:
If parts of this are applicable to 1.19.4 you should pr them against that branch instead.

PR #4452, opened 2024-08-01
Add settings to rotate/mirror schematics

### LINKED ISSUE #3711: Ability to rotate schematics / copied selections
## Describe your suggestion
This setting will allow you to more finely control the construction of structures, to customize which way they will look when building them.

## Settings
This should be a number field/switch with 4 positions (0, 90, 180, 270) that takes the value of how many degrees you want to turn the construction clockwise.

## Context
When building a downloaded scheme, the construction often doesn't fit in with things already built next to it, because it's not looking in the right direction for me. It would also be possible to do some sort of residential neighborhoods by putting buildings across from each other. And a lot of other cases as well.

## Final checklist

ZacSharp:
Closes #3711

### DIFF
--- a/src/api/java/baritone/api/schematic/MirroredSchematic.java
+++ b/src/api/java/baritone/api/schematic/MirroredSchematic.java
@@ -1,0 +1,114 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.api.schematic;
+
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.Mirror;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MirroredSchematic implements ISchematic {
+
+    private final ISchematic schematic;
+    private final Mirror mirror;
+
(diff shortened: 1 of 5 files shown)


src/api/java/baritone/api/schematic/MirroredSchematic.java
```diff
@@ -0,0 +1,114 @@
...
+                mirrorX(x, widthX(), mirror),
+                y,
+                mirrorZ(z, lengthZ(), mirror),
+                mirror(currentState, mirror)
+        );
+    }
+
+    @Override
+    public BlockState desiredState(int x, int y, int z, BlockState current, List<BlockState> approxPlaceable) {
+        return mirror(schematic.desiredState(
```
leijurv, line 48: It looks like this flips the schematic in memory each time you call desiredState, I feel as though this could get really laggy on large schematics. Can we instead mirror it once in the constructor?

src/api/java/baritone/api/schematic/RotatedSchematic.java
```diff
@@ -0,0 +1,136 @@
...
+                rotateX(x, z, widthX(), lengthZ(), inverseRotation),
+                y,
+                rotateZ(x, z, widthX(), lengthZ(), inverseRotation),
+                rotate(currentState, inverseRotation)
+        );
+    }
+
+    @Override
+    public BlockState desiredState(int x, int y, int z, BlockState current, List<BlockState> approxPlaceable) {
+        return rotate(schematic.desiredState(
```
leijurv, line 51: Same here

src/api/java/baritone/api/schematic/MirroredSchematic.java
ZacSharp: No, because the underlying schematic might have changed since the last call. We can only do this for static schematics, which most of our schematics are not (at least not reliably), and even then we might end up wasting memory by copying a sparse schematic into a dense array.
Also I think baking things in the wrapper schematic itself is actually a bad idea in most cases, because it forces the result to be static while we actually want it to be static iff it wraps a static schematic. Baking should be done separately from constructing the transformed schematic (i.e. `schematic = new BakedSchematic(schematic)`).

IIRC mirroring block states is fairly fast (cache lookup) so the potentially laggy part here is transforming up to 36 states from `approxPlaceable`. Given that we know how `BuilderProcess` uses `desiredState` we could commit a caching crime here (reuse last transformed if last untransformed == new untransformed, assuming the input wasn't mutated).

PR #4466, opened 2024-08-10
mark 1.21.1 as compatible

rfresh2:
1.21.1 is compatible, just needs this fmj update

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -25,6 +25,6 @@
 
   "depends": {
     "fabricloader": ">=0.14.22",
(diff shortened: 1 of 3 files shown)


leijurv:
Hm, do we think there should be a 1.21.1 branch, or is the update so small that it can just be marked compatible like this?

rfresh2:
> Hm, do we think there should be a 1.21.1 branch, or is the update so small that it can just be marked compatible like this?

No reason to make another branch

the diff in mc's source is like less than 50 lines and irrelevant to 99.999% of mods as it only affects servers

neoforge even allows all 1.21 mods to load on 1.21.1 already

PR #4472, opened 2024-08-15
Update fabric.mod.json (1.21.1)

mcgamer48ft:
maybe 1.21.1 support?


enorsu:
we all need this ngl

leijurv:
Looks like #4466 is better

PR #4485, opened 2024-08-28
1.20.1 Use pose system for calculating the height of crouching instead of hardcoded values

### LINKED ISSUE #4484: Using mods which change your eye height causes block placement to fail
When using a player scaling mod, breaking blocks will work completely fine, but trying to have baritone place blocks will end up with it constantly looking at the wrong angles to do so

May possibly be related to this hardcoded function here? I'm new to the codebase so excuse me if it isn't I'm just poking around to see https://github.com/cabaletta/baritone/blob/ef72c568330fb874ea6e912c7d6323cd79825057/src/api/java/baritone/api/utils/IPlayerContext.java#L102-L104

## Some information
Operating system: Windows
Java version: Fabric
Minecraft version: 1.20.1
Baritone version: 1.10.1 fabric
Other mods (if used): Pehkui

## How to reproduce
Use a mod like Pehkui to change your player height to around a block or less and baritone will consistently fail to aim at the correct location and get stuck trying to place the block

## Final checklist

Happyrobot33:
instead of using hardcoded crouching values, just use the built in method for geting a entitys eye height with a specific pose. Should close #4484 from personal testing

closes #4484

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -51,6 +51,8 @@ import java.util.function.Consumer;
  */
 public final class Settings {
 
+    public final Setting<Double> expectedEyeHeight = new Setting<>(1.27D);
+
     /**
      * Allow Baritone to break blocks
      */
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -51,8 +51,6 @@ import java.util.function.Consumer;
  */
 public final class Settings {
 
-    public final Setting<Double> expectedEyeHeight = new Setting<>(1.27D);
-
     /**
      * Allow Baritone to break blocks
      */
--- a/src/api/java/baritone/api/utils/IPlayerContext.java
+++ b/src/api/java/baritone/api/utils/IPlayerContext.java
@@ -99,16 +99,6 @@ public interface IPlayerContext {
         return new Rotation(player().getYRot(), player().getXRot());
     }
 
-    static double eyeHeight(boolean ifSneaking) {
-        //final double baseHeight = 1.62;
-        final double baseHeight = 0.65;
-        //sneaking is 24.2215% lower than standing
-        return ifSneaking ? baseHeight * 0.757785 : baseHeight;
-        //return ifSneaking ? Baritone.settings().expectedEyeHeight.value * 0.757785 : Baritone.settings().expectedEyeHeight.value;
-        //return player().getEyeHeight();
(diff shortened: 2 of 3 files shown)


Happyrobot33:
closes #4484

ZacSharp:
Removing a public method from the API package. I'd say an `@Deprecated` is more appropriate.

Otherwise a simple and sane looking change.

ZacSharp (approved):

wagyourtail:
IPlayerContext does have access to the player entity, so you could just put the pose thing in the old function

Happyrobot33:
> IPlayerContext does have access to the player entity, so you could just put the pose thing in the old function

the problem is its static so I'm not sure how to even reference the IPlayerContext directly here since it doesn't have a instance reference. and if it was changed to non static it would still be a breaking change

ZacSharp:
There is the possibility of `BaritoneAPI.getProvider().getAllBaritones().stream().map(Baritone::getPlayerContext).filter(c -> c.player() == entity).findFirst().get().getEyeHeight(true)`, but that's clearly worse and (re?)introducing `BaritoneProvider::getBaritoneForPlayer` wouldn't help much.
Also why assume the entity is a player? I'd say either the method has to take a `LocalPlayer` argument or assuming the entity is a player is plain wrong. Not sure whether we should assume there is a Baritone instance for the passed player.

EDIT: I've been tricked by an interface default implementation. `IBaritoneProvider::getBaritoneForPlayer` does exist, but not in the file I looked at. So it would be `BaritoneAPI.getProvider().getBaritoneForPlayer((LocalPlayer) entity).getPlayerContext().getEyeHeight(true)`, with potential for CCE and NPE for no reason.

Happyrobot33:
> There is the possibility of `BaritoneAPI.getProvider().getAllBaritones().stream().map(Baritone::getPlayerContext).filter(c -> c.player() == player).findFirst().get().getEyeHeight(true)`, but that's clearly worse and (re?)introducing `BaritoneProvider::getBaritoneForPlayer` wouldn't help much. Also why assume the entity is a player? I'd say either the method has to take a `LocalPlayer` argument or assuming the entity is a player is plain wrong. Not sure whether we should assume there is a Baritone instance for the passed player.

yeah that long method call is just painful to read as is and feels very breakable lol. I think deprecating it is a better option here, and anyone using it should just refer to the entity itself in the same way the codebase does

leijurv:
can this be done to the 1.19 branch?

ZacSharp:
Rebased onto 1.19.4: https://github.com/cabaletta/baritone/pull/4713

PR #4501, opened 2024-09-16
Make every Vec3i use BetterBlockPos hash

babbaj:
BetterBlockPos returning a different hashCode than the superclasses is a massively stupid correctness problem and is currently causing my game to crash.
In theory this might also improve performance a bit if it allows the jvm to devirtualize calls to hashCode.

### DIFF
--- a/src/api/java/baritone/api/utils/BetterBlockPos.java
+++ b/src/api/java/baritone/api/utils/BetterBlockPos.java
@@ -78,11 +78,6 @@ public final class BetterBlockPos extends BlockPos {
         return new BetterBlockPos(pos);
     }
 
-    @Override
-    public int hashCode() {
-        return (int) longHash(x, y, z);
-    }
-
     public static long longHash(BetterBlockPos pos) {
         return longHash(pos.x, pos.y, pos.z);
     }
--- a/src/launch/java/baritone/launch/mixins/MixinVec3i.java
+++ b/src/launch/java/baritone/launch/mixins/MixinVec3i.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of Baritone.
+ *
+ * Baritone is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Baritone is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


0-x-2-2 (approved):

leijurv:
This is a crazy big change, it could have tons of unintended side effects 😭

ZacSharp:
Better stop passing `BetterBlockPos` as if it were a `BlockPos`, ideally removing the subclassing (but that causes hundreds of compile errors).

Alternatively if no `BetterBlockPos` is ever equal to any non-`BetterBlockPos` (as is normal for most classes) this won't be a problem at all. Just very annoying and error prone to work with. (And it would require a mixin into `BlockPos.equals`)

babbaj:
I think the only other solution is to just not override the hashcode at all and figure out a different way to use the better hash function when using hashmaps but I don't think this could cause any problems anyways

ZacSharp:
That different way of using the better hash would be a class equivalent to `BetterBlockPos` but without extending from `BlockPos`. That could be either `BetterBlockPos` itself (lots of breakage) or a reimplementation (unknown effort, probably easier).
In either case we'd probably want to migrate Baritone to use either our own position class or `BlockPos` exclusively (I vote for using our own) and convert explicitly were needed.

babbaj:
The different way could just be fastutil's OpenCustomHashMaps or a simple class that just extends BetterBlockPos and is explicitly only to be used as keys to a hashmap. Creating a completely separate class would be unnecessarily annoying and would inevitably create a ton of unnecessary conversions that would make performance worse.

ZacSharp:
Please not a class only meant to be used for hash maps. That's doomed to end where we are already.

wagyourtail:
I don't think overriding the hashCode should cause any issues in practice, nothing should actually rely on the hashCode being the same between game launches...

anyway, the other way would be to extend HashMap/HashSet to overwrite what it querys for hashCode like babbaj said

leijurv:
the map already grabs the hashcode manually: https://github.com/cabaletta/baritone/blob/1.19.4/src/main/java/baritone/pathing/calc/AbstractNodeCostSearch.java#L169

ZacSharp:
Is there any other reason to overwrite the hash code? (Other than general Minecraft performance, which I'd leave to optimization mods and mojang)

ZacSharp:
This issue has been known for long enough, so can we either gets this in or completely remove the override?
If there are mixin/other conflicts we can still remove the mixin (can mixin priority handle the former per run?).

Murat65536 (approved):

PR #4505, opened 2024-09-19
Interpolated Looking implementation

TheThouZands:
Replaced old, in many cases not working "smoothlook" with "interpolatedLook", which forces global player rotation in both axis to follow a "path" from source looking position(rotation) to target, removing jittering when, for instance, mining blocks; it WILL hinder baritone's default mining speed, but the main purpose is to make actions seem slightly more humanly possible, time of the interpolation is controlled by interpolatedLookLength, in ticks.

### DIFF
--- a/src/api/java/baritone/api/utils/VecUtils.java
+++ b/src/api/java/baritone/api/utils/VecUtils.java
@@ -19,6 +19,7 @@ package baritone.api.utils;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.util.Mth;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.BaseFireBlock;
@@ -119,5 +120,9 @@ public final class VecUtils {
      */
     public static double entityFlatDistanceToCenter(Entity entity, BlockPos pos) {
         return distanceToCenter(pos, entity.position().x, pos.getY() + 0.5, entity.position().z);
+    }
+
+    public static Vec3 vDivide (Vec3 A, Vec3 B) {
+        return new Vec3(A.x / B.x, A.y / B.y, A.z / B.z);
     }
 }
--- a/src/api/java/baritone/api/utils/VecUtils.java
+++ b/src/api/java/baritone/api/utils/VecUtils.java
@@ -19,7 +19,6 @@ package baritone.api.utils;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
-import net.minecraft.util.Mth;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.BaseFireBlock;
@@ -124,9 +123,5 @@ public final class VecUtils {
      */
     public static double entityFlatDistanceToCenter(Entity entity, BlockPos pos) {
         return distanceToCenter(pos, entity.position().x, pos.getY() + 0.5, entity.position().z);
-    }
-
-    public static Vec3 vDivide (Vec3 A, Vec3 B) {
-        return new Vec3(A.x / B.x, A.y / B.y, A.z / B.z);
     }
 }
--- a/src/api/java/baritone/api/utils/VecUtils.java
+++ b/src/api/java/baritone/api/utils/VecUtils.java
@@ -19,6 +19,7 @@ package baritone.api.utils;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.util.Mth;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.BaseFireBlock;
@@ -123,5 +124,9 @@ public final class VecUtils {
      */
     public static double entityFlatDistanceToCenter(Entity entity, BlockPos pos) {
         return distanceToCenter(pos, entity.position().x, pos.getY() + 0.5, entity.position().z);
+    }
+
+    public static Vec3 vDivide (Vec3 A, Vec3 B) {
+        return new Vec3(A.x / B.x, A.y / B.y, A.z / B.z);
     }
 }
--- a/src/main/java/baritone/behavior/LookBehavior.java
+++ b/src/main/java/baritone/behavior/LookBehavior.java
@@ -25,7 +25,11 @@ import baritone.api.behavior.look.ITickableAimProcessor;
 import baritone.api.event.events.*;
 import baritone.api.utils.IPlayerContext;
 import baritone.api.utils.Rotation;
+import baritone.api.utils.RotationUtils;
+import baritone.api.utils.VecUtils;
 import baritone.behavior.look.ForkableRandom;
+import baritone.utils.BaritoneMath;
+import net.minecraft.world.phys.Vec3;
 import net.minecraft.network.protocol.game.ServerboundMovePlayerPacket;
 
 import java.util.ArrayDeque;
@@ -53,8 +57,16 @@ public final class LookBehavior extends Behavior implements ILookBehavior {
 
     private final AimProcessor processor;
 
-    private final Deque<Float> smoothYawBuffer;
-    private final Deque<Float> smoothPitchBuffer;
+    /**
+     * the interpolation's current stage, to be used by {@link #onPlayerUpdate(PlayerUpdateEvent)} whenever {@link Settings#interpolatedLook} is true.
+     */
+    private static int stage = 0;
+    private static Vec3 Source;
+    private static Vec3 Destiny;
+    private static Vec3 Center;
+
+    //private final Deque<Float> smoothYawBuffer;
+    //private final Deque<Float> smoothPitchBuffer;
 
     public LookBehavior(Baritone baritone) {
         super(baritone);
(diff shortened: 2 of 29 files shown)


TheThouZands:
https://github.com/user-attachments/assets/26937d72-5cdc-42a6-aefb-c4f46492a7b9

"showcase" video

src/api/java/baritone/api/utils/VecUtils.java
```diff
@@ -120,4 +121,8 @@ public static double entityDistanceToCenter(Entity entity, BlockPos pos) {
     public static double entityFlatDistanceToCenter(Entity entity, BlockPos pos) {
         return distanceToCenter(pos, entity.position().x, pos.getY() + 0.5, entity.position().z);
     }
+
+    public static Vec3 vDivide (Vec3 A, Vec3 B) {
```
sqlerrorthing, line 125: use google codestyle guidelines plz

src/main/java/baritone/behavior/LookBehavior.java
```diff
@@ -87,49 +99,146 @@ public void onPlayerUpdate(PlayerUpdateEvent event) {
...
-                        if (ctx.player().isFallFlying()) {
-                            ctx.player().setXRot((float) this.smoothPitchBuffer.stream().mapToDouble(d -> d).average().orElse(this.prevRotation.getPitch()));
+                }
+                case POST: {
+                    // Same as before, I really do wish that both PRE and POST are both done before frame generation, but that would be just not to trigger epilepsy
+                    if (this.prevRotation != null) {
+                        if (this.target.mode == Target.Mode.SERVER) {
+                            ctx.player().setYRot(this.prevRotation.getYaw());
+                            ctx.player().setXRot(this.prevRotation.getPitch());
+                        }// else if (ctx.player().isFallFlying() && Baritone.settings().elytraSmoothLook.value) {
```
sqlerrorthing, line 206: what is it? :skull:. Commits the commented code is pretty bullshit

sqlerrorthing (commented):
use google guidelines rn & dont commit the commented code

Slorbin:
The interpolation is very cool, but given how abysmally slow the speed is you might want to 'boost' remainWithExistingLookDirection so it REALLY prefers to mine blocks in the existing look direction, to minimize turning. It looks more human anyway

Or add some sort of penalty setting for it at least?

Slorbin:
Actually, having played around with this a bit, it's real nice. MUCH better and FASTER somehow than smoothlook, just have to nerf the ticks. It's nice!

LopezJLD:
> Actually, having played around with this a bit, it's real nice. MUCH better and FASTER somehow than smoothlook, just have to nerf the ticks. It's nice!

hey can you help me to compile with this changes? idk why it doesnt work when i do it :( im new in all this things

weedforspeeed:
how can i use your modification mod. btw it looks really good and smooth

weedforspeeed:
hey i really like your mod how can i use it? is there a .jar file which i can use?

weedforspeeed:
> Actually, having played around with this a bit, it's real nice. MUCH better and FASTER somehow than smoothlook, just have to nerf the ticks. It's nice!

how did you use it? im new to github an all this so i dont have a clue. is there like a .jar file or anything i would really like to use the showcased mod

Slorbin:
Kind of a miracle I got it working. You need to download their repo as a
folder or something and compile it into a jar yourself. It was rough but
there should be a guide or two online

weedforspeeed:
alright thank you bro imma try it out. but wouldnt it be possible that you could send me the jar file?

Slorbin:
I lost the file sadly and I dunno how to remake it, you’re on your own

weedforspeeed:
alright still thank you for the advice

weedforspeeed:
im not able to do it :(

weedforspeeed:
alright i made it to make the .jar file but which one to use that it works like in the video fabric or forge and the "unoptimized" one or the other one? because its still laggy with interpolated look

weedforspeeed:
and which version to use?

TheThouZands:
> alright i made it to make the .jar file but which one to use that it works like in the video fabric or forge and the "unoptimized" one or the other one? because its still laggy with interpolated look

Hey there, it's been a while, i have been kinda occupied with work (not coding related sadly) hence i haven't been able to correct the pr for coding guidelines.

As for this, you should be able to use the normal or obfuscated version, the unoptimized version is only for debugging, but the function should be fully implemented on all jar types

Also, i made this pr for, at it's moment, the main branch, 1.19, should work best there, but could work for other versions as well, but would need some precisions

Also yes, it may look laggy because it is based on tick updates rather than frame updates; most of baritone works on that same 20 TPS refresh rate, may have not been noticeable before as movement took less than one frame, implementing this in a way that updates frame per frame is possible, however would require mixin work, which i'm not precisely familiar with, however, upon doing playtesting, the main arc calculation function, which is the heaviest part of the pr, took near 10 ms, which again means it's entirely plausible to update it on a frame per frame basis

weedforspeeed:
hey glad youre back so i got your version and im playing 1.19.4 fabric and i just the standalone fabric .jar file in the mods folder but eventho i got a good pc its not nearly as smooth as in your video. what could possibly be wrong?

weedforspeeed:
https://github.com/user-attachments/assets/add6d87d-a0c3-41fe-bed4-60d65dc7d84e

this is how it looks on my end

TheThouZands:
I see, i believe you are referring to the pauses between each mined block

This is indeed something that tends to happen often, and basically it's because the interpolation is run always.

My function is directly called by baritone, since it was designed to have intant movement, it checks every single tick for moving the camera to point at the block
In this case, my function does manage to override part of that, successfully placing itself on top of the location requests and smoothing the movement within one point and the other

However, when that point is reached, the override no longer applies, and baritone attempts to call the looking function again, so if the block hasn't yet been mined, the arc will be calculated for the same point, running the cycle for the same point (hence why it looks static)

In that case, it does hinder speed a bit due to those pauses, things that could reduce it would be to lower the interpolation ticks

A more permanent solution is to know when the block has been successfully mined by the player, but once again, it involves non-implemented mixins, and as i'm still in a (slightly horrible) job, i haven't had time to explore mixins as well as implementing the solution for this

TheThouZands:
So yes, it's working without lag issues (not performance problem) but the same interpolation gets in the way if mining a block takes too little time or too much time

TheThouZands:
> So yes, it's working without lag issues (not performance problem) but the same interpolation gets in the way if mining a block takes too little time or too much time

If it was a performance issue, all minecraft would stop, otherwise it would generate frames at a slowe rate or directly freeze, just need to figure out a way for baritone to see the block itself instead of a position, and know when it hasn't/has been mined

weedforspeeed:
i mean like the camera movement is more "laggy" than in your video. how can i make it buttersmooth like you?

weedforspeeed:
and i somehow keep looking at the bottom when im walking

TheThouZands:
> i mean like the camera movement is more "laggy" than in your video. how can i make it buttersmooth like you?

Oh yeah, i see what you mean, but in the showcase it's also pretty jittery, because the movement is only called 20 times per second instead of 60, as it depends on the world's tick, just as most of Baritone, thing is that it's less noticeable because the video is heavily compressed and running at 30 fps, as at the moment I didn't have enough space

> and i somehow keep looking at the bottom when im walking

and not sure what you mean, do you perchance have a video showcasing this as well?

weedforspeeed:
unfortunatly, i cant put any videos in here that are anylonger than 10 sec. "and i somehow keep looking at the bottom when im walking" i think this is just something that happens sometimes with baritone. and what i also noticed #interpolatedlook sets itsself during the programm to false. but dont get me wrong youre programm is great and i think the only one who makes baritone a little smoother

weedforspeeed:
when you make for example #goto 1 1 1 #interpolatedlook goes automaticly to false after that

weedforspeeed:
but i dont know how i can fix that jittery look it doesnt look jittery in your video

Murat65536:
@TheThouZands are you willing to continue development?

Camille1358:
why it's not on the main branch ?

TheThouZands:
> @TheThouZands are you willing to continue development?

I forgot about this, i do feel like taking on this with a nicer approach as soon as i have some time if you are still interested

> why it's not on the main branch ?

when i made this i was not a contributor, still i am not, and it hasn't been merged given i didn't follow guidelines, TODO.

Camille1358:
@TheThouZands So I'm glad I sent another message if it helped you remember that, because personally, I'm really, really interested in this! ^^ Feel free to include the pull request I mentioned (5043) it would make things even better. Thanks for everything ^^ (Take as much time as you need.)

src/api/java/baritone/api/utils/VecUtils.java
TheThouZands: solved in [5a48840](https://github.com/cabaletta/baritone/pull/4505/commits/5a48840c274a7f9d26bdd74c79916b94affa9d2e)

src/main/java/baritone/behavior/LookBehavior.java
TheThouZands: no longer commiting commented code

TheThouZands:
Update:
- Adds reusable rotation arc logic in `RotationUtils` (improved)
- Interpolates look movement by angular velocity instead of fixed subdivision count
- Smooths the rendered camera using Minecraft partial-tick sampling
- Keeps interpolated server-side rotations compatible with freelook
- Restarts interpolation cleanly when the player origin or target visibility changes
- Holds aim steady while actively mining the intended target block
- Retries targeting if the held ray drifts off the target
- Avoids pole-crossing arcs by preferring latitudinal arcs near view poles
- Adds reachability hints for block breaking and placing
- Pauses block interaction behavior while the game is paused or the player is dead

TheThouZands:
for windows after downloading this branch, as per instructions in root
`gradlew.bat build`

look in the dist folder for `baritone-standalone-"YOURMODLOADER"` or api and place that in mods as normal

TheThouZands:
[showcaseagain](https://www.youtube.com/watch?v=c9CM4Vq0PS8)

MigVic:
I experimented with the mod a bit, even made a 1.21.11 version if anyone needs, but I have a bit of problem. Whenever you build something it can sometimes loop looking between two empty spaces (where its trying to place blocks but doesnt). Im guessing it has to do with the fact that interpolation messes with the goal timing, which in this case is placing block. Maybe a delay or a look-check should be added so it properly works? This issue doesnt happen with interpolation set to false, I tested.

TheThouZands:
> a bit of problem. Whenever you build something it can sometimes loop looking between two empty spaces (where its trying to place blocks but doesnt).

thank you, something like that, block placement didn;t have a goal timer, it just attempted to get a new goal per tick

i also checked and fixed farming, and some movement odds

MigVic:
Im sorry if you fixed it, potentially with [Target movement block interactions] but when baritone is moving from location to location it does not follow the path it set exactly, instead it sometimes follow where the camera is pointing, smoothing out corners and sometimes leading to player falling down if pathing on a 1 block wide area with turns

TheThouZands:
> Im sorry if you fixed it, potentially with [Target movement block interactions] but when baritone is moving from location to location it does not follow the path it set exactly, instead it sometimes follow where the camera is pointing, smoothing out corners and sometimes leading to player falling down if pathing on a 1 block wide area with turns

actually ty,, any issues are great to know

though it's funny, i haven't touched traversal code, so i don't know why it's also smooth, will be taking a look into it

Issue #209, opened 2018-10-06
Fix path starting position while sneaking


leijurv:
If you sneak off the edge of a block and start a path, the starting position will be not actually supported. If it was off the edge of a cliff, it's possible that the first movement will cause the player to just fall off.

Right now there is a heuristic, here:
https://github.com/cabaletta/baritone/blame/master/src/main/java/baritone/behavior/PathingBehavior.java#L259
https://github.com/cabaletta/baritone/commit/ee8e44d3143e83720ee546e4438a17710c838e61

It's okay, but only if you're sneaking over a 1-high ledge.
It also requires this sketchy workaround:
https://github.com/cabaletta/baritone/blob/master/src/main/java/baritone/pathing/path/PathExecutor.java#L97

Which isn't great.

Instead we should figure out which block is actually supporting us. I don't really know how to do this.
First off, don't do any of this if the block below playerFeet is canWalkOn. 
In the scenario where our current position isn't a valid place to start a path, we need to check the 8 adjacent blocks to see if they're supporting us.

ZeroMemes:
Once the ``pathStart`` method is fixed, that workaround could be removed, correct?

leijurv:
yes

leijurv:
lol this is the same as #70

leijurv:
I've come up with a "good enough" solution, in my opinion

- In the scenario where we are on the ground, yet the block directly below our feet isn't canWalkOn (i.e. we're sneaking off an edge), figure out the block we're sneaking off of, and start a path from there
- In the scenario where we are not on the ground, and the block directly below our feet isn't canWalkOn, but the block directly below that **is** canWalkOn, move our starting block down by one
- In all other scenarios, playerFeet is the path start

Here is a video demonstrating that algorithm in action, the block highlighted in white is where a path would start:
https://youtu.be/5jlVmMXFCxU

PR #4519, opened 2024-10-14
Fix not moving from edge of block

babbaj:
To try to keep you safe, if you are standing on the edge of a block baritone will choose to start the path on that block rather than the feet because you are standing on air. If the goal happens to also be on that block, the start will already be in the goal and A* will return no movements so baritone will not move. If that is the case then it should be safe to just path from the feet anyways.

tl;dr this fixes baritone not moving when you are on the edge of a block and the goal is in your face

more information in #209

### DIFF
--- a/src/main/java/baritone/behavior/PathingBehavior.java
+++ b/src/main/java/baritone/behavior/PathingBehavior.java
@@ -265,7 +265,7 @@ public final class PathingBehavior extends Behavior implements IPathingBehavior,
         if (goal == null) {
             return false;
         }
-        if (goal.isInGoal(ctx.playerFeet()) || goal.isInGoal(expectedSegmentStart)) {
+        if (goal.isInGoal(ctx.playerFeet())) {
             return false;
         }
         synchronized (pathPlanLock) {
@@ -553,7 +553,7 @@ public final class PathingBehavior extends Behavior implements IPathingBehavior,
         });
     }
 
-    private static AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal, IPath previous, CalculationContext context) {
+    private AbstractNodeCostSearch createPathfinder(BlockPos start, Goal goal, IPath previous, CalculationContext context) {
         Goal transformed = goal;
         if (Baritone.settings().simplifyUnloadedYCoord.value && goal instanceof IGoalRenderPos) {
             BlockPos pos = ((IGoalRenderPos) goal).getGoalPos();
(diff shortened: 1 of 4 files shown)


ZacSharp:
> If that is the case then it should be safe to just path from the feet anyways.

No, it's not. Walking onto the supporting block should be safe, but there is no guarantee that a* will decide to do so.
E.g. if all four directions lead directly to a goal, evaluation order decides which is picked. In this video that's consistently -z even if the only safe option is +z.

https://github.com/user-attachments/assets/19e548e2-a858-40a9-9176-ee81d818b03a

leijurv:
Thanks for checking that zac, this is the exact case that I was avoiding back in 2018. Particularly, when pathing around 2b2t spawn, it would sometimes get stuck and recalculate, meaning this could happen in the middle of a path not just at the beginning. Sometimes it would be just off the edge of a block, and so if there was no foot support underneath the first movement, it would sometimes decide to do a walk (or a parkour) that would instantly fall like that.

babbaj:
Just pushed a completely different way to fix this buy prepending the feet position the the path if the start position is different from the feet and only 1 block away

leijurv:
This is a much better approach I think.

If I'm understanding right, if I'm sneaking off the edge of a block, this will insert an initial movement from playerFeet to pathStart?

This solves your issue because if I'm sneaking off the edge of the goal, this will create a 1 movement long path that basically just centers the player on the block? Does that actually work though? Like will AStarPathFinder actually return an empty path, which you prepend that one movement to?

In the actual technique here, I think there might be a more elegant way:

https://github.com/cabaletta/baritone/blob/1e2ae34dbe452a1eebc12019ae632b2279210cad/src/main/java/baritone/pathing/calc/AStarPathFinder.java#L54-L58

Could there just be like a "if we are doing the fake extra movement, make a node called realStart, and do startNode.prev=realStart? That way AStar won't freak out on a path of length 0?

babbaj:
> If I'm understanding right, if I'm sneaking off the edge of a block, this will insert an initial movement from playerFeet to pathStart?

yes

I did some debugging to understand what happens and A* does return a path of 0 length. This only changes what happens after A* returns. That change would be a bit more elegant but wouldn't make any difference because the PathNodes aren't treated like a linked list after that (`assembleMovements` uses list indexing).

leijurv:
Okay! So, for example, in zacsharp's example gif, this version would not fall into lava, it would instead path from the supporting block around the lava to the goal, then insert a movement from the above lava position to the supporting block? So it would make like a `?` shape?

If you want to do it this way, IMO that's fine. Prepending a blockpos is fine. I just think "realStart" is a bit of a vague name, but fine. Can you add a block comment somewhere explaining the idea and linking to this PR?

babbaj:
> then insert a movement from the above lava position to the supporting block? So it would make like a ? shape?

yes

>  I just think "realStart" is a bit of a vague name

It could be called playerAt or something but that's not the case when baritone is pathfinding ahead

ZacSharp:
Found an edge case:  #4528

PR #4536, opened 2024-10-26
1.21.3

rfresh2:


rfresh2:
should be rebased on https://github.com/cabaletta/baritone/pull/4534 after that's merged

src/main/java/baritone/utils/ToolSet.java
```diff
@@ -87,12 +88,9 @@ public double getStrVsBlock(BlockState state) {
...
     private int getMaterialCost(ItemStack itemStack) {
-        if (itemStack.getItem() instanceof TieredItem) {
-            TieredItem tool = (TieredItem) itemStack.getItem();
-            return (int) tool.getTier().getAttackDamageBonus();
-        } else {
-            return -1;
-        }
+        Tool toolComponent = itemStack.get(DataComponents.TOOL);
+        if (toolComponent == null) return -1;
+        return toolComponent.damagePerBlock(); // todo: i have no idea what "material cost" means anymore
```
ZacSharp: It's supposed to be a rough estimate how valuable the tool is. Just so if you happen to have two equally fast tools, Baritone will pick the cheaper one.

src/api/java/baritone/api/Settings.java
```diff
@@ -653,13 +653,6 @@ public final class Settings {
      */
     public final Setting<Boolean> logAsToast = new Setting<>(false);
 
-    /**
-     * The time of how long the message in the pop-up will display
-     * <p>
-     * If below 1000L (1sec), it's better to disable this
-     */
-    public final Setting<Long> toastTimer = new Setting<>(5000L);
```
ZacSharp, line 664: Not needed/possible anymore?

rfresh2: probably possible but the timer is encoded as a final field in `SystemToastId`

so we'd have to revert back to implementing a custom `Toast` subclass or do some other hack

reusing system toasts is much simpler and i don't think its worth the maintenance burden doing the alternatives

src/main/java/baritone/utils/ToolSet.java
rfresh2: gotcha, that makes more sense.

i think there's an enchantment cost data component for something like the prev behavior then. i'll update and also add clarifying comment about what this "cost" means

src/api/java/baritone/api/Settings.java
ZacSharp: Ok, fine.

andrew121410:
#4534 has been merged

Edit: I'm blind

ZacSharp:
And rfresh already did the rebase.

leijurv:
Should this have a new base branch 1.21.3 or should I actually merge into 1.21.1?

rfresh2:
new base branch

Issue #3537, opened 2022-07-01
Farming on baritone (and multiple goals not working)


Flupzy:
There seems to be a fundemental problem with the farming command in baritone, the problem is that when you execute the farming command it just farms plantable crops in your selection and then stops with the message "farm failed", when it should just farm inside the specified area until not told to. To work around this, in issue #2841 user ZacSharp tells you to "use #sel replace <crop> air and #sel replace air <crop>" but this has 3 reasons to why it doesnt work. 1: it destroys the plant as the ungrown state, so even if you could run both commands it would just endlessly plant and destroy a seed/crop multiple times a second because it cant recognize when it is in its grown state. 2: it doesnt pickup the dropped items.

3: im not sure if this is just for me but i cant have the commands #sel replace <crop> air and #sel replace air <crop> running at the same time because i cant use the ; and & to do multiple commands it just says "Error at argument #(argument that you put the ; or & in): Expected ForBlockOptionalMeta

if this cant be fixed relatively soon (or its on purpose somehow?!?) it would be absolutely lovely if someone could name me some alternatives to baritone for farming 

:)

Flupzy:
i meant #sel replace air (crop) and #sel replace (crop) air

Issue #2315, opened 2020-12-31
Farm command settings


GDBlaster:
## Suggestion UwU
having a more customizable farm command
## sum lil parameters UwU
something like ".b farm wheat" to farm only wheat/sugarcanes etc...
## Context OwO
beacause its really annoying having the bot running away from the farm beacause it saw some pumpkin or stuff like that on the other side of a mountain
## edit
Added some UwU's
## Final checklist
- [x] I know how to properly use check boxes 
- [ ] I have not used any OwO's or UwU's in this issue.

m-sarabi:
And also the selection command could be used here, like "#sel farm" to only farm crops that are within the selection area.

wagyourtail:
duplicate #893

wagyourtail:
oops wrong one, was using this as context for a newer one

Tytova:
did they add this yet?

Issue #893, opened 2019-09-01
#farm Imporvements and ideas


Googlelt:
## Describe your suggestion

So let me get straight to the point, Auto-farm is a game-changer for economy based servers I can build a farm with an auto-sell script or mod with baritone and make tons of money although the Auto-farm feature currently has a lot of flaws.

I would like to suggest a couple of key improvements I feel might improve the reliability and performance of the auto farm feature.  

- **An option to prioritize drops over replanting or breaking grown crops**
- **layer order for farming**

## Settings

**farmDroppedItemsPriority**
While farming, Prioritize picking up drops over breaking and replanting crops.

true = Bot will prioritize picking up dropped items when farming
false = bot will prioritize fully grown crops.

**(The already existing layerOrder setting should apply to farming)**

## Context
**farmDroppedItemsPriority**
When leaving the bot running I noticed the bot will try to replant crops before picking up the items causing the bot to move too far away to see the dropped item and no longer picking up the dropped item. which means that it's not the most optimal it could be for collecting the crops and I lose out on a lot of crops I could be selling.

**LayerOrder | for farming**
This is more of an Efficiency thing but I've noticed with multiple layers farms the bot will get halfway through and adjust the path to the next floor up I'm assuming because the range to break blocks is 5 blocks which means that it breaks blocks 5 blocks away and the next set of crops in some cases will be 3-4 blocks above so it thinks the closest crop would be on the next floor rather than the same layer.

## Final checklist
- [owo] I have not used any OwO's or UwU's in this issue.

Baritone is going to have huge implications on Economy and faction servers I've already started testing empty chunk schematics to clear bedrock to the sky for bases. I gave 6 bots a set of unbreaking efficiency picks and told them to clear a 3x3 chunk schematic and turned layer order on from the top and let it do its thing. and went and watched a movie came back and it was nearly done no issues. this is huge because that would normally take a group of actual players a lot of time this means that the actual players can focus more on PvP or anything else they could be doing. I'm Loving Baritone so much I cant wait to see it evolve to the point were the bots can communicate with each other and not start mining the same blocks when they're all doing the same schematic or farming the same chunk
--

wellsilver:
Completely agree, but somehow this is dissapeared for some reason it soesnt show up when you look up issues.

wellsilver:
They actually can communicate with each other, though you haved to set it up yourself and theres no information available.
[...]

PR #4561, opened 2024-11-18
Some `#farm` improvements

### LINKED ISSUE #396: Operation Area
Maybe so we can define a specific working area for the bot like ".b area x1 y1 z1 x2 y2 z2" and then the bot only operates in this specific field.
This is very usefull on server that have plots so the bot wont run to another house while auto farm is enabled.

### LINKED ISSUE #775: Feature Request: Update the farm module to farm only specified crops
## Describe your suggestion
It would be incredibly useful if you could direct the farming mod on baritone to farm specific crops in the same manner that you are able direct the auto-mine mod to mine only specific ores. 

## Context
This feature would prevent the farm module from planting crops in the wrong area which can be a problem.

Avtera:
added options:
- farm continuously, configurable task threshold and check interval
- farm whitelist for crop and pickup
- farm prioritize pickup

changed:
- farm compensate pickup pos on soulsand
- farm pickup now respect `#farm <range>` 

partially resolves #396 #893 #3537
closes #775 #2315

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -982,25 +982,43 @@ public final class Settings {
     public final Setting<Integer> farmMaxScanSize = new Setting<>(256);
 
     /**
-     * Prioritize gathering farm related drops
+     * Prioritize pickup crop drops before another farm task
      */
-    public final Setting<Boolean> farmPrioritizeGather = new Setting<>(false);
+    public final Setting<Boolean> farmPrioritizePickup = new Setting<>(false);
 
     /**
      * true = farm farm farm, i don't want to stop!
-            * false = allow farm to stop or fail
+     * false = allow farm to stop or fail
      */
     public final Setting<Boolean> farmContinuously = new Setting<>(false);
 
     /**
-     * How much farming task is enough for the Continuous Farm to resume?
+     * How much farm task queue is enough for Baritone to resume the continuous farm?
+     * {@link #farmContinuously}
      */
     public final Setting<Integer> farmContinuouslyThreshold = new Setting<>(16);
 
     /**
-     * Time interval (in seconds) for the Continuous Farm to check if threshold is fulfilled?
+     * How long Baritone should wait (in seconds) to check if continuous farm threshold is fulfilled?
+     * {@link #farmContinuously}
      */
     public final Setting<Long> farmContinuouslyIntervalSecs = new Setting<>(TimeUnit.MINUTES.toSeconds(2));
+
+    /**
+     * Farm whitelist, only interact with crop that is on the {@link #farmWhitelist} list
+     */
+    public final Setting<Boolean> farmEnableWhitelist = new Setting<>(false);
+
+    /**
+     * Crop block that Baritone is allowed to farm and collect
+     * {@link #farmEnableWhitelist}
+     */
+
+    public final Setting<List<Block>> farmWhitelist = new Setting<>(new ArrayList<>(Arrays.asList(
+            Blocks.WHEAT,
+            Blocks.POTATOES,
+            Blocks.CARROTS
+    )));
 
     /**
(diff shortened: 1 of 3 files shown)


src/api/java/baritone/api/Settings.java
```diff
@@ -981,6 +981,45 @@ public final class Settings {
...
+    /**
+     * Crop block that Baritone is allowed to farm and collect
+     * {@link #farmEnableWhitelist}
+     */
+
+    public final Setting<List<Block>> farmWhitelist = new Setting<>(new ArrayList<>(Arrays.asList(
+            Blocks.WHEAT,
+            Blocks.POTATOES,
+            Blocks.CARROTS
+    )));
```
ZacSharp, line 1021: The whitelist should default to allowing everything so `farmEnableWhitelist` is not needed.

Also I'm not sure whether we really should use `Block` rather than a new enum type for farm targets.

src/api/java/baritone/api/Settings.java
```diff
@@ -1397,7 +1436,7 @@ public final class Settings {
     /**
      * Desktop notification on farm fail
      */
-    public final Setting<Boolean> notificationOnFarmFail = new Setting<>(true);
+    public final Setting<Boolean> notificationOnFarmProcess = new Setting<>(true);
```
ZacSharp, line 1439: Why did you rename this?

src/main/java/baritone/process/FarmProcess.java
```diff
@@ -121,6 +159,14 @@ public void farm(int range, BlockPos pos) {
...
     }
 
+    public Long getTime() {
+        return time;
+    }
+
+    public void setTime(Long time) {
+        this.time = time;
+    }
+
```
ZacSharp, line 169: Why are these public?

end-user:
I noticed the original `#farm` did not seem to prefer a tool (specifically with _fortune 3_) when harvesting.  Could that be added?

ZacSharp (commented):
Just noticed that I have pending comments on this. The review is probably incomplete, but I'll submit it now before I forget it again.

Having `FarmProcess` enable and disable itself in a kind of paused-but-not-quite state looks really weird. Is there a reason to not have it return a `DEFER` pathing command while it is waiting for crops?

angelonger:
why was this never added in the stable branch?
IMO insanely useful commit - being able to farm for indefinite amount of times is what #farm currently lacks

src/main/java/baritone/process/FarmProcess.java
Avtera: okay, should be private, i guess? im new to this

src/api/java/baritone/api/Settings.java
Avtera: to avoid confusion while testing, i forgot to change it back

Avtera:
> Just noticed that I have pending comments on this. The review is probably incomplete, but I'll submit it now before I forget it again.
> 
> Having `FarmProcess` enable and disable itself in a kind of paused-but-not-quite state looks really weird. Is there a reason to not have it return a `DEFER` pathing command while it is waiting for crops?

-to conserve Power Usage for 24/7 auto farming (i don't know if there is a better way)
-for a small-scale farm or a farm of plant that grows slowly, like pumpkin

src/api/java/baritone/api/Settings.java
Avtera: > The whitelist should default to allowing everything so `farmEnableWhitelist `is not needed.

you're right.

> Also I'm not sure whether we really should use `Block` rather than a new enum type for farm targets.

i don't know, but yeah creating an enum would be consistent with the `FarmProcess.class`

Golaco:
Seems such a nice addition to #farm, b ut out of curiosity, is it supposed to be working as is? I tried it but when running #farm  i get: "Farming" and absolutly nothing else happens.

informationtoyou:
This should be merged, it's pretty decent.

PR #4553, opened 2024-11-09
Fix registry loading when custom registries present

rfresh2:
Resolves mod compat issues including https://github.com/cabaletta/baritone/issues/4526

Should be forward merged up through 1.21 as same issue is present there

PR #4589, opened 2024-12-14
Merge 1.20.6 into 1.21.1

ZacSharp:
Includes
* Fix `#litematica` in obfuscated environments (#4560)
* Don't load non-vanilla registries (#4553)


ZacSharp:
BTW is our version support policy still "latest minor of supported majors"?
If so I'll create a new pr directly to 1.21.3

rfresh2:
in my opinion bug fixes should be merged to all affected versions.

but for new features you can reduce the set to something like:

* 1.20.1 -> active LTS of 1.20 for most mods
* 1.20.4 -> current 2b2t version
* 1.21.1 -> active LTS of 1.21 for most mods
* (optional) latest -> currently 1.21.4

all other versions including 1.19.4 and 1.20.6 are used by very few players.

PR #4634, opened 2025-02-04
Expose current layer in IBuilderProcess

IceTank:
This exposes the current layer the build process in on.
This makes developing for external tools easier.

Thanks


leijurv:
seems fine

PR #4633, opened 2025-02-02
add(Settings): added settings to control block placement in fluid sources and flowing fluids

Lyzev:
### Summary
This PR introduces new settings to control block placement in fluid sources and flowing fluids within the Baritone pathfinding system.

### Changes
- **Settings.java**
  - Added `allowPlaceInFluidsSource` setting to control block placement in fluid source blocks.
  - Added `allowPlaceInFluidsFlow` setting to control block placement in flowing fluids.

- **CalculationContext.java**
  - Updated `costOfPlacingAt` method to respect the new settings `allowPlaceInFluidsSource` and `allowPlaceInFluidsFlow`.

### Motivation
These changes provide more granular control over block placement behavior in different fluid states, enhancing the flexibility and customization of the Baritone pathfinding system. Previously, the pathfinder could place blocks in flowing fluids, which could cause unintended rerouting of the fluids and potentially lead to dangerous situations, such as walking into lava. By introducing these settings, users can prevent block placement in fluid sources and flowing fluids, thereby improving the safety and reliability of the pathfinding process.

### Testing
- Verified that blocks are not placed in fluid sources when `allowPlaceInFluidsSource` is set to `false`.
- Verified that blocks are not placed in flowing fluids when `allowPlaceInFluidsFlow` is set to `false`.

### Related Issues
#2840 
There are probably more, but I'm too lazy to look for them.

Issue #2840, opened 2021-06-30
Disallow placing blocks which would redirect the flow of fluids


Shy2Soul:
## Describe your suggestion
Disallow placing blocks that would redirect the flow of fluids. This would prevent a lot of deaths in the nether by preventing lava from pouring on the player caused by a change in flow.

## Settings
Able to turn off/on, able to choose if the rule only applies to water or lava, or both.

## Context
Baritone sometimes dies in very stupid ways, this is one of them.

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

PR #4636, opened 2025-02-06
add(Settings): added settings to control block placement in fluid sources and flowing fluids

Lyzev:
> [!IMPORTANT] 
> This PR previously existed for version 1.21.4.  
>  
> First of all, thanks for your response in #4633!  
>  
> To address your questions (@ZacSharp):  
> Here is an example demonstrating the feature in action:  
> **Video:** https://www.youtube.com/watch?v=gQQUcJAelg0
>  
> The video also illustrates how this impacts both the pathfinding and builder processes. As shown, the builder process remains unaffected because the `costOfPlacingAt` method is overridden ([Source Code](https://github.com/cabaletta/baritone/blob/a5668532b16f2c2be0b9221b2e3038a0a807822c/src/main/java/baritone/process/BuilderProcess.java#L1114-L1146)).  
>  
> If anything remains unclear, I’d be happy to clarify further and make any necessary adjustments to this PR.  

### Summary
This PR introduces new settings to control block placement in fluid sources and flowing fluids within the Baritone pathfinding system.

### Changes
- **Settings.java**
  - Added `allowPlaceInFluidsSource` setting to control block placement in fluid source blocks.
  - Added `allowPlaceInFluidsFlow` setting to control block placement in flowing fluids.

- **CalculationContext.java**
  - Updated `costOfPlacingAt` method to respect the new settings `allowPlaceInFluidsSource` and `allowPlaceInFluidsFlow`.

### Motivation
These changes provide more granular control over block placement behavior in different fluid states, enhancing the flexibility and customization of the Baritone pathfinding system. Previously, the pathfinder could place blocks in flowing fluids, which could cause unintended rerouting of the fluids and potentially lead to dangerous situations, such as walking into lava. By introducing these settings, users can prevent block placement in fluid sources and flowing fluids, thereby improving the safety and reliability of the pathfinding process.

### Testing
- Verified that blocks are not placed in fluid sources when `allowPlaceInFluidsSource` is set to `false`.
- Verified that blocks are not placed in flowing fluids when `allowPlaceInFluidsFlow` is set to `false`.

### Related Issues
#2840 
There are probably more, but I'm too lazy to look for them.

### DIFF
--- a/src/main/java/baritone/pathing/movement/CalculationContext.java
+++ b/src/main/java/baritone/pathing/movement/CalculationContext.java
@@ -161,10 +161,10 @@ public class CalculationContext {
         if (!worldBorder.canPlaceAt(x, z)) {
             return COST_INF;
         }
-        if (current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsSource.value) {
+        if (!Baritone.settings().allowPlaceInFluidsSource.value && current.getFluidState().isSource()) {
             return COST_INF;
         }
-        if (!current.getFluidState().isEmpty() && !current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsFlow.value) {
+        if (!Baritone.settings().allowPlaceInFluidsFlow.value && !current.getFluidState().isEmpty() && !current.getFluidState().isSource()) {
             return COST_INF;
         }
         return placeBlockCost;
--- a/src/main/java/baritone/pathing/movement/CalculationContext.java
+++ b/src/main/java/baritone/pathing/movement/CalculationContext.java
@@ -161,6 +161,12 @@ public class CalculationContext {
         if (!worldBorder.canPlaceAt(x, z)) {
             return COST_INF;
         }
+        if (current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsSource.value) {
+            return COST_INF;
+        }
+        if (!current.getFluidState().isEmpty() && !current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsFlow.value) {
+            return COST_INF;
+        }
         return placeBlockCost;
     }
 
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -78,6 +78,16 @@ public final class Settings {
     public final Setting<Boolean> allowPlace = new Setting<>(true);
 
     /**
+     * Allow Baritone to place blocks in fluid source blocks
+     */
+    public final Setting<Boolean> allowPlaceInFluidsSource = new Setting<>(true);
+
+    /**
+     * Allow Baritone to place blocks in flowing fluid
+     */
+    public final Setting<Boolean> allowPlaceInFluidsFlow = new Setting<>(true);
+
+    /**
      * Allow Baritone to move items in your inventory to your hotbar
      */
     public final Setting<Boolean> allowInventory = new Setting<>(false);


src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -161,6 +161,12 @@ public double costOfPlacingAt(int x, int y, int z, BlockState current) {
         if (!worldBorder.canPlaceAt(x, z)) {
             return COST_INF;
         }
+        if (current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsSource.value) {
+            return COST_INF;
+        }
+        if (!current.getFluidState().isEmpty() && !current.getFluidState().isSource() && !Baritone.settings().allowPlaceInFluidsFlow.value) {
```
leijurv: Can the settings check be reordered to first rather than last in both of these ifs?

Lyzev: I hope the latest commit aligns with your expectations. The order has been updated as follows:
```java
if (!Baritone.settings().allowPlaceInFluidsSource.value && current.getFluidState().isSource()) {
    return COST_INF;
}
if (!Baritone.settings().allowPlaceInFluidsFlow.value && !current.getFluidState().isEmpty() && !current.getFluidState().isSource()) {
    return COST_INF;
}
```

ZacSharp:
Just noticed during testing on 1.21.2/1.21.4: What is the intended behavior for flowing source blocks and stationary non-source blocks?

Lyzev:
> Just noticed during testing on 1.21.2/1.21.4: What is the intended behavior for flowing source blocks and stationary non-source blocks?

I am not quite sure, what your question is about, but the allowPlaceInFluidsSource allows the path finder to place blocks in source blocks and allowPlaceInFluidsFlow allows to place in flowing fluids aka. non-source blocks. The video shows the behavior of the settings: https://www.youtube.com/watch?v=gQQUcJAelg0 

If you meant something else, please expand on your question.

ZacSharp:
> flowing fluids aka. non-source blocks

The question is about exactly that part. Source blocks can be flowing and non-source blocks can be stationary. (E.g. if you have a straight stream of water with a single source block, that source block is flowing as well. Similarly if you place two source blocks with 3 blocks of space between them, you get a stationary non-source in the middle)
When porting to newer version I noticed a behavior change, so I was wondering what the intended behavior for `allowPlaceInFluidsFlow` is. i.e. whether it should apply to any flowing liquid or to any non-source liquid.

(A flowing fluid block is one which pushes entities, a stationary fluid does not push entities)

Lyzev:
My idea was basically just that `allowPlaceInFluidsFlow` are all fluid blocks that are not a source. The main reason I came up with the idea in the first place was when I was playing on 2b2t and traveling in the Nether, the Path Finder killed itself about 3 times (2 of which I was fast enough to save myself from the fire damage). After I died, I decided to modify the code (which led to this PR as I figured I wasn't the only one having this problem). I wanted the Path Finder to stop placing blocks in fluids that are not a source (even if they are stationary), they could redirect the fluid/lava. The issue doesn't appear for source fluid blocks since they don't reroute when placed into.

Therefore, `allowPlaceInFluidsFlow` should include all non-source fluid block types. 

But that's just my opinion, and I also feel like that in the case of `source blocks with 3 blocks of space between them, you get a stationary non-source in the middle`, they still push, but when you're in between the direction, cancel each other out velocity wise. (I may be incorrect at this part)

ZacSharp:
Thank you for the clarification. I hope I managed to preserve that behavior.

> they still push, but when you're in between the direction, cancel each other out velocity wise

Uh, I forgot to mention that the water should be placed in a canal so it doesn't escape to the sides.

PR #4639, opened 2025-02-09
fix chuckpacker y bug

babbaj:
`y` in `pack` is already normalized to 0 - 384 and `getPathingBlockType` was trying to normalize it again, causing `adjY` to be offset 64 blocks, which caused an oobe in the call to `BlockStateInterface.getFromChunk`.
The exception got caught but the result seems to be that special blocks close to the build limit didn't get cached.
This probably also broke the caching of water.

### DIFF
--- a/src/main/java/baritone/cache/ChunkPacker.java
+++ b/src/main/java/baritone/cache/ChunkPacker.java
@@ -78,7 +78,7 @@ public final class ChunkPacker {
                         for (int x = 0; x < 16; x++) {
                             int index = CachedChunk.getPositionIndex(x, y, z);
                             BlockState state = bsc.get(x, y1, z);
-                            boolean[] bits = getPathingBlockType(state, chunk, x, y, z).getBits();
+                            boolean[] bits = getPathingBlockType(state, chunk, x, y + chunk.getMinBuildHeight(), z).getBits();
                             bitSet.set(index, bits[0]);
                             bitSet.set(index + 1, bits[1]);
                             Block block = state.getBlock();


babbaj:
https://github.com/cabaletta/baritone/commit/d22a52976b2988a0daa7ec8324de32ab2c095d0c

everyone point and laugh at this user for introducing this bug

ZacSharp:
Did GitHub miss a usage of `getPathingBlockType` or are we now denormalizing the y coordinate just to renormalize it immediately in the called method?

leijurv:
Could someone verify a before/after that this actually works now and didn't before? Shouldn't be that hard, just enable `pathThroughCachedOnly`

ZacSharp:
Most cases are handled looking just at the passed-in state, so finding a difference is quite hard actually.
Here is one I found:

https://github.com/user-attachments/assets/32a349db-2b57-4ae3-b2f6-b13133cce8fb

The fixed version does look better, sure, but it's wrong. Flowing water is supposed to be `AVOID` and cornering over that is not allowed.
Testing this is difficult since both versions are buggy:


https://github.com/user-attachments/assets/88e3cbd7-38e6-4457-8a73-f4ad2c393933

The "type with absolute/relative y" lines are just similar to what the packer does with/without the fix. The real behavior is even worse (look at the bsi state):


https://github.com/user-attachments/assets/82b7d0d2-db43-4b04-b596-6316607fdbc9

babbaj:
I can't see how this patch could be wrong. I bet there's another bug that isn't as obvious as this one.

ZacSharp:
> I can't see how this patch could be wrong.

Me neither. Just saying that I cannot confirm it empirically because both versions are buggy.

I bet the problem is somewhere [here](https://github.com/cabaletta/baritone/blob/9de30740411ff636a3ca763446f1d51b70d6963d/src/main/java/baritone/cache/ChunkPacker.java#L127-L143). It explicitly behaves differently at chunk borders and that flow check looks very suspicious. Either `flow` has a weird meaning or this is testing for horizontal flow.

ZacSharp:
<details><summary><code>possiblyFlowing</code> is broken</summary>




</details>

It seems to check whether the block is a non-source fluid.

The screenshot also confirms that `getFlow` does indeed return the direction of flow so here the packer explicitly returns `WATER` for horizontally flowing water and `AVOID` for falling water. (nested inside other conditions)
https://github.com/cabaletta/baritone/blob/9de30740411ff636a3ca763446f1d51b70d6963d/src/main/java/baritone/cache/ChunkPacker.java#L137-L141

PR #4640, opened 2025-02-09
add(ToolSet): implement fortune level check for crop harvesting

Lyzev:
### Summary
This PR introduces new settings to control farming crops with silk touch.

### Changes
- **Settings.java**
  - Added `useFortuneForCrops` setting to control the behavior.

- **ToolSet.java**
  - Updated `getBestSlot(Block, boolean, boolean)` method to use the highest level fortune tool when farming a crop.

### Motivation  
Enhances the `#farm` command and other mechanics, increasing the amount, you get, from your crops.  

As far as I know, this does not consume durability, so there’s no downside to adding it - making it an obvious improvement.

For more details, check out the Minecraft Wiki: [Fortune and Seeds](https://minecraft.wiki/w/Fortune#Seeds).  

### Testing
- Verified that crops are farmed by the highest fortune level tool.
- Verified that breaking other blocks are not affected by this change.

### DIFF
--- a/src/main/java/baritone/utils/ToolSet.java
+++ b/src/main/java/baritone/utils/ToolSet.java
@@ -139,29 +139,29 @@ public class ToolSet {
             if (Baritone.settings().itemSaver.value && (itemStack.getDamageValue() + Baritone.settings().itemSaverThreshold.value) >= itemStack.getMaxDamage() && itemStack.getMaxDamage() > 1) {
                 continue;
             }
-            int fortuneLevel = getFortuneLevel(itemStack);
-            if (isCrop && fortuneLevel > highestFortune && Baritone.settings().useFortuneForCrops.value) {
-                highestFortune = fortuneLevel;
+
+            double speed = calculateSpeedVsBlock(itemStack, blockState);
+            boolean silkTouch = hasSilkTouch(itemStack);
+            if (speed > highestSpeed) {
+                highestSpeed = speed;
                 best = i;
                 lowestCost = getMaterialCost(itemStack);
-                bestSilkTouch = hasSilkTouch(itemStack);
-            } else if (highestFortune == 0) {
-                double speed = calculateSpeedVsBlock(itemStack, blockState);
-                boolean silkTouch = hasSilkTouch(itemStack);
-                if (speed > highestSpeed) {
+                bestSilkTouch = silkTouch;
+            } else if (speed == highestSpeed) {
+                int cost = getMaterialCost(itemStack);
+                int fortuneLevel = getFortuneLevel(itemStack);
+                if (Baritone.settings().useFortuneForCrops.value && isCrop && ((fortuneLevel >= highestFortune && cost < lowestCost) || (fortuneLevel > highestFortune && highestFortune == 0))) {
                     highestSpeed = speed;
                     best = i;
-                    lowestCost = getMaterialCost(itemStack);
+                    lowestCost = cost;
                     bestSilkTouch = silkTouch;
-                } else if (speed == highestSpeed) {
-                    int cost = getMaterialCost(itemStack);
-                    if ((cost < lowestCost && (silkTouch || !bestSilkTouch)) ||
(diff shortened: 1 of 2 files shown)


src/main/java/baritone/utils/ToolSet.java
```diff
@@ -132,21 +139,29 @@ possible, this lets us make pathing depend on the actual tool to be used (if aut
...
+            if (isCrop && fortuneLevel > highestFortune && Baritone.settings().useFortuneForCrops.value) {
+                highestFortune = fortuneLevel;
                 best = i;
                 lowestCost = getMaterialCost(itemStack);
-                bestSilkTouch = silkTouch;
-            } else if (speed == highestSpeed) {
-                int cost = getMaterialCost(itemStack);
-                if ((cost < lowestCost && (silkTouch || !bestSilkTouch)) ||
-                        (preferSilkTouch && !bestSilkTouch && silkTouch)) {
+                bestSilkTouch = hasSilkTouch(itemStack);
```
ZacSharp: I think the way this is currently written these two are not needed.

ZacSharp (changes requested):
This makes tool selection completely ignore secondary criteria as soon as crops/fortune are involved. This could be fine for speed because crops break instantly anyway, but even if that's true without exceptions for vanilla minecraft mods do exist. For material costs this does matter, since according to the "if everything else is equal, prefer the cheaper tool" policy used so far it should choose wooden hoes with fortune 3 over diamond hoes with fortune 3.

I'm also concerned about the comparison logic growing more and more branches, but maybe I'll address that myself in a separate pr.


And can you please not autogenerate your comments?

Lyzev:
> This makes tool selection completely ignore secondary criteria as soon as crops/fortune are involved. This could be fine for speed because crops break instantly anyway, but even if that's true without exceptions for vanilla minecraft mods do exist. For material costs this does matter, since according to the "if everything else is equal, prefer the cheaper tool" policy used so far it should choose wooden hoes with fortune 3 over diamond hoes with fortune 3.

To be honest, I didn't have any mods in mind that might add crops that can't be insta-broken. Also, I didn't think the material cost was really relevant since breaking crops doesn't use durability (if insta-broken ofc).

I made a new commit that addresses this problem. The auto tool now prefers the lower material cost if the fortune levels and speed are equal. Since I moved the `if` to a different position, it also prefers speed over fortune (As far as I understand that is what you want). So if a mod adds a crops that can't be broken instantly, it will prefer speed over fortune.

If there are any other concerns, feel free to tell me.

Lyzev:
I was thinking that it would probably make more sense to make a `preferFortune` setting and a setting to select the blocks where fortune tools should be preferred and keep the `preferSilktouch` setting, but add a setting to select specific blocks where it should be preferred. This probably gives better control over the auto tool. Is this a solution that would like to be seen?

ZacSharp:
> it also prefers speed over fortune (As far as I understand that is what you want)

I expected fortune,speed,cost, but I'm fine with speed first as well. (Actually, I think speed first is better since `preferSilktouch` already works that way)

Better control would definitely be a desired feature, though there's a lot of things people want or might want to Baritone to do, so if you can come up with a way simple idea for more flexible configuration that would be even better. (e.g. they might want to have cost be the most important criterium, as long as the tool is usable for the block)

An idea would be a `Map<Block, List<Criterium>>` setting, but that might be tedious to configure. A simpler alternative would be to have the `List<Block>` type settings you proposed, but an extra `List<Criterium>` setting to allow reordering them.

ZacSharp:
Anyway this feature has been in design limbo for long enough, so if you don't feel like expanding the scope of this pr that far, I'd say `preferFortune` itself can be the list of blocks (but doing the same for `preferSilktouch` would be a breaking change).

Lyzev:
I'm too busy to focus on this as I have many other projects and other things to do at the moment. I'm closing this PR for now, but I may come back with a better solution in the future.

PR #4662, opened 2025-02-27
Fix `MineProcess` stalling on blocks which are `canWalkThrough` but shouldn't be broken

### LINKED ISSUE #4654: Baritone can't mine flowers next to the world boarder
## Some information
Operating system:
Java version: Runtime 21.0.1
Minecraft version: 1.21.3
Baritone version: 1.11.1 (latest)
Other mods (if used): Plenty honestly. It's a lot of fps improvement mods and other client side mods. I don't think this issue is relating to any mod interference but I'll write them here if you need me to.

## How to reproduce
For some reason Britone is unable to mine flowers next to the world boarder. It paths to the flower just fine, but it doesn't break it when it arrives. I have to break it manually.

The flower must be next to the world boarder, not in it or past it. However if the flower is only 1 block away or more. It can mine it just fine.

## Modified settings
It's also alot. I'll add them if I need to.

## Final checklist
- [ ] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [ ] I have included logs, exceptions and / or steps to reproduce the issue.

ZacSharp:
Before breaking a block, `MineProcess` checks `MovementHelper.avoidBreaking` to avoid falling blocks and similar problems. Normally it also does that check before setting the goal implicitly via `MovementHelper.getMiningDurationTicks`. However that method (correctly) short-circuits for `canWalkThrough` blocks so such blocks could end up in the goal, even if they cannot be mined.

Note: This fixes the linked issue by completely ignoring blocks next to the world border, which is consistent with the pathfinder and mining solid blocks. Doing so might be considered a bug, because those blocks *are* breakable.

Fixes #4654

### DIFF
--- a/src/main/java/baritone/process/MineProcess.java
+++ b/src/main/java/baritone/process/MineProcess.java
@@ -77,7 +77,6 @@ public final class MineProcess extends BaritoneProcessHelper implements IMinePro
             int curr = ctx.player().getInventory().items.stream()
                     .filter(stack -> filter.has(stack))
                     .mapToInt(ItemStack::getCount).sum();
-            System.out.println("Currently have " + curr + " valid items");
             if (curr >= desiredQuantity) {
                 logDirect("Have " + curr + " valid items");
                 cancel();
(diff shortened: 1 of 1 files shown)


pennilessSplatter:
Thanks

Issue #1089, opened 2019-10-23
'Block' parameters in commands should support tags


MMK21Hub:
## Describe your suggestion
So when you select a block in a command such as `#mine oak_log`, it would be useful if you could specify a [Minecraft tag](https://minecraft.gamepedia.com/Tag) instead, such as `#mine #logs` or `.b mine #logs` in Impact.

## Settings
N/A

## Context
This will be useful because after 1.13, many block types have been split off and you will sometimes want to select a group of blocks (e.g. any wool colour or any log type) to path to.

## Final checklist
- [✔] I have not used any OwO's or UwU's in this issue.

end-user:
This has also become more important as block names are getting longer (cracked_polished_blackstone_bricks) and the chat input is fixed length.

shinji257:
I'm going to also request this.  I may drop this into modpacks and they tend to have many versions of the same ore integrated into different block types.  It would be easier if I could just reference the forge tag for the ore that typically encompasses them all.

Sqaaakoi:
I'm also asking for this too, as deepslate ores are annoying to manually specify and make the command longer

weegeeday:
is this implemented? Doesn't seem to currently work.

HungryBoy02:
> is this implemented? Doesn't seem to currently work.

no, the issue is still open.

ghost:
+1 Request, would be a super useful feature especially for modded

PR #4678, opened 2025-03-17
Add regex support for mine command

vasyza:
## Overview
This pull request adds regular expression support to the `mine` command, allowing users to mine blocks that match a specific pattern.

## Features
- Added a new `-r` flag to the `mine` command to indicate that the following argument is a regular expression
- Implemented pattern matching against block IDs in the Minecraft registry
- Added informative feedback messages about the number of blocks matched by the regex
- Updated tab completion to suggest the `-r` flag and common regex patterns
- Updated command documentation to include the new regex functionality

## Usage Examples
- `mine -r .*ore.*` - Mines all blocks that have "ore" in their ID (e.g., diamond_ore, iron_ore, gold_ore, etc.)
- `mine -r .*log.*` - Mines all types of logs
- `mine -r .*stone.*` - Mines all types of stone blocks

### DIFF
--- a/src/main/java/baritone/command/defaults/MineCommand.java
+++ b/src/main/java/baritone/command/defaults/MineCommand.java
@@ -24,10 +24,13 @@ import baritone.api.command.argument.IArgConsumer;
 import baritone.api.command.datatypes.ForBlockOptionalMeta;
 import baritone.api.command.exception.CommandException;
 import baritone.api.utils.BlockOptionalMeta;
+import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.world.level.block.Block;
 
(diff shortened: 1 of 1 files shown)


ZacSharp:
Firstly, all pullrequests should target the 1.19.4 branch and will be ported from there. (Likely by me)

Flag style arguments are so far not used by Baritone and I don't think we should add them (rather stay similar to Minecraft commands). Another option would be to fall back to a regex when no direct match was found.

More fundamentally though, I do not think regex makes for a good user interface. If we have the data available on the client I'd prefer #1089 (and/or a way do define custom tags maybe). Full regex is just overkill for most people and cases.

PR #4692, opened 2025-03-30
1.21.5

rfresh2:
- [x] Fabric
- [x] Forge
- [x] NeoForge  
- [ ] ~~Goal Beacon Renderer~~ (actually has been broken since at least 1.20.4)
- [x] Testing (unable to test litematica as it has not been ported yet)


ZacSharp:
Also need to bump the end of the version range in SETUP.md under "Command Line".

rfresh2:
goal beacon renderer has actually been broken since at least 1.20.4

so i dont think ill resolve that in this pr, problem for another day

src/launch/java/baritone/launch/mixins/MixinScreen.java
```diff
@@ -36,15 +33,19 @@
...
+    @Inject(method = "handleComponentClicked", at = @At(
+        value = "INVOKE",
+        target = "Lnet/minecraft/client/multiplayer/ClientPacketListener;sendUnsignedCommand(Ljava/lang/String;)Z"
+    ), cancellable = true)
+    public void handleCustomClickEvent(final Style style, final CallbackInfoReturnable<Boolean> cir) {
         ClickEvent clickEvent = style.getClickEvent();
         if (clickEvent == null) {
             return;
         }
+        if (!(clickEvent instanceof ClickEvent.RunCommand(String command))) return;
```
ZacSharp, line 47: Cursed, but kinda nice?
Didn't know we can pattern match now. (only knew about `o instanceof C o_`)

src/main/java/baritone/utils/IRenderer.java
```diff
@@ -52,23 +79,8 @@ static void glColor(Color color, float alpha) {
     }
 
     static BufferBuilder startLines(Color color, float alpha, float lineWidth, boolean ignoreDepth) {
```
ZacSharp: `ignoreDepth` here is now completely ignored

src/main/java/baritone/utils/PathRenderer.java
```diff
@@ -266,20 +261,22 @@ private static void drawGoal(@Nullable BufferBuilder bufferBuilder, PoseStack st
...
-                textureManager.getTexture(TEXTURE_BEACON_BEAM).bind();
+                // todo: fix beacon renderer (has been broken since at least 1.20.4)
+                //  issue with outer beam rendering, probably related to matrix transforms state not matching vanilla
+                //  possible solutions:
+                //      inject hook into LevelRenderer#renderBlockEntities where the matrices have already been set up correctly
+                //      copy out and modify the vanilla beacon render code
                 if (settings.renderGoalIgnoreDepth.value) {
-                    RenderSystem.disableDepthTest();
+
                 }
```
ZacSharp: Leaving this (and the one below) behind as a reminder?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -58,9 +58,13 @@ public LitematicaSchematic(CompoundTag nbt) {
...
-        return nbt.getCompound("Regions").getAllKeys().stream()
-                .map(nbt.getCompound("Regions")::getCompound)
-                .toArray(CompoundTag[]::new);
+        return nbt.getCompound("Regions")
+            .map(CompoundTag::values)
+            .map(r -> r.stream()
+                .filter(v -> v instanceof CompoundTag)
+                .map(CompoundTag.class::cast)
+                .toArray(CompoundTag[]::new)
+            ).orElse(new CompoundTag[0]);
```
ZacSharp, line 67: This looks like it also fixes an NPE with malformed schematics. Should this be backported?

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
```diff
@@ -105,9 +109,9 @@ private static BlockState[] getBlockList(ListTag blockStatePalette) {
     private static BlockState getBlockState(Block block, CompoundTag properties) {
         BlockState blockState = block.defaultBlockState();
 
-        for (Object key : properties.getAllKeys()) {
-            Property<?> property = block.getStateDefinition().getProperty((String) key);
-            String propertyValue = properties.getString((String) key);
+        for (String key : properties.keySet()) {
+            Property<?> property = block.getStateDefinition().getProperty(key);
```
ZacSharp, line 113: `Object` -> `String` replacement also backportable?

ZacSharp (approved):
Won't pretend I thoroughly tested this or understand the new rendering system, but the rest looks good.

ZacSharp:
There is a bug report for this (see https://github.com/cabaletta/baritone/issues/4697#issuecomment-2812397099), though I'm not yet sure whether that bug is actually introduced here.

ZacSharp:
The bug also exists on previous versions.

src/launch/java/baritone/launch/mixins/MixinScreen.java
florianreuth: Quite sure this breaks if a book for example contains a click event running a command (as you are cancelling the original code away)

ZacSharp: You are right, this cancels any commands, even if Baritone doesn't handle them.

Now  I do wonder though: Didn't the previous injection do the same thing?

florianreuth: No because the injection point was different, the new injection point is also wrong I think. The code was previous only called when the command sending failed I think.

ZacSharp: Ok, fell for a little gotcha (tried to use `/say hi` in a click event) and the command generator I used was broken, but I managed to confirm that Baritone does correctly handle clickable text in 1.19.4 and does not allow running vanilla commands in 1.21.5.

Minecraft (<=1.21.4) used to display an error message when running a chat command without a leading slash, so my guess is that the previous injection point was exactly that error message, meaning it would only run for Baritone commands (or from anything else using anything but "/" as its prefix).
Problem with that behavior is that Minecraft (>=1.21.5) allows commands without prefix in click events now.

ZacSharp: > the new injection point is also wrong I think

Because it is triggered for vanilla commands as well or is there another reason?

florianreuth: > > the new injection point is also wrong I think
> 
> Because it is triggered for vanilla commands as well or is there another reason?

This was where the old injection was placed, so the code was never called when the command sending succeeded.
image

ZacSharp: Maybe both the highlighted line and the second `LOGGER.error` call were targeted? Otherwise e.g. the text from Baritones help menu was broken.

I have a potential fix, I'm just trying to fit it into GitHubs incredibly limited "suggest a diff" feature.

ZacSharp: Maybe both the highlighted line and the second `LOGGER.error` call were targeted? Otherwise e.g. the text from Baritones help menu cannot work (it starts commands with "<<", definitely not with "/").

florianreuth: Nah the ordinal was set to 1, meaning only the highlighted one was targetted. They changed this behaviour already in the past (in 1.19 versions) so I could imagine the Mixin silently broke without anyone noticing. (For the exact behaviour changes, see https://github.com/ViaVersion/ViaFabricPlus/blob/main/src/main/java/com/viaversion/viafabricplus/injection/mixin/features/run_command_action/MixinScreen.java#L46)

ZacSharp: Here's what I think should make this work. Not usable as is because it needs an additional import statement.
```suggestion
        if (!(clickEvent instanceof ClickEvent.RunCommand(String command))) return;
        if (!command.startsWith(FORCE_COMMAND_PREFIX)) {
            return;
        }
```

ZacSharp: Isn't `ordinal` zero-based and the two calls in the screenshot the only candidates, so the second one is targeted?

florianreuth: There is another logger usage above which is not part of the screenshot

ZacSharp: That one (~and the fourth one below~ Nope, only the first one) has the wrong signature.

ZacSharp:
If you rebase and resolve the conflicts the problem with clickable text should be gone. (I might be resposible for some of the conflicts, sorry)

Litematica has also updated in the meantime and our Litematica integration seems to be functional.

src/main/java/baritone/utils/schematic/format/defaults/LitematicaSchematic.java
rfresh2: up to you ig

its not necessarily incorrect for malformed inputs to throw exceptions

this optional chaining is more natural to do now that the nbt api is returning them

but i think there is existing code up the stack that would catch and log those exceptions anyway

rfresh2:
rebased, everything seems to still work

src/main/java/baritone/command/defaults/SetCommand.java
```diff
@@ -188,12 +188,11 @@ public void execute(String label, IArgConsumer args) throws CommandException {
             MutableComponent oldValueComponent = Component.literal(String.format("Old value: %s", oldValue));
             oldValueComponent.setStyle(oldValueComponent.getStyle()
                     .withColor(ChatFormatting.GRAY)
-                    .withHoverEvent(new HoverEvent(
-                            HoverEvent.Action.SHOW_TEXT,
+                    .withHoverEvent(new HoverEvent.ShowText(
+
```
ZacSharp: Blank line

ZacSharp (approved):

picarica:
i tested your build for 2 weeks in 1.21.5 didnt experience any issues, but i didnt even know there is some litematica integration? is there some info for it ?

ZacSharp:
You can use `#litematica` to let Baritone build Litematica placements.

PR #4698, opened 2025-04-04
Elytra in overworld

### LINKED ISSUE #4187: Add support for auto-flight with Elytras in the overworld and the End.
## Describe your suggestion
I've just discovered how to go from one point to the other of the Nether automatically flying with the Elytras, this mod is a work of engineering, my respects to the developer.

Although being able to fly only in the Nether is quite limiting, it would be nice to be able to fly in the overworld and the End as well.

## Settings
I guess it would work the same as it works so far, if a specific command needs to be made for each dimension then there would be no problem.

## Context
It would be great to be able to fly automatically in the overworld and in the End, especially for exploration purposes. In a survival world with cheats enabled for example, after locating a structure with the command, we could fly there automatically. Also, after doing a long exploration on foot and having travelled thousands of blocks, it would be great to be able to fly back home automatically with the Elytras. 

This would undoubtedly make the mod much, much better, as it would not only be a mining and auto-build mod, it would be an exploration mod and that would attract a different audience, therefore more downloads and therefore more donations.

## Final checklist

### LINKED ISSUE #4867: Issue with elytra command
I apologize if my report may be unclear, forgive me, I was asked to report a bug I made.
Good luck!

[2025-09-07-5.log.gz](https://github.com/user-attachments/files/22199418/2025-09-07-5.log.gz)

[19:34:21] [Render thread/INFO]: [CHAT] [Baritone] > elytra
[19:34:21] [Render thread/INFO]: [CHAT] [Baritone] To disable this message, enable the setting elytraTermsAccepted\nBaritone Elytra is an experimental feature. It is only intended for long distance travel in the Nether using fireworks for vanilla boost. It will not work with any other mods ("hacks") for non-vanilla boost. If you want Baritone to attempt to take off from the ground for you, you can enable the elytraAutoJump setting (not advisable on laggy servers!). If you want Baritone to go slower, enable the elytraConserveFireworks setting and/or decrease the elytraFireworkSpeed setting. Baritone Elytra wants to know the seed of the world you are in. If it doesn't have the correct seed, it will frequently backtrack. It uses the seed to generate terrain far beyond what you can see, since terrain obstacles in the Nether can be much larger than your render distance. \nBaritone Elytra is not predicting terrain. If you don't know the seed, this is the correct thing to do. If you do know the seed, input it with #set elytraNetherSeed seedgoeshere, and then enable it with #set elytraPredictTerrain true
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: java.lang.IllegalArgumentException: Invalid y1 or y2
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at knot//dev.babbaj.pathfinder.NetherPathfinder.pathFind(Native Method)
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at knot//baritone.ep.a(Unknown Source)
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
[19:34:21] [pool-40-thread-1/INFO]: [STDERR]: 	at java.base/java.lang.Thread.run(Thread.java:1583)
[19:34:21] [Render thread/INFO]: [CHAT] [Baritone] An unhandled exception occurred. The error is in your game's log, please report this at https://github.com/cabaletta/baritone/issues

Mods:
  [✔] appleskin-fabric-mc1.21.6-3.0.6
  [✔] architectury-17.0.8-fabric
  [✔] Axiom-5.0.1-for-MC1.21.8
  [✔] badpackets-fabric-0.9.0
  [✔] baritone-api-fabric-1.15.0
  [✔] baritone-standalone-fabric-1.15.0
  [✔] BetterF3-15.0.0-Fabric-1.21.6
  [✔] better-pie-chart-1.1.0+1.21.6
  [✔] betterstats-4.0.0+fabric-1.21.8
  [✔] c2me-fabric-mc1.21.8-0.3.5+alpha.0.8
  [✔] capes-1.5.9+1.21.8-fabric
  [✔] Carpet-AMS-Addition-mc1.21.8-fabric-v2.57.6
  [✔] carpet-extra-extras-1.21.8-1.2.3
  [✔] carpet-org-addition-mc1.21.8-v1.38.0-2507192331
  [✔] carpet-tis-addition-v1.68.0-mc1.21.8
  [✔] ChunkDebug-2.4.2+1.21.6
  [✔] cicada-lib-0.14.2+1.21.6-and-above
  [✔] clientcommands-2.11.3
  [✔] cloth-config-19.0.147-fabric
  [✔] continuity-3.0.1-beta.1+1.21.6
  [✘] DistantHorizons-2.3.4-b-1.21.8-fabric-neoforge.jar (disabled)
  [✔] do_a_barrel_roll-fabric-3.8.3+1.21.6
  [✔] Drip Sounds-0.5.1+1.21.5-Fabric
  [✔] dynamic-fps-3.9.6+minecraft-1.21.6-fabric
  [✔] elytratrims-fabric-4.3.2+1.21.8
  [✔] emotecraft-fabric-for-MC1.21.7-3.0.0-a.build.112
  [✔] emotecraft-flashback-addon-1.1.0
  [✔] fabric-api-0.132.0+1.21.8
  [✔] fabric-bind-cmd-1.7.2-1.21.8
  [✔] fabric-carpet-1.21.7-1.4.177+v250630
  [✔] fabric-language-kotlin-1.13.5+kotlin.2.2.10
  [✔] FallingTree-1.21.8-1.21.8.2
  [✔] fastquit-3.1.1+mc1.21.6
  [✔] ferritecore-8.0.0-fabric
  [✔] Flashback-0.37.0-for-MC1.21.8
  [✔] forcecloseloadingscreen-2.3.0
  [✔] freecam-fabric-1.3.4+mc1.21.6
  [✔] fzzy_config-0.7.2+1.21.6
  [✘] g4mespeed-1.4.11-mc1.21.8.jar (disabled)
  [✔] Gamma-Utils-2.3.5+mc1.21.6
  [✔] ImmediatelyFast-Fabric-1.12.1+1.21.8
  [✔] iris-fabric-1.9.1+mc1.21.7
  [✔] itemscroller-fabric-1.21.8-0.28.3
  [✘] Jade-1.21.8-Fabric-19.3.2.jar (disabled)
  [✔] Ksyxis-1.3.4
  [✔] language-reload-1.7.4+1.21.6
  [✔] litematica-fabric-1.21.8-0.23.3
  [✔] litematica-printer-1.21.8-3.2.1-sakura.8
  [✔] lithium-fabric-0.18.0+mc1.21.8
  [✔] malilib-fabric-1.21.8-0.25.4
  [✔] midnightlib-1.7.5-fabric+1.21.6
  [✔] minihud-fabric-1.21.8-0.36.3
  [✔] modmenu-15.0.0-beta.3
  [✔] moreculling-fabric-1.21.6-1.4.0-beta.1
  [✔] optigui-2.3.0-beta.7+1.21.6
  [✔] owo-lib-0.12.22+1.21.8
  [✔] packetfixer-3.1.4-1.20.5-1.21.X-merged
  [✔] particle_core-0.2.6+1.21.6
  [✘] particlerain-4.0.0-alpha.3+1.21.6-fabric.jar (disabled)
  [✘] particular-1.1.1+1.21.7.jar (disabled)
  [✔] placeholder-api-2.7.1+1.21.6
  [✔] PlayerAnimationLib-1.0.7+mc1.21.7
  [✔] puzzle-fabric-2.1.1+1.21.6
  [✔] reeses-sodium-options-fabric-1.8.4+mc1.21.6
  [✔] rei-search-bar-calculations-1.0.2+1.21.6-8
  [✔] Resourcify (1.21.8-fabric)-1.7.4
  [✔] RoughlyEnoughItems-20.0.811-fabric
  [✔] skinlayers3d-fabric-1.9.0-mc1.21.6
  [✔] sodium-extra-fabric-0.6.6+mc1.21.6
  [✔] sodium-fabric-0.6.13+mc1.21.6
  [✔] sound-physics-remastered-fabric-1.21.8-1.4.15
  [✔] soundsbegone-fabric-1.4.7+mc1.21.6
  [✔] syncmatica-fabric-1.21.8-0.3.15
  [✔] tcdcommons-4.0.1+fabric-1.21.8
  [✔] tweakermore-v3.26.4-mc1.21.8
  [✔] tweakeroo-fabric-1.21.8-0.25.3
  [✔] vmp-fabric-mc1.21.8-0.2.0+beta.7.207-all
  [✔] voicechat-fabric-1.21.8-2.5.35
  [✔] vt-downloader-2.3.5+1.21.6
  [✘] wakes-0.4.2+1.21.8.jar (disabled)
  [✔] worldedit-mod-7.3.16
  [✔] wthit-fabric-16.0.1
  [✔] Xaeros_Minimap_25.2.12_Fabric_1.21.8
  [✔] XaerosWorldMap_1.39.13_Fabric_1.21.8
  [✔] yet_another_config_lib_v3-3.7.1+1.21.6-fabric
  [✔] Zoomify-2.14.4+1.21.6

Params:
  --username  --version 1.21.8 --gameDir C:/Users/Admin/AppData/Roaming/PrismLauncher/instances/Performium 1.21.8/minecraft --assetsDir C:/Users/Admin/AppData/Roaming/PrismLauncher/assets --assetIndex 26 --uuid  --accessToken  --userType  --versionType release

Window size: 854 x 480

Launcher: standard

Java Arguments:
[-XX:HeapDumpPath=MojangTricksIntelDriversForPerformance_javaw.exe_minecraft.exe.heapdump, -Xms512m, -Xmx49152m, -Duser.language=en]

-- System Details --
Details:
	Minecraft Version: 1.21.4
	Minecraft Version ID: 1.21.4
	Operating System: Windows 11 (amd64) version 10.0
	Java Version: 21.0.7, Microsoft
	Java VM Version: OpenJDK 64-Bit Server VM (mixed mode), Microsoft
	Memory: 3671767832 bytes (3501 MiB) / 6174015488 bytes (5888 MiB) up to 51539607552 bytes (49152 MiB)
	CPUs: 24
	Processor Vendor: GenuineIntel
	Processor Name: 13th Gen Intel(R) Core(TM) i7-13700K
	Identifier: Intel64 Family 6 Model 183 Stepping 1
	Microarchitecture: Raptor Lake
	Frequency (GHz): 3.42
	Number of physical packages: 1
	Number of physical CPUs: 16
	Number of logical CPUs: 24
	Graphics card #0 name: Intel(R) UHD Graphics 770
	Graphics card #0 vendor: Intel Corporation
	Graphics card #0 VRAM (MiB): 2048.00
	Graphics card #0 deviceId: VideoController1
	Graphics card #0 versionInfo: 32.0.101.6129
	Graphics card #1 name: NVIDIA GeForce RTX 4080 SUPER
	Graphics card #1 vendor: NVIDIA
	Graphics card #1 VRAM (MiB): 16376.00
	Graphics card #1 deviceId: VideoController2
	Graphics card #1 versionInfo: 32.0.15.7602
	Memory slot #0 capacity (MiB): 16384.00
	Memory slot #0 clockSpeed (GHz): 4.80
	Memory slot #0 type: DDR5
	Memory slot #1 capacity (MiB): 16384.00
	Memory slot #1 clockSpeed (GHz): 4.80
	Memory slot #1 type: DDR5
	Memory slot #2 capacity (MiB): 16384.00
	Memory slot #2 clockSpeed (GHz): 4.80
	Memory slot #2 type: DDR5
	Memory slot #3 capacity (MiB): 16384.00
	Memory slot #3 clockSpeed (GHz): 4.80
	Memory slot #3 type: DDR5
	Virtual memory max (MiB): 69394.59
	Virtual memory used (MiB): 38164.73
	Swap memory total (MiB): 4096.00
	Swap memory used (MiB): 165.96
	Space in storage for jna.tmpdir (MiB): <path not set>
	Space in storage for org.lwjgl.system.SharedLibraryExtractPath (MiB): <path not set>
	Space in storage for io.netty.native.workdir (MiB): <path not set>
	Space in storage for java.io.tmpdir (MiB): available: 1480537.13, total: 1952557.00
	Space in storage for workdir (MiB): available: 1480537.13, total: 1952557.00
	JVM Flags: 3 total; -XX:HeapDumpPath=MojangTricksIntelDriversForPerformance_javaw.exe_minecraft.exe.heapdump -Xms512m -Xmx49152m

### LINKED ISSUE #4971: Nether roof elytra
## Description
Flying in nether with Elytra is fine but consuming lots of firework for far distance, how about making it flyable on nether roof?

## Settings
"#elytraNetherRoof" command for toggle or best with automatically detect whether the player is on the nether roof or not.

## Context
Right now, elytra command will issue error when used on nether roof. For lazy players like me who hate walking or using elytra for long distance even the surface is flat and no obstacles, it would be awesome if elytra can be used on nether roof and know how to glide up and down to reduce fireworks consumption instead of straight line.

## Another suggestion
Since elytra in nether is already doing a good job, it would be a waste if it's not useable in overworld and the end. I have seen other forks which are improved versions of baritone but I prefer original one.

## Final checklist

babbaj:
Started by @underscore-zi 
nether-pathfinder changes have been merged in https://github.com/babbaj/nether-pathfinder/pull/20
Closes #4187

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1553,6 +1553,10 @@ public final class Settings {
      */
     public final Setting<Boolean> elytraAllowTightSpaces = new Setting<>(false);
 
+    /**
+     * Allow the pathfinder to fly above y 128 in the nether.
+     */
+    public final Setting<Boolean> elytraAllowAboveRoof = new Setting<>(false);
 
     /**
      * A map of lowercase setting field names to their respective setting
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1561,7 +1561,7 @@ public final class Settings {
     /**
      * Allow the pathfinder to access the baritone cache to improve pathing
      */
-    public final Setting<Boolean> elytraUseBaritoneCache = new Setting<>(true);
+    public final Setting<Boolean> elytraUseCache = new Setting<>(true);
 
     /**
      * A map of lowercase setting field names to their respective setting
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1564,6 +1564,16 @@ public final class Settings {
     public final Setting<Boolean> elytraUseCache = new Setting<>(true);
 
     /**
+     * Allow the pathfinder to fly above the build limit in the overworld and end.
+     */
+    public final Setting<Boolean> elytraAllowAboveBuildLimit = new Setting<>(true);
+
+    /**
+     * Minimum distance in blocks of an elytra trip before the pathfinder will try to fly above build limit. (Minimum: 32). Requires {@link #elytraAllowAboveBuildLimit} to be enabled.
+     */
+    public final Setting<Integer> elytraLongDistanceThreshold = new Setting<>(500);
+
+    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1544,6 +1544,17 @@ public final class Settings {
     public final Setting<Boolean> elytraChatSpam = new Setting<>(false);
 
     /**
+     * May reduce memory usage by using a custom allocator for pathfinding
+     */
+    public final Setting<Boolean> elytraCustomAllocator = new Setting<>(false);
+
+    /**
+     * Allow the pathfinder to attempt flight in tighter spaces, useful in caves but can be dangerous.
+     */
+    public final Setting<Boolean> elytraAllowTightSpaces = new Setting<>(false);
+
+
+    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1546,7 +1546,7 @@ public final class Settings {
     /**
      * May reduce memory usage by using a custom allocator for pathfinding
      */
-    public final Setting<Boolean> elytraCustomAllocator = new Setting<>(false);
+    public final Setting<Boolean> elytraCustomAllocator = new Setting<>(true);
 
     /**
      * Allow the pathfinder to attempt flight in tighter spaces, useful in caves but can be dangerous.
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1559,6 +1559,11 @@ public final class Settings {
     public final Setting<Boolean> elytraAllowAboveRoof = new Setting<>(false);
 
     /**
+     * Allow the pathfinder to access the baritone cache to improve pathing
+     */
+    public final Setting<Boolean> elytraUseBaritoneCache = new Setting<>(true);
+
+    /**
      * A map of lowercase setting field names to their respective setting
      */
     public final Map<String, Setting<?>> byLowerName;
--- a/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
+++ b/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
@@ -47,7 +47,7 @@ public final class BlockStateOctreeInterface {
         if (this.chunkPtr == 0 | ((chunkX ^ this.prevChunkX) | (chunkZ ^ this.prevChunkZ)) != 0) {
             this.prevChunkX = chunkX;
             this.prevChunkZ = chunkZ;
-            this.chunkPtr = NetherPathfinder.getOrCreateChunk(this.contextPtr, chunkX, chunkZ);
+            this.chunkPtr = NetherPathfinder.getChunkOrDefault(this.contextPtr, chunkX, chunkZ, true);
         }
         return Octree.getBlock(this.chunkPtr, x & 0xF, y, z & 0xF);
     }
--- a/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
+++ b/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
@@ -28,21 +28,21 @@ public final class BlockStateOctreeInterface {
 
     private final NetherPathfinderContext context;
     private final long contextPtr;
-    private final DimensionType dimType;
+    private final int minY;
     transient long chunkPtr;
 
     // Guarantee that the first lookup will fetch the context by setting MAX_VALUE
     private int prevChunkX = Integer.MAX_VALUE;
     private int prevChunkZ = Integer.MAX_VALUE;
 
-    public BlockStateOctreeInterface(final NetherPathfinderContext context, final DimensionType dimType) {
+    public BlockStateOctreeInterface(final NetherPathfinderContext context) {
         this.context = context;
         this.contextPtr = context.context;
-        this.dimType = dimType;
+        this.minY = context.minY;
     }
 
     public boolean get0(final int x, final int y, final int z) {
-        final int adjustedY = y - dimType.minY();
+        final int adjustedY = y - this.minY;
         if (adjustedY < 0 || adjustedY > 383) {
             return false;
         }
--- a/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
+++ b/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
@@ -19,6 +19,7 @@ package baritone.process.elytra;
 
 import dev.babbaj.pathfinder.NetherPathfinder;
 import dev.babbaj.pathfinder.Octree;
+import net.minecraft.world.level.dimension.DimensionType;
 
 /**
  * @author Brady
@@ -27,19 +28,22 @@ public final class BlockStateOctreeInterface {
 
     private final NetherPathfinderContext context;
     private final long contextPtr;
+    private final DimensionType dimType;
     transient long chunkPtr;
 
     // Guarantee that the first lookup will fetch the context by setting MAX_VALUE
     private int prevChunkX = Integer.MAX_VALUE;
     private int prevChunkZ = Integer.MAX_VALUE;
 
-    public BlockStateOctreeInterface(final NetherPathfinderContext context) {
+    public BlockStateOctreeInterface(final NetherPathfinderContext context, final DimensionType dimType) {
         this.context = context;
         this.contextPtr = context.context;
+        this.dimType = dimType;
     }
 
     public boolean get0(final int x, final int y, final int z) {
-        if (y < 0 || y > 383) {
+        final int adjustedY = y - dimType.minY();
+        if (adjustedY < 0 || adjustedY > 383) {
             return false;
         }
         final int chunkX = x >> 4;
@@ -49,6 +53,6 @@ public final class BlockStateOctreeInterface {
             this.prevChunkZ = chunkZ;
             this.chunkPtr = NetherPathfinder.getChunkOrDefault(this.contextPtr, chunkX, chunkZ, true);
         }
-        return Octree.getBlock(this.chunkPtr, x & 0xF, y, z & 0xF);
+        return Octree.getBlock(this.chunkPtr, x & 0xF, adjustedY, z & 0xF);
     }
 }
--- a/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
+++ b/src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
@@ -39,7 +39,7 @@ public final class BlockStateOctreeInterface {
     }
 
     public boolean get0(final int x, final int y, final int z) {
-        if ((y | (127 - y)) < 0) {
+        if (y < 0 || y > 383) {
             return false;
         }
         final int chunkX = x >> 4;
@@ -49,6 +49,6 @@ public final class BlockStateOctreeInterface {
             this.prevChunkZ = chunkZ;
             this.chunkPtr = NetherPathfinder.getOrCreateChunk(this.contextPtr, chunkX, chunkZ);
         }
-        return Octree.getBlock(this.chunkPtr, x & 0xF, y & 0x7F, z & 0xF);
+        return Octree.getBlock(this.chunkPtr, x & 0xF, y, z & 0xF);
     }
 }
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -152,7 +152,6 @@ public final class ElytraBehavior implements Helper {
             this.clear();
         }
 
-        // requires read lock to be held
         public void tick() {
             // Recalculate closest path node
             this.updatePlayerNear();
@@ -168,8 +167,13 @@ public final class ElytraBehavior implements Helper {
             int minY = ctx.world().dimensionType().minY();
             int y = ctx.playerFeet().y;
             if (y >= minY && y < minY + context.maxHeight) {
-                // Obstacles are more important than an incomplete path, handle those first.
-                this.pathfindAroundObstacles();
+                context.readLock.lock();
+                try {
+                    // Obstacles are more important than an incomplete path, handle those first.
+                    this.pathfindAroundObstacles();
+                } finally {
+                    context.readLock.unlock();
+                }
                 this.attemptNextSegment();
             }
         }
@@ -306,6 +310,7 @@ public final class ElytraBehavior implements Helper {
                     .thenAcceptAsync(this::setPath, ctx.minecraft()::execute);
         }
 
+        // required read lock to be held
         private void pathfindAroundObstacles() {
             if (this.recalculating) {
                 return;
@@ -518,12 +523,7 @@ public final class ElytraBehavior implements Helper {
     }
 
     public void onTick() {
-        this.context.readLock.lock();
-        try {
-            this.onTick0();
-        } finally {
-            this.context.readLock.unlock();
-        }
+        this.onTick0();
         final long now = System.currentTimeMillis();
         if ((now - this.timeLastCacheCull) / 1000 > Baritone.settings().elytraTimeBetweenCacheCullSecs.value) {
             this.context.queueCacheCulling(ctx.player().chunkPosition().x, ctx.player().chunkPosition().z, Baritone.settings().elytraCacheCullDistance.value, this.boi);
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -660,6 +660,7 @@ public final class ElytraBehavior implements Helper {
         }
     }
 
+    // calls passable which requires a read lock
     private Solution solveAngles(final SolverContext context) {
         final NetherPath path = context.path;
         final int playerNear = landingMode ? path.size() - 1 : context.playerNear;
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -530,7 +530,7 @@ public final class ElytraBehavior implements Helper {
         this.pendingSolution = null;
         if (this.solver != null) {
             try {
-                this.pendingSolution = this.solver.get();
+                this.pendingSolution = this.solver.get(50 * 5, TimeUnit.MILLISECONDS);
             } catch (Exception ignored) {
                 // it doesn't matter if get() fails since the solution can just be recalculated synchronously
             } finally {
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -164,7 +164,9 @@ public final class ElytraBehavior implements Helper {
                 this.ticksNearUnchanged = 0;
             }
 
-            if (ctx.player().position().y < context.maxHeight && ctx.player().position().y > 0) {
+            int minY = ctx.world().dimensionType().minY();
+            int y = ctx.playerFeet().y;
+            if (y >= minY && y < minY + context.maxHeight) {
                 // Obstacles are more important than an incomplete path, handle those first.
                 this.pathfindAroundObstacles();
                 this.attemptNextSegment();
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -22,6 +22,8 @@ import baritone.api.Settings;
 import baritone.api.behavior.look.IAimProcessor;
 import baritone.api.behavior.look.ITickableAimProcessor;
 import baritone.api.event.events.*;
+import baritone.api.pathing.elytra.IElytraPathfinderContext;
+import baritone.api.pathing.elytra.UnpackedSegment;
 import baritone.api.pathing.goals.GoalBlock;
 import baritone.api.utils.*;
 import baritone.api.utils.input.Input;
@@ -75,7 +77,7 @@ public final class ElytraBehavior implements Helper {
     private List<BetterBlockPos> visiblePath;
 
     // :sunglasses:
-    public final NetherPathfinderContext context;
+    public final IElytraPathfinderContext context;
     public final PathManager pathManager;
     private final ElytraProcess process;
 
@@ -102,7 +104,6 @@ public final class ElytraBehavior implements Helper {
     private final int[] nextTickBoostCounter;
 
     private BlockStateInterface bsi;
-    private final BlockStateOctreeInterface boi;
     public final BetterBlockPos destination;
     private final boolean appendDestination;
 
@@ -134,7 +135,6 @@ public final class ElytraBehavior implements Helper {
                 Baritone.settings().elytraUseCache.value ? baritone.getWorldProvider().getCurrentWorld().directory.resolve("cache") : null,
                 ctx.world()
         );
-        this.boi = new BlockStateOctreeInterface(context, ctx.world().dimensionType());
     }
 
     public final class PathManager {
@@ -166,7 +166,7 @@ public final class ElytraBehavior implements Helper {
 
             int minY = ctx.world().dimensionType().minY();
             int y = ctx.playerFeet().y;
-            if (y >= minY && y < minY + context.maxHeight) {
+            if (y >= minY && y < minY + context.getMaxHeight()) {
                 // Obstacles are more important than an incomplete path, handle those first.
                 this.pathfindAroundObstacles();
                 this.attemptNextSegment();
@@ -457,7 +457,7 @@ public final class ElytraBehavior implements Helper {
     public void onChunkEvent(ChunkEvent event) {
         if (event.isPostPopulate() && this.context != null) {
             final LevelChunk chunk = ctx.world().getChunk(event.getX(), event.getZ());
-            this.context.queueForPacking(chunk, boi);
+            this.context.queueForPacking(chunk);
         }
     }
 
@@ -510,22 +510,22 @@ public final class ElytraBehavior implements Helper {
                 LevelChunk chunk = chunkProvider.getChunk(x, z, false);
 
                 if (chunk != null && !chunk.isEmpty()) {
-                    this.context.queueForPacking(chunk, boi);
+                    this.context.queueForPacking(chunk);
                 }
             }
         }
     }
 
     public void onTick() {
-        this.context.readLock.lock();
+        this.context.RLock();
         try {
             this.onTick0();
         } finally {
-            this.context.readLock.unlock();
+            this.context.RUnlock();
         }
         final long now = System.currentTimeMillis();
         if ((now - this.timeLastCacheCull) / 1000 > Baritone.settings().elytraTimeBetweenCacheCullSecs.value) {
-            this.context.queueCacheCulling(ctx.player().chunkPosition().x, ctx.player().chunkPosition().z, Baritone.settings().elytraCacheCullDistance.value, this.boi);
+            this.context.queueCacheCulling(ctx.player().chunkPosition().x, ctx.player().chunkPosition().z, Baritone.settings().elytraCacheCullDistance.value);
             this.timeLastCacheCull = now;
         }
     }
@@ -649,11 +649,11 @@ public final class ElytraBehavior implements Helper {
 
             final SolverContext context = this.new SolverContext(true);
             this.solver = this.solverExecutor.submit(() -> {
-                this.context.readLock.lock();
+                this.context.RLock();
                 try {
                     return this.solveAngles(context);
                 } finally {
-                   this.context.readLock.unlock();
+                   this.context.RUnlock();
                 }
             });
             this.solveNextTick = false;
@@ -1017,7 +1017,7 @@ public final class ElytraBehavior implements Helper {
             return clear;
         }
 
-        return this.context.raytrace(8, src, dst, NetherPathfinderContext.Visibility.ALL);
+        return this.context.raytrace(8, src, dst, IElytraPathfinderContext.Visibility.ALL);
     }
 
     public boolean clearView(Vec3 start, Vec3 dest, boolean ignoreLava) {
@@ -1286,7 +1286,7 @@ public final class ElytraBehavior implements Helper {
             final Material mat = this.bsi.get0(x, y, z).getMaterial();
             return mat == Material.AIR || mat == Material.LAVA;
         } else {
-            return !this.boi.get0(x, y, z);
+            return this.context.passable(x, y, z);
         }
     }
 
--- a/src/main/java/baritone/process/elytra/ElytraBehavior.java
+++ b/src/main/java/baritone/process/elytra/ElytraBehavior.java
@@ -45,6 +45,7 @@ import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.ClipContext;
+import net.minecraft.world.level.Level;
 import net.minecraft.world.level.chunk.ChunkSource;
 import net.minecraft.world.level.chunk.LevelChunk;
 import net.minecraft.world.level.material.Material;
@@ -174,7 +175,7 @@ public final class ElytraBehavior implements Helper {
 
         public CompletableFuture<Void> pathToDestination(final BlockPos from) {
             final long start = System.nanoTime();
-            return this.path0(from, ElytraBehavior.this.destination, UnaryOperator.identity())
+            return this.path0(from, destinationFixed(), UnaryOperator.identity())
                     .thenRun(() -> {
                         final double distance = this.path.get(0).distanceTo(this.path.get(this.path.size() - 1));
                         if (this.completePath) {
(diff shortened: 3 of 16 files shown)


thelampgod (approved):
wow!

underscore-zi:
I do have a couple minor commits over on https://github.com/underscore-zi/baritone 

First is respecting the allowRoof setting when searching for a landing spot. I had left a TODO in for this but hadn't done so yet.

Second is adding Cobblestone and Obsidian to allowed landing blocks. In vanilla the current list is fine but I was using this around 2b's spawn and it had some issues finding landing spots.

ZacSharp:
Can you please wait three days with merging? I'd like to properly read through and test this, but won't be able to do so before April 8.

babbaj:
leijurv never merges my PRs anyways

leijurv:
> Can you please wait three days with merging?

say no more, i'll check back next month

src/api/java/baritone/api/Settings.java
```diff
@@ -1543,6 +1543,26 @@ public final class Settings {
...
+
+    /**
+     * Allow the pathfinder to fly above y 128 in the nether.
+     */
+    public final Setting<Boolean> elytraAllowAboveRoof = new Setting<>(false);
+
+    /**
+     * Allow the pathfinder to access the baritone cache to improve pathing
+     */
+    public final Setting<Boolean> elytraUseBaritoneCache = new Setting<>(true);
```
ZacSharp: This setting name / documentation is odd. From a user perspective it's all Baritone.

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
```diff
@@ -39,16 +39,16 @@ public BlockStateOctreeInterface(final NetherPathfinderContext context) {
     }
 
     public boolean get0(final int x, final int y, final int z) {
-        if ((y | (127 - y)) < 0) {
+        if (y < 0 || y > 383) {
```
ZacSharp: This always checks for the maximum chunk height, is that correct? i.e. does nether-pathfinder always allocate full-size chunks?

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -24,98 +24,153 @@
...
-    public NetherPathfinderContext(long seed) {
-        this.context = NetherPathfinder.newContext(seed);
+    public NetherPathfinderContext(long seed, Path cache, Level world) {
+        this.dimension = world.dimension();
+        final int dim;
+        if (this.dimension == Level.NETHER) dim = NetherPathfinder.DIMENSION_NETHER;
+        else if (this.dimension == Level.END) dim = NetherPathfinder.DIMENSION_END;
+        else dim = NetherPathfinder.DIMENSION_END;
+        int height = Math.min(world.dimensionType().height(), 384);
+        if (!Baritone.settings().elytraAllowAboveRoof.value && dim == NetherPathfinder.DIMENSION_NETHER) height = Math.min(height, 128);
```
ZacSharp: Ouch. Can you please autoformat your code with whatever formatting the rest of Baritone uses?

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -24,98 +24,153 @@
...
+        this.context = NetherPathfinder.newContext(seed, cache.toString(), dim, height, Baritone.settings().elytraCustomAllocator.value);
```
ZacSharp: If `elytraUseBaritoneCache` is disabled then `cache` is null

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -24,98 +24,153 @@
...
             ChunkPos chunkPos = event.getChunkPos();
-            long ptr = NetherPathfinder.getChunkPointer(this.context, chunkPos.x, chunkPos.z);
-            if (ptr == 0) return; // this shouldn't ever happen
-            event.getBlocks().forEach(pair -> {
-                BlockPos pos = pair.first();
-                if (pos.getY() >= 128) return;
-                boolean isSolid = pair.second() != AIR_BLOCK_STATE;
-                Octree.setBlock(ptr, pos.getX() & 15, pos.getY(), pos.getZ() & 15, isSolid);
-            });
+            // not inserting or deleting from the cache hashmap so we should be fine not being exclusive
```
ZacSharp: Except iirc C++ considers any data race undefined behavior and doing this across an ffi boundary won't make things better.

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -24,98 +24,153 @@
...
+        this.writeExecutor.execute(() -> {
             // TODO: Prioritize packing recent chunks and/or ones that the path goes through,
             //       and prune the oldest chunks per chunkPackerQueueMaxSize
             final LevelChunk chunk = ref.get();
             if (chunk != null) {
-                long ptr = NetherPathfinder.getOrCreateChunk(this.context, chunk.getPos().x, chunk.getPos().z);
-                writeChunkData(chunk, ptr);
+                writeLock.lock();
+                try {
+                    long ptr = NetherPathfinder.allocateAndInsertChunk(this.context, chunk.getPos().x, chunk.getPos().z);
```
ZacSharp, line 129: I assume this always returns a completely zeroed chunk?

ZacSharp (changes requested):
Ok, didn't expect quite as much breakage, but this adds at least two ways to crash the game.
* Disabling `elytraUseBaritoneCache` unconditionally causes an NPE, crashing the game unless this happens inside a command
* Disabling `elytraAllowAboveRoof` while flying above the nether roof and then causing recalculation crashes the game. I didn't even get a crash report, but no segfault either?
* Enabling `elytraPredictTerrain` in the overworld leads to endless recalculation and chat spam. Actually not a crash though.

Some less severe points
* The wall of text printed by `#elytra` unless `elytraTermsAccepted` is now outdated. It mentions elytra flight only being for the nether (now wrong) and also recommends `elytraAutoJump`, which has hardly any chance of working in the overworld, and should also mention that terrain prediction is nether-only.
* The path is automatically recalculated when `elytraNetherSeed` changes, `elytraAllowThighsSpaces` does not have this effect. Not sure whether automatically recalculating is better.

I also really do not like how we are offsetting positions back and forth all across the code. Just look at how many times this went wrong with the existing chunk cache.
Can't we just use exactly one coordinate type (absolute world coordinates) in Baritone and shift to 0 based at the API boundary?
You didn't want to do this in nether-pathfinder itself, but maybe we can wrap the java side of the API to do the shifting?

babbaj:
> Disabling elytraUseBaritoneCache unconditionally causes an NPE, crashing the game unless this happens inside a command

fixed

> * Disabling `elytraAllowAboveRoof` while flying above the nether roof and then causing recalculation crashes the game. I didn't even get a crash report, but no segfault either?

I don't think this description is accurate. If elytraAllowAboveRoof is off and we put ourself above the roof then there is a crash (nether-pathfinder printing "retard" to stderr and calling std::terminate) but changing that setting while flying has no effect except possibly when it finds a landing spot.

I have not yet fixed this but I will probably fix it by pausing pathfinding until the player goes back below the roof.

> Enabling elytraPredictTerrain in the overworld leads to endless recalculation and chat spam. Actually not a crash though.

fixed

> The path is automatically recalculated when elytraNetherSeed changes, elytraAllowThighsSpaces does not have this effect. Not sure whether automatically recalculating is better.

fixed to behave the same

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
babbaj: yes

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
babbaj: using properly formatted braces would make it take up more lines for no benefit

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
babbaj: It's undefined behavior if another thread is writing to the data at the same time we are reading it but that is not the case here.

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
babbaj: Yes

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
ZacSharp: We are the writing thread here, but since we only have the read lock another thread may do the reading required to trigger undefined behavior.

ZacSharp: In case you forgot / didn't see it, there's an `Octree.setBlock` call in this block. The hashmap reading is hopefully fine, as the comment says.

babbaj: The pathfinder racing with this would be inconsequential however multiple calls to this to nearby blocks could cause bad data to be written so i will change it to use a write lock. The read executor is still single threaded so there currently isn't a difference anyways.

ZacSharp:
You want to fix the out-of-bounds crash before merging this or later?

babbaj:
before. i also found another bug last night that's more important.

underscore-zi:
> I also really do not like how we are offsetting positions back and forth all across the code. Just look at how many times this went wrong with the existing chunk cache.
> Can't we just use exactly one coordinate type (absolute world coordinates) in Baritone and shift to 0 based at the API boundary?

I've added a commit on my branch (https://github.com/underscore-zi/baritone/commit/d0c3aeac783d8d97c859553849d1da8c888900db) that pulls all those adjustments into the  `NetherPathfinderContext` methods that make the pathfinder calls along with one Octree usage in `BlockStateOctreeInterface`.

On a related note, is there a better way for me to share commits? Or would you just prefer I left it all over to Babbaj now since its not my PR?

babbaj:
Just fixed the rest of the known bugs in https://github.com/cabaletta/baritone/pull/4698/commits/75096e1d352d0cc208d1f1be471ecf8f0e82259d

> I've added a commit on my branch (https://github.com/underscore-zi/baritone/commit/d0c3aeac783d8d97c859553849d1da8c888900db)

merged

src/main/java/baritone/process/elytra/ElytraBehavior.java
```diff
@@ -159,9 +164,11 @@ public void tick() {
                 this.ticksNearUnchanged = 0;
             }
 
-            // Obstacles are more important than an incomplete path, handle those first.
-            this.pathfindAroundObstacles();
-            this.attemptNextSegment();
+            if (ctx.player().position().y < context.maxHeight && ctx.player().position().y > 0) {
```
ZacSharp: This looks like comparing world coordinates against a range of zero based coordinates?

babbaj: fixed

ZacSharp:
Still get a crash by these steps
1. Stand on the nether roof with some firework rockets and an elytra
2. Run `#settings reset all`, `#elytraAllowAboveRoof true`, `#thisway 1000`, `#elytra`
3. Start flying
4. Run `#elytraAllowAboveRoof false`, `#elytra reset`

src/main/java/baritone/process/elytra/ElytraBehavior.java
```diff
@@ -1262,6 +1280,7 @@ private static Vec3 step(final Vec3 motion, final Vec3 lookDirection, final floa
         return new Vec3(motionX, motionY, motionZ);
     }
 
+    // any call to this must be done with the lock held
```
ZacSharp, line 1287: Can you maybe apply this comment to all methods which require synchronization and specify that the readlock is required?

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
```diff
@@ -27,28 +28,31 @@ public final class BlockStateOctreeInterface {
...
 
     // Guarantee that the first lookup will fetch the context by setting MAX_VALUE
     private int prevChunkX = Integer.MAX_VALUE;
     private int prevChunkZ = Integer.MAX_VALUE;
 
-    public BlockStateOctreeInterface(final NetherPathfinderContext context) {
+    public BlockStateOctreeInterface(final NetherPathfinderContext context, final DimensionType dimType) {
         this.context = context;
         this.contextPtr = context.context;
+        this.dimType = dimType;
```
ZacSharp: Could potentially take the dimension info from `context`

ZacSharp:
> On a related note, is there a better way for me to share commits? Or would you just prefer I left it all over to Babbaj now since its not my PR?

You could open pull requests against babbaj's pull request branch, but I'm not sure whether that's better than a writing a comment.

src/main/java/baritone/process/elytra/ElytraBehavior.java
babbaj: This is the only function run on the main thread that requires synchronization but i added a comment to solveAngles.

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
babbaj: NetherPathfinderContext just has the id of the dimension which the min y can't directly be taken from

babbaj:
> Still get a crash by these steps

fixed

src/main/java/baritone/process/elytra/BlockStateOctreeInterface.java
ZacSharp: It does have the min y here https://github.com/cabaletta/baritone/blob/04813a8312550ba5c18a0611b175f5062d21db46/src/main/java/baritone/process/elytra/NetherPathfinderContext.java#L83

ZacSharp:
> > Still get a crash by these steps
> 
> fixed

Replace `#elytra reset` with changing `elytraNetherSeed`/`elytraPredictTerrain`/`elytraAllowTightSpaces` and it crashes once again.

src/main/java/baritone/process/elytra/ElytraBehavior.java
ZacSharp: As far as I can tell we have
```
passable called by simulate and pathFindAroundObstacles
simulate called by solvePitch called by the other solvePitch called by solveAngles
solveAngles called by onPostTick with synchronisation and by tick (and tick is public)
tick called by ElytraProcess.onTick with synchronization on the main thread
pathFindAroundObstacles called by PathManager.tick called by onTick0 called by onTick with synchronization on the main thread
```
At least `tick` needing synchronization should be documented.

babbaj:
fixed again

ZacSharp (approved):
I'm out of crashes now. The only thing I managed to find is an "Unhandled exception" error message when you change to an invalid goal mid-flight, but that's not new. (e.g. `#goal ~`).

leijurv:
I will merge this..... on one condition........

When you #elytra to a goal in the overworld that is at least some distance away (configurable setting, defaults to, let's say 1000 blocks), it should fly upwards (if it cannot fly upwards it'll go towards the goal of course). Once it gets above y=320, it will have a single target point that is simply the goal (ignoring the Y of the goal). I think this will work fine in the code because raytraces in minecraft have a maximum number of blocks they visit so it won't be too laggy? If it is laggy just add partial points at fractions along the way. Then once it is near the goal it dips back down below 320 and does landing or whatever at the goal

I think this would be a compromise that would allow it to fly through interesting overworld terrain like 2b2t spawn, while still going straight any-angle diagonal towards the goal (rather than the straight + 45° thing that I just think is low quality).

babbaj:
It is currently not possible to fly above y 320 without increasing the height in nether-pathfinder even more. That should be an easy change.

pizzav-xyz:
This is amazing

leijurv:
> It is currently not possible to fly above y 320 without increasing the height in nether-pathfinder even more. That should be an easy change.

No the whole point is to **not** use nether-pathfinder above the ceiling. When above y=320, you point in a straight line to the goal, without **any** pathfinding. Don't increase the height of nether-pathfinder's chunks

Have nether-pathfinder generate a path to y=320, then on the java side, add a path point after that that's at like same x,z but y=350 or something. Then the next path point is the goal x,z at y=350.

Might need to edit nether-pathfinder to support GoalYLevel

leijurv:
(and it should only do this if the destination is at least, e.g., 1000 blocks away. configurable setting)

underscore-zi:
> Have nether-pathfinder generate a path to y=320, then on the java side, add a path point after that that's at like same x,z but y=350 or something. Then the next path point is the goal x,z at y=350.

I don't have much time to work on this at the moment, but I originally used an implementation kinda like that and was working on a less hacked together implementation more recently but haven't finished. So I'm just kinda dumping what I had for my future-self or if someone else gives it a stab:

To allow it to follow a path that goes outside of the pathfinders limits `ElytraBehavior:clearView` has to be patched to use `ctx.world().clip(...)` instead of the native pathfinder when above the y-limit. And `ElytraBehavior:passable` to use `this.bsi` like the ignoreLava path instead of `this.boi`. 

Then for actually generating the path without pathfinder support for GoalY, I took advantage of the fact that it would attempt to pathFind again when near the end of an incomplete(finished=false) path.

When pathFinding it would check the source. If its above or near 320 and above the height map of the area it would just generate point path. If going far enough it would attempt to pathfind to 75%[0] the minimum distance with y=320, or it would just pathfind to the destination.

This allowed it to fly under a roof by just repeatedly creating incomplete paths going 75% the minimum distance and eventually switch to the point path whenever the route became clear.

Also its worth noting that the point path did need to be more detailed than just the destination point as it didn't like to fly if the next point wasn't loaded in. Though that might be fixable by patching `clearView` to be true when at elevation instead of using `clip(...)`. 

[0] 75% wasn't chosen for any particular reason, just giving some breathing room

CREAsTIVE:
any updates?

lawnguy1201:
What has happened with this?

pizzav-xyz:
> What has happened with this? 
> 

A random guy has commented on this 3 months later

lawnguy1201:
> > What has happened with this?
> 
> A random guy has commented on this 3 months later

no shit

pizzav-xyz:
> > > What has happened with this?
> > 
> > A random guy has commented on this 3 months later
> 
> no shit

So why are you asking if you already know

underscore-zi:
Still actively working on it.

 I have something that generally works and goes above the build limit. Though I've been working out some bugs where it seems to get stuck momentarily as if it can't see the next node (despite clear raytraces to it).

Also some issues with generated the transitions between the different path generators since it usually tries to transition right at the edge of loaded chunks when it tries to pathfind right when a new chunk gets loaded.

lawnguy1201:
> Still actively working on it.
> 
> I have something that generally works and goes above the build limit. Though I've been working out some bugs where it seems to get stuck momentarily as if it can't see the next node (despite clear raytraces to it).
> 
> Also some issues with generated the transitions between the different path generators since it usually tries to transition right at the edge of loaded chunks when it tries to pathfind right when a new chunk gets loaded.

Nice, would there be anyway I could get the .jar. working on a project and it need B to fly in the OW.

underscore-zi:
I'm trying to troubleshoot some behavior and wondering if any of you all might be able to point me to where in code this behavior might be coming from.

[2025-08-22-113218482-2c80a.webm](https://github.com/user-attachments/assets/a48c0508-7bb9-4cd6-8c59-e18cd3ec2c15)

Basically, as the Elytra crosses the build limit it starts to have trouble following the path. It'll follow the path, then seem to get stuck and start spinning. This is similar to the behavior that happens when about to enter an unloaded chunk though it is loaded in this case. 

I suspect something sees it as unloaded (`ctx.world().isLoaded` returns false for y > build limit, but I've patched that already in `ElytraBehavior`). Though I'm not sure where this spin behavior comes from exactly. And if I reduce the flight level to be under the build limit, it works without issue. So I'm thinking there is some check buried away I've missed besides the one in `attemptNextSegment`

Edit: and for what its worth raytraces all come up green.

pizzav-xyz:
> I'm trying to troubleshoot some behavior and wondering if any of you all might be able to point me to where in code this behavior might be coming from.
> 2025-08-22-113218482-2c80a.webm
> 
> Basically, as the Elytra crosses the build limit it starts to have trouble following the path. It'll follow the path, then seem to get stuck and start spinning. This is similar to the behavior that happens when about to enter an unloaded chunk though it is loaded in this case.
> 
> I suspect something sees it as unloaded (`ctx.world().isLoaded` returns false for y > build limit, but I've patched that already in `ElytraBehavior`). Though I'm not sure where this spin behavior comes from exactly. And if I reduce the flight level to be under the build limit, it works without issue. So I'm thinking there is some check buried away I've missed besides the one in `attemptNextSegment`
> 
> Edit: and for what its worth raytraces all come up green.

are you sure the chunks arent actually unloaded

underscore-zi:
> are you sure the chunks arent actually unloaded

Yes

ZacSharp:
Having a short look at the solver it looks like `isHitboxClear` might also need patching to not use `context.raytrace`. Not sure why `passable` needs to use `bsi` instead of `boi` since `boi.get` returns `false`/air for out of bounds positions, which is what we need here.

leijurv:
yeah looks like something messed up with the raytracer. out of bounds (above the ceiling) should return air

underscore-zi:
Thanks! Fixed it. Was able to trace the bug down from `isHitBoxClear`.

It was indeed a raytracing bug, miscalculated when a target was in/out of bounds resulting in misusing the NetherPathfinder for OOB raytraces.

nonokirby:
Is this pull stable enough to fork and run currently, or does it have major issues still? specifically looking to elytrafly in/around 2b's spawn if knowing terrain helps.

ZacSharp:
As far as I know there are no known (new) bugs, so should definitely be usable.

(The reason this isn't merged is a pending functional change)

underscore-zi:
I have some commits implementing flight above the build limit over on: https://github.com/underscore-zi/baritone 

I do make one major structural change which might raise some eyebrows. I've extended the API with `IElytraPathfinderContext` and `IElytraPathfinderContextFactory` 

So now, the NetherPathfinderContext basically already implemented the interface, and I've added a new SkyPathfinderContext which also implements the interface and wraps the NetherPathfinder to extend it adding support for flight above the build limit. ElytraBehavior uses the appropriate context for the dimension/settings, keeping the original performance if users want that.

In defense of having this in the API is does serve to implement a feature I've wanted, the ability to navigate between waypoints without doing the whole takeoff/fly/land process for every point or changing the destination when the flight gets close to each point. I made a nice like flight replayer for myself with this.

ZacSharp:
Tried to come up with a funky way to review this. Wouldn't have expected that two people working on one pr is such a big problem on a "collaboration platform".

`SkyPathfinderContext` uses the wrong naming convention for it's private helpers.

https://github.com/underscore-zi/baritone/compare/20c1631abfe8b64fa119e8012b348ebdd55a32a8...94a466be9018e87e3f55e9e5148cc32b4ed5e608#diff-aa0f9f28257c586978349ec09d0ec52c19858ebfc79ff63b56b498a69ca16d14R28
I don't like exposing nether-pathfinder API directly in our API. And `UnpackedSegmend` is imho a prime example of what's considered an implementation detail.

https://github.com/underscore-zi/baritone/compare/20c1631abfe8b64fa119e8012b348ebdd55a32a8...94a466be9018e87e3f55e9e5148cc32b4ed5e608#diff-f881174eae586d93b19f8eb3f9c26af571c0ca278f7d9a108cba4d66a71f519fR397-R398
Aren't there fields caching these two values?


https://github.com/underscore-zi/baritone/compare/20c1631abfe8b64fa119e8012b348ebdd55a32a8...94a466be9018e87e3f55e9e5148cc32b4ed5e608#diff-f881174eae586d93b19f8eb3f9c26af571c0ca278f7d9a108cba4d66a71f519fR443-R458
It looks weird to have both a special case when the field is null in the getter and a null-replacement in the setter.


https://github.com/underscore-zi/baritone/compare/20c1631abfe8b64fa119e8012b348ebdd55a32a8...94a466be9018e87e3f55e9e5148cc32b4ed5e608#diff-eebf5e50d8fe55d39379412b51279851bba3a6f3019584be8ea2bfd36ce71fe1R85
I like how this is no longer being passed around. It uses the exact same internal storage and is fundamentally tied to the context, so might as well be part of it.

<br>

Honestly I'm not happy with the two new interfaces. `IElytraPathfinderContextFactory` feels like it's mostly boilerplate and `IElytraPathfinderContext` is just bad API. It is what netherpathfinder expects and using it internally makes sense, but it's a messy implementation detail.
There's also a concern about deadlocks here, because giving API users control of the context lock means the thing can lock and unlock in ways we do not expect. The thread "safety" in this part of Baritone is a mess already.

Flying around without landing sounds very reasonable though, so maybe there's a way to implement that feature without putting implementation details into the API package?

<br>

A somewhat unrelated idea I had is that iirc the pathfinder uses a symmetric movement set so there is no difference between going forwards or backwards. That means if the goal is already in render distance and we are still above the world we could run the pathfinder from the goal towards the build limit / our current position to determine where to transition down. Not sure whether that helps with anything.

babbaj:
just merged his branch into mine. he was also a few commits behind so I had to merge and fix some conflicts.

underscore-zi:
> And `UnpackedSegmend` is imho a prime example of what's considered an implementation detail.

I feel like the `Stream<BetterBLockPos>` and boolean wrapper are what would be necessary to expose to allow someone else to provide a path. The name is a bit off though since the public side would have no knowledge of the "packed" segments. 

Perhaps a simpler option would be to only allow the API user to provide waypoints and leave the path to Baritone. That perhaps fits better with Baritone more generally and would probably be the main usecase. One use-case I had for full pathing was flying down 2b diagonal highways where there was some active wither griefing so I couldn't bounce.

> I don't like exposing nether-pathfinder API directly in our API.

Yeah, I wasn't the biggest fan of my own approach there though I mostly was just trying not to break anything outside of my own code. So isolated it via the abstraction. But yeah, I think I can clean that up quite a bit.

Raytracing stuff can be pulled back internal, no reason to have that overloadable if the base implementation supports OOB traces it handles all cases. Though it adds a bit of overhead as it needs to check which tracer to use. Chunk/block information can be obtained by any mod already so no point in providing that either; that also removes the locks.

Which would leave pathFindAsync, cancel and maybe destroy which feels much more sane. And I think that might remove the Factory too as there wouldn't be a need to provide the Baritone cache path into it.

underscore-zi:
Committed to my fork again: https://github.com/underscore-zi/baritone

I pulled everything new out of the API, though I still use a simpler interface internally so the original Nether code can be swapped in depending on settings.

Murat65536:
Fixing a few of the things Codacy is mad about might also be a good idea. Mostly naming things. I wouldn't worry about the NPath complexity.

underscore-zi:
I believe I've addressed all of the issues from Codacy that were still relevant. 👍

ZacSharp:
Looks better now, thanks.

Feeding a set of waypoints sounds good (honestly for the normal pathfinder as well, but that can be kept separate).

underscore-zi:
I was thinking to leave waypoints out of this PR and I'll make that a separate one for that.

I do have one bug to fix in this set though. The semaphore I introduced leads (I suspect it's the cause atleast) to a bunch of lag when it gets stuck and can't find a path forward and keeps attempting to recalculate.

ZacSharp:
> I was thinking to leave waypoints out of this PR and I'll make that a separate one for that.

:+1:

underscore-zi:
I pushed three more commits

1. First was just a bug when it no landing spot was found, instead of hovering ti would land on whatever block was there. Not sure exactly when/how that behavior changed but easily restored.
2. I thought the lag was due to the semaphore and some path adjustments leading to creation/destroying of NetherPathfinderContext with new pathing requests. So it pulls the context object out of the Behavior allowing it to persist on new pathTo request if the context already exists. Ultimately this commit might not be necessary as it wasn't the issue, but I didn't revert it as performance-wise it still makes sense to not need to destroy and recreate the chunk cache every time especially with use-cases like Jeff Mod's trail follower which frequently sets new pathing goals.
3. Actual issue was just that the PathCalculationException was getting wrapped as a RuntimeException which once propagated up led to a bunch of extra logging and handling and they would happen in rapid succession.

kaliko69:
Hey! I’m really excited about the Elytra auto-flight in the Overworld/End feature. I tried versions 1.21.4 and 1.21.8, but it seems like only Nether Elytra works there. Is there a way to get the Overworld/End Elytra version right now, or is it still work-in-progress? Thanks! ( i dont know who to mention @babbaj @underscore-zi  @ZacSharp @leijurv  @kittenvr )

underscore-zi:
@babbaj would you mind pulling in my commits? I believe its good to go now; though I'll address any new issues from codacy.

leijurv:
ignore codacy

ZacSharp:
@babbaj by GitHub permissions logic I have push access to this pr so if you don't mind I can also transfer zi's commits here.

dep4:
Can this be merged, pretty please?

PR #4706, opened 2025-04-13
Break blocks from near to far, place blocks from far to near. Support…

0xhtml:
… replaceable blocks. e.g. don't mine single layer snow before placing


ZacSharp:
This completely breaks breaking replaceable blocks.


https://github.com/user-attachments/assets/de9ecbd2-f5f3-464a-a822-5e4d6fd6ec74


Besides that, it would be nice if you split your changes into multiple commits. Changing the iteration order is mostly white space changes (plus the new fields), but since you combined it with moving lots of checks around it is hard to see what you did.

Also, since you combined `int` triples into `BetterBlockPos` wherever you came by one, I'd like to let you know that using `int`s in places which are executed a lot is done intentionally to avoid allocating and throwing away a thousand `BetterBlockPos` objects per tick. Obviously, allocating the same position multiple times is right against that goal.

PR #4720, opened 2025-04-21
Don't let servers trick users into running arbitrary Baritone commands

ZacSharp:
During a discussion on #4692 I ended up handcrafting books and chat messages with clickable text, which made me notice that servers can simply send `ChatComponent`s with Baritone commands and somehow get the user into clicking them to have Baritone execute the commands.
Since the server cannot know the `FORCE_COMMAND_PREFIX` this change prevents it from crafting such messages/books/signs.

I do not see how this could be used for more than deleting waypoints/settings and disconnecting the player, but this could change in future. For example I have a private branch which has a setting to specify the location of an external helper program. A single click on a sign is enough to let the server change that setting and run the command which uses it, allowing the server to run any program the user has installed.


ZacSharp:
That was quick

Issue #3949, opened 2023-05-16
Rename `allowJumpAt256` to `allowJumpAtWorldTop`


PoolloverNathan:
## Describe your suggestion
Since newer Minecraft versions have a Y range of [-64,320) instead of [0,256), this suggestion renames `allowJumpAt256` to `allowJumpAtWorldTop`.

## Settings
* Rename `allowJumpAt256` to `allowJumpAtWorldTop`

## Context
Clarifying the actual option, since newer Minecraft versions have a Y range of [-64,320) instead of [0,256).

## Final checklist
- [ ] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

ZacSharp:
This would have been a good idea when porting to 1.17, but randomly renaming a long standing setting doesn't feel like it's a good idea.

ZacSharp:
Done, it is `allowJumpAtBuildLimit` now.

Issue #4734, opened 2025-05-03
Baritone maximum and minimum build heights are incorrect for 1.18+


Murat65536:
## What do you need help with?
I am using the Baritone API to direct the player towards a `BlockPos`. The player is on an elevated surface about 5 blocks above the goal and 2 blocks away. I have `maxFallHeightNoWater` set to 10. However, Baritone seems to prefer breaking the blocks underneath it until it reaches the destination. Even when the player is only elevated about 2 blocks above the ground, Baritone still refuses to let the player fall. When I make the platform and target out of bedrock to prevent Baritone from breaking underneath itself, the bot just can't find a target. Any idea what I'm doing wrong?

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

ZacSharp:
<details><summary>You mean it doesn't find a path like this?</summary>

![Image](https://github.com/user-attachments/assets/c475ec7c-bfbf-4273-a0f4-a30879d0804f)

</details>

Murat65536:
Yes.

Murat65536:
Are there any settings changes I would need to make for this to work?

ZacSharp:
No, I only had two unrelated  changed settings.

Murat65536:
I think I figured it out but Baritone's behavior is still strange. I tried this in a superflat world with the preset "the void". There, I wasn't able to jump down and had to break the block under me. However, when I opened a default world and tried the same thing, I was able to jump down. I'm decently sure that I don't have any incorrect settings since I also switched to Windows and tried it there. I got the same result. Do you know why this is? it seems weird that Baritone can't jump in that preset.

Murat65536:
Ok, I once again figured it out. Baritone can't jump down when the y coordinate is negative. wtf

ZacSharp:
That absolutely makes sense. The world used to stop there and you don't want to jump into the void, do you? It's an oversight.

ZacSharp:
Here is the culprit

https://github.com/cabaletta/baritone/blob/d74420b2653e092b1936881d42aba2f7375159ee/src/main/java/baritone/pathing/movement/movements/MovementDescend.java#L150-L154

Murat65536:
Shouldn't the `allowJumpAt256` setting also be changed to 320?

Murat65536:
Here:
[...]

PR #4736, opened 2025-05-05
Update build limits

Murat65536:
Update minimum and maximum build limit to account for the 1.18 update.
In response to #4734.
Also remove `allowJumpAt256` since it's a redundancy (an alternative option would be to go with #3949).

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -181,9 +181,10 @@ public final class Settings {
     public final Setting<Boolean> assumeSafeWalk = new Setting<>(false);
 
     /**
-     * Does nothing since it's intended functionality is redundant
+     * If true, parkour is allowed to make jumps when standing on blocks at the maximum height, so player feet is y=256
+     * <p>
+     * Defaults to false because this fails on constantiam. Please let me know if this is ever disabled. Please.
      */
-    @Deprecated
     public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
 
     /**
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -181,6 +181,13 @@ public final class Settings {
     public final Setting<Boolean> assumeSafeWalk = new Setting<>(false);
 
     /**
+     * If true, parkour is allowed to make jumps when standing on blocks at the maximum height, so player feet is y=256
+     * <p>
+     * Defaults to false because this fails on constantiam. Please let me know if this is ever disabled. Please.
+     */
+    public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
+
+    /**
      * This should be monetized it's so good
      * <p>
      * Defaults to true, but only actually takes effect if allowParkour is also true
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -181,10 +181,9 @@ public final class Settings {
     public final Setting<Boolean> assumeSafeWalk = new Setting<>(false);
 
     /**
-     * If true, parkour is allowed to make jumps when standing on blocks at the maximum height, so player feet is y=256
-     * <p>
-     * Defaults to false because this fails on constantiam. Please let me know if this is ever disabled. Please.
+     * Does nothing since it's intended functionality is redundant
      */
+    @Deprecated
     public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
 
     /**
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -192,7 +192,7 @@ public final class Settings {
      */
     @Deprecated
     @JavaOnly
-    public final Settings<Boolean> allowJumpAt256 = new Setting<>(false);
+    public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
 
     /**
      * This should be monetized it's so good
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -185,7 +185,14 @@ public final class Settings {
      * <p>
      * Defaults to false because this fails on constantiam. Please let me know if this is ever disabled. Please.
      */
-    public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
+    public final Setting<Boolean> allowJumpAtBuildLimit = new Setting<>(false);
+
+    /**
+     * Just here so mods that use the API don't break. Does nothing.
+     */
+    @Deprecated
+    @JavaOnly
+    public final Settings<Boolean> allowJumpAt256 = new Setting<>(false);
 
     /**
      * This should be monetized it's so good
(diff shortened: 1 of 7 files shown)


ZacSharp:
Thanks for the pull request!

Since the world height is configurable, this should be the actual world height instead of hardcoding the default. This should be doable using `ctx.world().dimensionType().minY()` + `ctx.world().dimensionType().height()`, though maybe there's a more direct way to get the maxY.

ZacSharp (approved):

Murat65536:
Looks like you can just do `ctx.world().getMinBuildHeight()` and `ctx.world().getHeight()`. Although, there is also just `ctx.world().getMaxBuildHeight()`.

src/main/java/baritone/pathing/movement/movements/MovementDescend.java
```diff
@@ -147,9 +147,9 @@ public static boolean dynamicFallCost(CalculationContext context, int x, int y,
         int effectiveStartHeight = y;
         for (int fallHeight = 3; true; fallHeight++) {
             int newY = y - fallHeight;
-            if (newY < 0) {
+            if (newY < context.getBaritone().getPlayerContext().world().dimensionType().minY()) {
```
ZacSharp: Better use `context.world.getMinBuildHeight()`.
There is kind of a split between things that happen on the pathing thread (`CalculationContext`) and things that happen on the main thread (`IPlayerContext`) so using the `IPlayerContext` on the pathing thread is not nice.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -64,10 +64,6 @@ public static void cost(CalculationContext context, int x, int y, int z, Directi
         if (!context.allowParkour) {
             return;
         }
-        if (y == 256 && !context.allowJumpAt256) {
```
wagyourtail, line 67: I would recommend not actually removing this setting, but making it check against the height limit.
I guess that's #3949, but could also just *not* change the setting name while updating the behavior

Murat65536: Not changing the name would make the setting completely inaccurate. It would be checking y=320 while implying that its checking y=256. Also, have there been any issues with jumping at the height limit?

Murat65536: It looks like there really is no reason that the setting should be there according to @ZacSharp 
https://github.com/cabaletta/baritone/issues/4734#issuecomment-2850868590

ZacSharp: Renaming and removing break compatibility in the same way, so if we don't remove it we should not rename it either.
But to be honest I do not think there would be a lot of breakage, hence my proposal to remove it. Could also keep the field as `@Deprecated` and remove all related logic.

Murat65536: If people have this in their settings.txt, Baritone should just print a warning to chat and move on. Removing this setting shouldn't break anything to my knowledge. Maybe I'm wrong about this?

ZacSharp: Baritone is completely fine with removing its own settings, we could even have it migrate renamed settings. The problem is that users can have macros to change settings (unlikely in this case?) and the field itself is part of Baritones public java API, so removing it can crash the game if a mod tries to access it anyway.
I've already checked Meteor and they would gracefully handle this specific removal, but other mods may use a more straightforward way to access settings and break.

Murat65536: Thanks for the explanation. Removed functionality of `allowJumpAt256` but kept the setting with a `@Deprecated` annotation. We could just come back to this some time in the future and remove it entirely.

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -64,10 +64,6 @@ public static void cost(CalculationContext context, int x, int y, int z, Directi
...
-            return;
```
leijurv, line 68: I can't even remember what server it was, but, this feature was added because some anticheats don't let you jump up above the roof of the world or something?

Rename to allowJumpAboveBuildLimit or some such?

Murat65536: Sure, but are you fine with the setting being renamed breaking the API? I mean, it's highly unlikely that anyone would be changing **this** of all settings in the API and a quick glance at GitHub Code Search doesn't show anything, but it is still possible I guess.

leijurv: Yeah I don't care about breaking the API really

leijurv: can you swap this? to `if (!context.allowJumpAtBuildLimit && y >= context.world.getMaxBuildHeight()) {`

ZacSharp: Ok, so how about we rename the setting but leave behind a non-functional, `@Deprecated`, `@JavaOnly` setting field with the old name? (There is plenty of repositories which depend on the field for a settings GUI. They were last updated a couple months (3+) ago, but I'd like to avoid finding the maintained ones via crash reports.)

ZacSharp: The server was constantiam iirc.

src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -63,6 +63,9 @@ public class CalculationContext {
     public final List<Block> allowBreakAnyway;
     public final boolean allowParkour;
     public final boolean allowParkourPlace;
+    @Deprecated
+    @JavaOnly
+    public final boolean allowJumpAt256;
```
ZacSharp: `CalculationContext` doesn't need a dummy field. The public API is basically "everything in `baritone.api` and nothing else".

ZacSharp: (That's what I was trying to clarify by using "setting field" instead of "field", guess I failed)

Murat65536:
Could someone check constantiam to see if `allowJumpAtBuildLimit` is still needed? It's been a while.

src/api/java/baritone/api/Settings.java
```diff
@@ -185,7 +185,14 @@ public final class Settings {
...
      */
-    public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
+    public final Setting<Boolean> allowJumpAtBuildLimit = new Setting<>(false);
+
+    /**
+     * Just here so mods that use the API don't break. Does nothing.
+     */
+    @Deprecated
+    @JavaOnly
+    public final Settings<Boolean> allowJumpAt256 = new Setting<>(false);
```
ZacSharp:
```suggestion
    public final Setting<Boolean> allowJumpAt256 = new Setting<>(false);
```

Murat65536: My fault! Did this on GitHub since I didn't have access to an IDE then.

ZacSharp (approved):

leijurv:
i almost never squash but this is way too many commits lol

Murat65536:
Could we close #3949 as completed now?

babbaj:
for contingency `allowJumpAt256` should still be parsed in config files and used to set the new setting

Issue #4739, opened 2025-05-07
Assume SafeWalk does not affect bridging


Murat65536:
## Some information
Operating system: Arch Linux
Java version: 17
Minecraft version: 1.19.4
Baritone version: null
Other mods (if used): 

## How to reproduce
Just make Baritone bridge while AssumeSafeWalk is true

## Modified settings
AssumeSafeWalk is true

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [x] I have included logs, exceptions and / or steps to reproduce the issue.
- [x] I have not used any OwO's or UwU's in this issue.

PR #4740, opened 2025-05-07
Let external safewalk handle being safe.

Murat65536:
#4739


ZacSharp:
Sorry, I do not think this is a good idea. Pressing shift while placing is needed to avoid interacting with blocks (will change in most cases once I finish #4705) and in the second case you disable forward movement for whatever reason, which, according to the comment there, will make it get stuck in some cases.

ZacSharp:
Actually, I forgot about this:
https://github.com/cabaletta/baritone/blob/0fe8c628c75c89dd5a8f31c021a30d94953fc767/src/api/java/baritone/api/Settings.java#L175-L179

So I'd say if you remove the movement change (or explain why it is needed and doesn't break things) this is fine and #4705 can then reintroduce sneaking in some cases and remove the warning.

Murat65536:
Sorry, I kinda forgot about this. Removed movement change so it should be good now.

PR #4742, opened 2025-05-08
Magma block detection

### LINKED ISSUE #6: diagonal over magma

Murat65536:
Closes #6
@leijurv took too long to stop playing with their toenails.


Murat65536:
Should probably adjust cost since you can't sprint and sneak at the same time. Also shouldn't forget make it change with the allowSprint setting. And change ascend jump offset to account for lowered speed.

Murat65536:
So, I might be wrong about this, but MovementPillar keeps sneaking and standing back up each time it places a block for basically no reason. Besides, this being useless (at least to my knowledge), it's something almost all players never do, making it obvious that you're using Baritone to other people who know about this. Anyway, I removed it so now it always sneaks when placing blocks under itself.

ZacSharp:
> making it obvious that you're using Baritone to other people who know about this

If that's a problem then you are doing something wrong. Just don't use Baritone if you shouldn't be using it.

Murat65536:
If Baritone was meant to be only for when you should be using it, we wouldn't need to have things like `antiCheatCompatibility` and `legitMine`.

ZacSharp:
Anarchy servers also have anticheat, sometimes even pretty strong ones (iirc 2b2t uses Grim).

Murat65536:
I guess this change relates a little more to `legitMine` then. You're not technically meant to be using it, but if you are, you might as well not be detected as easily.

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -142,7 +143,11 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
         // For either possible soul sand, that affects half of our walking
         if (destWalkOn.getBlock() == Blocks.SOUL_SAND) {
             multiplier += (WALK_ONE_OVER_SOUL_SAND_COST - WALK_ONE_BLOCK_COST) / 2;
-        } else if (frostWalker) {
+        } else if (Baritone.settings().allowWalkOnMagmaBlocks.value && destWalkOn.getBlock().equals(Blocks.MAGMA_BLOCK)) {
```
leijurv: This should use `context.allowWalkOnMagmaBlocks` since it's in cost calculation

Murat65536:
I didn't see this before but I guess this also closes PR #3690.

src/main/java/baritone/pathing/movement/movements/MovementPillar.java
```diff
@@ -230,7 +230,10 @@ public MovementState updateState(MovementState state) {
             }
 
 
-            state.setInput(Input.SNEAK, ctx.player().position().y > dest.getY() || ctx.player().position().y < src.getY() + 0.2D); // delay placement by 1 tick for ncp compatibility
+            state.setInput(Input.SNEAK, true);
+//            if (ctx.player().position().y > dest.getY() || ctx.player().position().y < src.getY() + 0.2D) {
+//                state.setInput(Input.SNEAK, true); // delay placement by 1 tick for ncp compatibility
+//            }
```
leijurv: Why removed?

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
```diff
@@ -102,14 +102,16 @@ public static void cost(CalculationContext context, int x, int y, int z, Directi
...
-        } else {
-            if (context.canSprint) {
-                maxJump = 4;
-            } else {
-                maxJump = 3;
-            }
+        } else if (context.canSprint) {
+            maxJump = 4;
+        }
+        else {
```
leijurv: Put this back on the previous line

src/main/java/baritone/pathing/movement/movements/MovementPillar.java
leijurv: Generally baritone sneaks before right clicking because sometimes you can right click against a block that opens a GUI.

like imagine trying to pillar up from a crafting table

Does this code work in that case?

Murat65536: Yep. Line below sets it to always sneak, so all that's really changing is Baritone keeps sneaking instead of constantly sneaking and unsneaking on every block place.
`state.setInput(Input.SNEAK, true);`
I mentioned that [here](https://github.com/cabaletta/baritone/pull/4742#issuecomment-2868166975)

src/main/java/baritone/pathing/movement/movements/MovementParkour.java
Murat65536: Like this?
```java
else {
    if (context.canSprint) {
        maxJump = 4;
    }
    else {
        maxJump = 3;
    }
}
```

leijurv: no I literally just mean
`} else {`
theres a style that the rest of the codebase follows

src/main/java/baritone/pathing/movement/movements/MovementDiagonal.java
```diff
@@ -140,9 +141,13 @@ public static void cost(CalculationContext context, int x, int y, int z, int des
...
         // For either possible soul sand, that affects half of our walking
-        if (destWalkOn.getBlock() == Blocks.SOUL_SAND) {
+        if (destWalkOn.is(Blocks.SOUL_SAND)) {
             multiplier += (WALK_ONE_OVER_SOUL_SAND_COST - WALK_ONE_BLOCK_COST) / 2;
-        } else if (frostWalker) {
+        } else if (context.allowWalkOnMagmaBlocks && destWalkOn.is(Blocks.MAGMA_BLOCK)) {
+            multiplier += (SNEAK_ONE_BLOCK_COST - WALK_ONE_BLOCK_COST) / 2;
+            sneaking = true;
+        }
+        else if (frostWalker) {
```
leijurv: Still a few places like this where you put else on a new line

PR #4746, opened 2025-05-10
Add custom block place limit.

### LINKED ISSUE #4743: Add custom height limit for placing blocks
## Describe your suggestion
Some servers impose their own custom height limits. A setting that adjusts how far Baritone is able to build up would be useful.

## Settings
Maybe just a setting called heightLimit.

## Context
This suggestion would make Baritone more adaptable to the weird quirks of servers.

## Final checklist

Murat65536:
A very small PR. Closes #4743


src/main/java/baritone/pathing/movement/CalculationContext.java
```diff
@@ -167,6 +169,9 @@ public double costOfPlacingAt(int x, int y, int z, BlockState current) {
         if (!Baritone.settings().allowPlaceInFluidsFlow.value && !current.getFluidState().isEmpty() && !current.getFluidState().isSource()) {
             return COST_INF;
         }
+        if (y > Baritone.settings().blockPlaceHeightLimit.value) {
```
leijurv: `if (y > blockPlaceHeightLimit) {`

leijurv:
What server imposes a custom height limit ?

ChibiDD:
Any server running data packs such as Sky High which increases the default limit past the 320 limit.

leijurv:
This PR only decreases the limit, it can't increase it

ChibiDD:
Fair I was just assuming as to why people would want a custom height limit, I’m not personally sure what servers are imposing a limit on build height.

Murat65536:
A local SMP server I play on has a custom height limit. If you have any idea on how to make this increase and decrease the height limit, I would really want to hear it though since I couldn't come up with anything. I don't know if I'm supposed to get the instance of the Minecraft world from the settings page but I'm assuming that I'm not.

leijurv:
Can you clarify if this SMP server has a height limit that is lower or greater than vanilla?

Murat65536:
Lower.

wagyourtail:
any data pack raising/lowering the height limit should sync it to the client as world.dimension.height() or whatever as part of the dimension data, so this should only be necessary if some plugin is "artificially" lowering it or otherwise preventing block places above a y value

Murat65536:
The server doesn't use any data packs so I guess it won't sync. I'm thinking that it's done using command blocks.

PR #4747, opened 2025-05-10
Migrate `allowJumpAt256` to `allowJumpAtBuildLimit` in settings files

ZacSharp:
As requested by @babbaj in https://github.com/cabaletta/baritone/pull/4736#issuecomment-2868978202.
I assumed we are neither going to rename / remove more settings in the near future nor going to keep the migration code for long, so I did not try doing this properly.


ZacSharp:
This is currently a release blocker (the only one), solely because it doesn't make sense to merge it after creating a release without the migration code.

leijurv:
I think this is a bit silly but fine

PR #4752, opened 2025-05-11
Bridging Improvements

### LINKED ISSUE #4750: Improve Bridging
## Describe your suggestion
Instead of having the Bot stand at about a 0 degree yaw while placing, have it stand at a 45 degree. This should make it bridge a little faster. Also make the bot continually look backwards instead of looking up and going forwards every single time. ~~(I assume looking up and going forwards increases speed?)~~ Never mind, looking forward is something that was going to be removed anyway:
https://github.com/cabaletta/baritone/blob/c1aa2012b99172c033c2a72bf66e334e847890c1/src/main/java/baritone/pathing/movement/movements/MovementTraverse.java#L361

## Settings
~~Maybe a setting called `bridgingWalkForwards`.~~

## Context
If these two changes are implemented, the bot's bridging should look indistinguishable from a regular player. This would significantly reduce the chance of it being banned.

## Final checklist

Murat65536:
Closes #4750.


src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -352,7 +352,10 @@ public MovementState updateState(MovementState state) {
                 }
                 return state;
             }
-            MovementHelper.moveTowards(ctx, state, positionsToBreak[0]);
+            MovementHelper.moveTowardsWithRotation(ctx, state, dest,
+                    RotationUtils.calcRotationFromVec3d(ctx.playerHead(),
+                            VecUtils.getBlockPosCenter(dest),
+                    ctx.playerRotations()).add(new Rotation(225, 0).withPitch(ctx.playerRotations().getPitch())));
```
leijurv: 225?

Murat65536: 180+45

leijurv: why would you want the 45 degree angle?

Murat65536: 45 degrees is faster.
Just to back that up a little, a snippet from the Minecraft Wiki about sneaking:
> The player moves slower, at a speed of about 1.3 m/s.
> If two movement keys are pressed at once, the player sneaks diagonally at a faster 1.8 m/s.

https://minecraft.wiki/w/Sneaking
Here's also a longer (though less official) explanation of the movement mechanics:
https://youtu.be/piknGokM6rY?t=257
This also points out how sprinting is 2% faster when done sideways. Could be useful but probably not for something so small.

Murat65536: Found the bug report: https://bugs.mojang.com/browse/MC/issues/MC-271065
While it was fixed, it was then reverted: https://feedback.minecraft.net/hc/en-us/community/posts/33341069329549-Petition-to-Revert-Movement-Changes-MC-271065-and-MC-241951-WE-WIN-THESE

Murat65536: nvm this is terrible. I don't know how it worked when I was testing it a few weeks back but the player still spins around. Maybe I had `freeLook` off or something. It's a bit more stable now since it doesn't change the rotation when moving but `attemptToPlaceABlock` should really be the one changing to include a custom yaw. Though even at it's current state this would be a significant improvement to people with `smoothLook` turned on. Seeing your surroundings is also much easier (like when in f5) with this change.

Murat65536:
I don't know how do adjust yaw to a 45 degree angle while bridging and honestly, it might actually be better not to. `attemptToPlaceABlock` isn't just used by `MovementTraverse` so affecting other things would be bad. I feel like this PR still solves `smoothLook` being way slower when bridging so I'm going to say it's good.

Murat65536:
@leijurv Reminder to review.

Murat65536:
@leijurv please I literally had to make a branch from this branch to continue implementing #4838

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -659,6 +662,57 @@ static void moveTowards(IPlayerContext ctx, MovementState state, BlockPos pos) {
...
 
+    static void moveTowardsWithoutRotation(IPlayerContext ctx, MovementState state, BlockPos dest) {
+        float ax = Mth.sin(ctx.playerRotations().getYaw() * DEG_TO_RAD_F);
+        float az = Mth.cos(ctx.playerRotations().getYaw() * DEG_TO_RAD_F);
+        Rotation blockRotation = RotationUtils.calcRotationFromVec3d(ctx.playerHead(),
+                VecUtils.getBlockPosCenter(dest),
+                ctx.playerRotations());
+        int selection = getSelection(blockRotation, ax, az);
+        switch (selection) {
+            case 0 -> state.setInput(Input.MOVE_FORWARD, true);
```
leijurv: This is kinda weird. You are using this `int` as if it were an enum for the 8 possible rotations. I think it would be better if `getOptions` returned an `Option[]` which had as fields `List<Input>` and `Vec2`.

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -659,6 +662,57 @@ static void moveTowards(IPlayerContext ctx, MovementState state, BlockPos pos) {
...
+
+    private static int getSelection(Rotation blockRotation, float ax, float az) {
+        float targetAx = Mth.sin(blockRotation.getYaw() * DEG_TO_RAD_F);
+        float targetAz = Mth.cos(blockRotation.getYaw() * DEG_TO_RAD_F);
+        Vec2[] options = getOptions(ax, az);
+        int selection = -1;
+        float closestX = 100000;
+        float closestZ = 100000;
+        for (int i = 0; i < options.length; i++) {
+            if (Mth.abs(targetAx - options[i].x) + Mth.abs(targetAz - options[i].y) < closestX + closestZ) {
```
leijurv: This usage of x and y is unintuitive, maybe the Option should have as fields motionX and motionZ so that using `y` for `z` isn't happening anymore

src/main/java/baritone/pathing/movement/MovementHelper.java
```diff
@@ -659,6 +662,57 @@ static void moveTowards(IPlayerContext ctx, MovementState state, BlockPos pos) {
...
+            default -> {}
+        }
+    }
+
+    private static int getSelection(Rotation blockRotation, float ax, float az) {
+        float targetAx = Mth.sin(blockRotation.getYaw() * DEG_TO_RAD_F);
+        float targetAz = Mth.cos(blockRotation.getYaw() * DEG_TO_RAD_F);
+        Vec2[] options = getOptions(ax, az);
+        int selection = -1;
+        float closestX = 100000;
```
leijurv: I don't really like this, I feel like better would be to use streams. 

This function should look something like:
```
targetX = blah
targetZ = blah
getOptions(ax, az).stream().min(Comparator.comparingInt(option -> option.distanceTo(targetX, targetZ)).get()
```

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
```diff
@@ -356,9 +356,8 @@ public MovementState updateState(MovementState state) {
                 }
                 return state;
             }
-            MovementHelper.moveTowards(ctx, state, positionsToBreak[0]);
+            MovementHelper.moveTowardsWithoutRotation(ctx, state, dest);
```
leijurv: Is this even a good idea at all? Does this negatively impact baritone's control over the player's motion? It seems like you want to move in a certain direction but the actual direction of movement is up to 45° off from that. This is the sort of thing that can very rarely cause baritone to fall off the block in a way that's hard to catch.

I would feel better if, for example, the vector of movement is always correct. 

Example: let's say the player is currently looking at 30°, and we want to move in direction 200°.

- The current code would simply look at 200° then press W.
- Your code would keep looking at 30° and press S, which would cause it to move towards 210°

This is off by 10°. What about this alternative

- Look at 20° (only ten degrees off from continuing to face the same way, **so it's very likely still looking at the block that it wants to place against**), and press S. This would cause the player to move **exactly** in the direction we want to, without spinning 180° (it would only spin up to 45° at most). or maybe 22.5 at worst actually?

src/main/java/baritone/pathing/movement/MovementHelper.java
leijurv: And your math is kinda weird too, really the comparison is "what yaw is closest to my desired yaw" but you're doing trig? I don't think it's necessary. I think it would be better to compare differences in yaw rotations rather than distances in (x,z)

src/main/java/baritone/pathing/movement/MovementHelper.java
Murat65536: Oh, that was stupid of me. Thanks!

src/main/java/baritone/pathing/movement/MovementHelper.java
Murat65536: > And your math is kinda weird too, really the comparison is "what yaw is closest to my desired yaw" but you're doing trig? I don't think it's necessary. I think it would be better to compare differences in yaw rotations rather than distances in (x,z)

Oh... I didn't really think of it this way. for me, it was more like: what selection of inputs will get us closest to our desired direction without changing the yaw? Because the whole reason this thing exists is to not change the yaw while moving in any arbitrary direction. That's why I thought the trig was necessary. I wanted to see which direction I went in with the yaw I had and adjust the inputs I sent based off that.

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
Murat65536: From what I can see this isn't exactly the case.
> It seems like you want to move in a certain direction but the actual direction of movement is up to 45° off from that. 

Since we have 8 different movement options, the maximum angle that we can be off is 22.5 degrees (the middle of any 2 of the 8 options, or 360/16), which is better. Even then, being off doesn't really impact us that much. Since this method is called every time the movement updates, the calculation it gives also gets updated each time. So lets say it drifts enough for it to matter since our yaw is at some undesirable value. As soon as another movement option is more favorable (gets us going closer to the right direction), the bot would just switch to that one. You can really think of this as less of a straight line and more of a slight zig-zag when it's a little off.

Murat65536: I feel like I should emphasize “slight” zig-zag. With even the maximum of 22.5 degrees offset, it’s not really noticeable.

leijurv: I hear what you're saying about it updating every tick, and yes at the end of my comment I realized 22.5 is correct

However 1. the player has momentum and 2. from the center of the block to colliding with a side block is a very small distance. If you are standing at the center, that's 0.5, and touching a side block is 0.7. That is not that far of a distance. You could cover it very quickly.

With the suggestion I gave above, where it adds the "up to 22.5°" of error to the actual yaw, what do you think would go wrong? or what's the downside you think?

src/main/java/baritone/pathing/movement/MovementHelper.java
leijurv: After looking at the part about sprinting, I think you are correct because if you are sprinting and press WA you go at a different angle than walking while pressing WA. Right? It's a 1 to 1.3 ratio rather than 1 to 1? If that's true then comparing the XZ direction makes sense I guess. If you're sure that's true then I guess it makes sense.

src/main/java/baritone/pathing/movement/movements/MovementTraverse.java
Murat65536: It seems like it would work, but it might be in a separate method since that’s not really something #4838 would benefit from. I guess I am wondering if it’s all necessary. I don’t really know how much that zigzag gets it off track. As for what can go wrong, I don’t really have any reason why limiting us to 45 degrees angles is a bad thing, but like I said before, is it needed? I’ll try to figure out a way to measure this in the next few days if I got time. It’ll be done on the weekend if I don’t. I had a lot more time on summer break :(.

Murat65536: Yea... I've played around with it for a bit and it just isn't stable enough to have fully yaw-independent movement. (or maybe I'm missing something and it's just calculating wrong). I mean, it would be an amazing thing to see this improved and it doesn't exactly look like it's calculating correctly but imma leave that aside until I either have an epiphany or implement #4834.

Murat65536:
@leijurv ping

leijurv:
> it just isn't stable enough to have fully yaw-independent movement
https://github.com/cabaletta/baritone/pull/4752#discussion_r2296386106

This doesn't make sense to me. How could it be less stable? If you compute an exact combination of yaw, and WASD input, to go in a certain direction, then you set your yaw to exactly that, won't it move in exactly the desired direction? whereas with the current code, it could move up to 22.5° off from the desired direction?

Murat65536:
@leijurv I updated my code to add the change you wanted. That was my reasoning for doing that.
https://github.com/cabaletta/baritone/pull/4752/commits/a2a6769742d87235466d2a9d3014fe531d4728cd

leijurv:
That's really not the same at all.

It doesn't matter which direction the player is already facing. The point is that you'd take the desired direction, and move in EXACTLY that direction, through some combination of WASD and moving the yaw up to 22.5 degrees. Your approach rounds the yaw to a multiple of 45, which would only make sense if the desired direction was also a multiple of 45, which it probably isn't exactly

Murat65536:
My thought process was that attemptToPlaceBlock should set the needed rotation and that all that would be left is rounding the yaw after that.
Edit: I might be seeing some holes in this plan. I can check in 2 hours or tomorrow if I don’t got time. You might be right about the multiples of 45 (but baritone only has traverse and diagonal so idrk. I still want to fix it)
Edit 2: Yea, I should be doing modulus 45 or something my bad

leijurv:
The procedure should be
1. Attempt to place a block. This defines the approximate yaw that we want to be looking at.
2. Move towards the next block. This defines the EXACT direction we will move in.

Call those yaw1 and yaw2. We want to look as close as we can to yaw1, while moving towards yaw2.

We will go through all the possible MovementOptions. For each one, we will compute it in reverse. Instead of "if I looked at this yaw and did this movement, which direction would I go?", we need the opposite: "if I did this movement and I wanted to go in this direction, what yaw would I need?". We will pick the MovementOption that wants a yaw closest to yaw1. Call that yaw3.

So, we have picked a MovementOption, where if we look in direction yaw3, the player will move EXACTLY towards yaw2. Of all movement options, this one was chosen because it's the closest to yaw1, but it probably doesn't exactly equal yaw1.

We will then do that movement option while looking at yaw3. This means we will MOVE exactly towards the target block, but the direction we will look for placement will be slightly off. But that's really not a problem since the area where we can place a block is pretty big.

pizzav-xyz:
could you also make a mode that just sprints while extend scafolding

Murat65536:
Like the scaffold hack? I don't really have much experience with scaffold but I doubt that change is going to happen now. Lets at least get some more legit bridging methods in before we focus on something not everyone uses. I'm doing this because I want to look around in 3rd person while bridging, I want to bridge faster with `smoothLook`, and I want to implement attacking (this is one step of that).

Murat65536:
Ok, now it should be all good. Yaw distance is annoying

Murat65536:
> The procedure should be
> 1. Attempt to place a block. This defines the approximate yaw that we want to be looking at.
> 2. Move towards the next block. This defines the EXACT direction we will move in.
> 
> Call those yaw1 and yaw2. We want to look as close as we can to yaw1, while moving towards yaw2.
We will go through all the possible MovementOptions. For each one, we will compute it in reverse. Instead of "if I looked at this yaw and did this movement, which direction would I go?", we need the opposite: "if I did this movement and I wanted to go in this direction, what yaw would I need?". We will pick the MovementOption that wants a yaw closest to yaw1. Call that yaw3.
> So, we have picked a MovementOption, where if we look in direction yaw3, the player will move EXACTLY towards yaw2. Of all movement options, this one was chosen because it's the closest to yaw1, but it probably doesn't exactly equal yaw1.
> We will then do that movement option while looking at yaw3. This means we will MOVE exactly towards the target block, but the direction we will look for placement will be slightly off. But that's really not a problem since the area where we can place a block is pretty big.

@leijurv 
Just as an overview, this is what I did:
I thought of the player as a point with one ray coming out of said point at an angle of the current yaw.
I then thought of 7 more rays at 45 degree offsets from the current yaw (which would make them an equal distance away from each other) to represent the other directions the player can take.
I drew one last ray to represent the ideal yaw.
Since Minecraft's yaw system works by going 0 to 180, then straight to -180 and back down to 0, I did some math in `yawDistanceFromOffset` to calculate the shortest distance between the current yaw ray and the offset yaw ray.
Since I know for a fact that every 45 degrees there's going to be another ray, I can safely do modulus 45 to only get the discrepancy in the yaw that I need to account for.
After this, I need to figure out if the discrepancy in the yaw would be smaller by turning left or right, which I do with a little more math in `moveTowardsWithSlightRotation`.
That should be it, but there's always a chance that it's not. LMK if you find anything.

Murat65536:
@leijurv sorry for asking so many times but review please.

leijurv:
Yeah that looks like it could be right, it's a bit weirdly written so I can't quite tell from reading. If you send a screen recording of it working, that shows how it works (like, with freelook turned off), then I'll merge I think

Murat65536:
https://github.com/user-attachments/assets/9ec07a6d-3ba1-4202-92c0-52978cd9b456

PR #4764, opened 2025-05-22
reduce PrecomputedData memory usage

rfresh2:
there's around 25k block states

so this reduces mem needed from ~800kb -> ~200kb

### DIFF
--- a/src/main/java/baritone/pathing/precompute/PrecomputedData.java
+++ b/src/main/java/baritone/pathing/precompute/PrecomputedData.java
@@ -29,23 +29,21 @@ public class PrecomputedData {
     /**
      * byte layout
      *
-     *          0              1             2              3              4              5              6             7
+     *          7              6             5              4              3              2              1             0
      *          |              |             |              |              |              |              |             |
      *      unused         canWalkOn       maybe       canWalkThrough    maybe        fullyPassable    maybe       completed
      */
 
-    private static final byte COMPLETED_MASK = (byte) 0x01;
-    private static final byte FULLY_PASSABLE_MAYBE_MASK = (byte) 0x02;
-    private static final byte FULLY_PASSABLE_MASK = (byte) 0x04;
-    private static final byte CAN_WALK_THROUGH_MAYBE_MASK = (byte) 0x08;
-    private static final byte CAN_WALK_THROUGH_MASK = (byte) 0x10;
-    private static final byte CAN_WALK_ON_MAYBE_MASK = (byte) 0x20;
-    private static final byte CAN_WALK_ON_MASK = (byte) 0x40;
-
-    private static final byte zero = (byte) 0;
+    private static final byte COMPLETED_MASK = (byte) 1 << 0;
+    private static final byte FULLY_PASSABLE_MAYBE_MASK = (byte) 1 << 1;
+    private static final byte FULLY_PASSABLE_MASK = (byte) 1 << 2;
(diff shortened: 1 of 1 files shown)


src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -22,46 +22,47 @@
...
+
+    /**
+     * byte layout
+     *
+     *          0              1             2              3              4              5              6             7
+     *          |              |             |              |              |              |              |             |
+     *      unused         canWalkOn       maybe       canWalkThrough    maybe        fullyPassable    maybe       completed
+     */
+
+    private static final byte COMPLETED_MASK = (byte) 0x01;
```
leijurv: I think this is worse, can this go back to `(byte) 1 << 0` or some such?

src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -22,46 +22,47 @@
...
-
 public class PrecomputedData {
 
-    private final int[] data = new int[Block.BLOCK_STATE_REGISTRY.size()];
+    private final byte[] data = new byte[Block.BLOCK_STATE_REGISTRY.size()];
+
+    /**
+     * byte layout
+     *
+     *          0              1             2              3              4              5              6             7
```
leijurv: This is big endian 👎

leijurv: (I think it would be more clear, relative to the subsequent lines, if "0" meant the least significant bit, because you could see that "1 << 0" corresponds with bit 0, "1 << 1" corresponds with bit 1, etc) (in other words, little endian ordering where "0" is the LSB not the MSB)

src/main/java/baritone/pathing/precompute/PrecomputedData.java
```diff
@@ -22,46 +22,47 @@
...
+    private static final byte CAN_WALK_ON_MASK = (byte) 0x40;
 
-    private static final int COMPLETED_MASK = 1 << 0;
-    private static final int CAN_WALK_ON_MASK = 1 << 1;
-    private static final int CAN_WALK_ON_SPECIAL_MASK = 1 << 2;
-    private static final int CAN_WALK_THROUGH_MASK = 1 << 3;
-    private static final int CAN_WALK_THROUGH_SPECIAL_MASK = 1 << 4;
-    private static final int FULLY_PASSABLE_MASK = 1 << 5;
-    private static final int FULLY_PASSABLE_SPECIAL_MASK = 1 << 6;
+    private static final byte zero = (byte) 0;
```
leijurv: huh??

leijurv: is this to prevent implicit upcasting to int on both sides when you would do `myByte == 0`?

src/main/java/baritone/pathing/precompute/PrecomputedData.java
rfresh2: java is big endian afaik

but yeah thats fine

src/main/java/baritone/pathing/precompute/PrecomputedData.java
rfresh2: yes, i have seen bugs and issues due to int <-> byte casting

most commonly when combined with arithmetic

so i prefer being very explicit

its fine though, not causing a bug here, so this is just preference and  i'll defer to you

PR #4820, opened 2025-07-20
1.21.8

rfresh2:
- [x] Fabric
- [x] Forge
- [x] NeoForge
- [x] Testing
- [ ] ~~Happy Ghast parkour~~

Download: https://nightly.link/cabaletta/baritone/actions/runs/16396403911/Artifacts.zip

**Baritone 1.21.8 is also compatible with MC 1.21.6 and 1.21.7**

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -25,7 +25,7 @@
   ],
   "depends": {
     "fabricloader": ">=0.14.22",
-    "minecraft": ["1.21.6", "1.21.7"]
+    "minecraft": ["1.21.6", "1.21.7", "1.21.8"]
   },
   "custom": {
     "modmenu": {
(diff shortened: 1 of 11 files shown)


ZacSharp:
Just noticed that bumping the version field in update prs like this without creating a tag means that the git vs. non-git version numbers differ not only in the extra information they provide but also in the major/minor/patch triple (e.g. `1.14.0-1-g612a8a6d` vs `1.15.0-local` in this case).

I'll just do what I apparently already did for v1.14.0 and put the version tag directly on the merge commit, without an extra commit to bump the version. That means there's still a month of builds using `1.14.0-1-g612a8a6d` instead of plain `1.15.0`, but since they include the commit hash that should be fine.

ZacSharp:
Released as [v1.15.0](https://github.com/cabaletta/baritone/releases/tag/v1.15.0)

PR #4826, opened 2025-07-24
Pause if Inventory open.

### LINKED ISSUE #4715: Toggle throwaway slot
## Describe your suggestion
The throwaway slot can be annoying to many individuals, so I'm asking that you either make it only swap when building, or implement my change.

## Settings
+ to Settings.java: `public final Setting<Boolean> autoMoveThrowawayBlocks = new Setting<>(true);`
edit to InventoryBehavior.java:
`
        if (Baritone.settings().autoMoveThrowawayBlocks.value && firstValidThrowaway() >= 9) { // aka there are none on the hotbar, but there are some in main inventory
            requestSwapWithHotBar(firstValidThrowaway(), 8);
        }
`

## Context
Users get annoyed by it automatically adding dirt or cobblestone to their 9th slot when they aren't building, so I thought this would be a useful change! Please add settings for this too if you add it. Thank you :)

## Final checklist

### LINKED ISSUE #4825: Stop if inventory has been manually opened
## Describe your suggestion
Baritone currently has a setting called `inventoryMoveOnlyIfStationary`, which comes to a stop before doing any inventory actions. But if you manually open the inventory, Baritone keeps on moving. This is able to be detected by anticheat. Maybe just integrate it with `inventoryMoveOnlyIfStationary` and change the name to `stopWhenInventoryOpen` when adding this functionality. In fact, this applies to all GUIs, not just inventories.

## Final checklist

Murat65536:
Closes #4825
Closes #4715
Stops executing any movements when it's in the inventory. Also should send inventory open events to server when it's changing item positions.
Servers are able to detect if a player has interacted with their inventory while they are still moving. This would stop Baritone to prevent that.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -103,6 +103,13 @@ public final class Settings {
     public final Setting<Boolean> stopWhenInventoryOpen = new Setting<>(false);
 
     /**
+     * Just here so mods that use the API don't break. Does nothing.
+     */
+    @Deprecated
+    @JavaOnly
+    public final Setting<Boolean> inventoryMoveOnlyIfStationary = new Setting<>(false);
+
+    /**
      * Disable baritone's auto-tool at runtime, but still assume that another mod will provide auto tool functionality
      * <p>
      * Specifically, path calculation will still assume that an auto tool will run at execution time, even though
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -98,9 +98,9 @@ public final class Settings {
     public final Setting<Integer> ticksBetweenInventoryMoves = new Setting<>(1);
 
     /**
-     * Come to a halt before doing any inventory moves. Intended for anticheat such as 2b2t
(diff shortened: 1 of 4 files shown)


Murat65536:
Used @babbaj's code to add conditional item moving.

Murat65536:
I found another issue while testing this, which is that pillaring makes the player continuously jump when you have a high `ticksBetweenInventoryMoves` setting. But since this issue wasn't caused by me and it shouldn't really affect anyone (I tested it with a value of 10 and was fine. Showed up with a value of 100 though), I'm going to leave it. It was slightly worsened though since `ticksSinceLastInventoryMove` only increments when you call `requestSwapWithHotBar`.

Murat65536:
Looking back on this, the thing that servers are able to detect is when the player interacts with their inventory while moving. I'm thinking that instead of stopping Baritone when an inventory is open, we could just add a setting that doesn't let the player interact with anything in the inventory when Baritone is running. This would be better because the player might still want to see what's in their inventory while the bot is running. To actually interact with the inventory, the player can just pause the bot.

Murat65536:
Abandoned

PR #4836, opened 2025-08-04
Build System Revamp

Murat65536:
First update to dependencies in 2 years.
- [x] Update Unimined to 1.4.1
- [x] Update Gson to 2.9.1
- [x] Update Commons IO to 2.20.0
- [x] Fix Forge `runClient`
- [x] Refactor the whole thing to be Kotlin DSL
- [x] Update to Gradle 9
- [ ] Hide Baritone from servers in Forge
- [x] ~Fix Tweaker~ Remove Tweaker (I don't even know why it exists)
- [x] Update Shadow
- [x] ~Fix Proguard~ Switch from ProGuard to R8


Murat65536:
I'm waiting on #4698 to merge. That will upgrade the nether pathfinder version to 1.6.

Murat65536:
Changed shadow to new URL because of GradleUp/shadow#908.

Murat65536:
@leijurv Instead of working on 1.19.4 forever, I feel like Baritone should be developed on the latest Minecraft version. I know that matching the current 2b2t version might be a reason to keep it at 1.19.4, but people can just use ViaFabricPlus and also 2b2t updated to 1.21.4 anyway.

Murat65536:
Also Dependabot or something would be good

leijurv:
My position is that baritone should just follow the version of 2b2t. So yes we could update the development version accordingly

Murat65536:
wtf why is copilot here

Murat65536:
Abandoned

PR #4838, opened 2025-08-05
Attac

Murat65536:
Closes #1933
Closes #4517
Attacking functionality for Baritone. Could stop creepers from blowing up the player or just be used for general PVP. This PR has an advantage over impact since it allows the player to actually move while they're attacking. (Also Impact is dead lol).
~- [ ] Compatibility with `smoothLook` (or just completely replace it and get something like #4505).~ *Imma say that if `allowParkourPlace` doesn't need to be compatible with `smoothLook`, attacking doesn't either. (100% not just me being lazy)*
- [x] Option to randomly switch the circling direction with a setting on how often to switch.
~- [ ] Make attacking work with a GUI open.~ This would probably trigger anticheat
- [x] Change attacking functionality to work even when there isn't a goal. Maybe a `BaritoneProcessHelper`?
- [x] Switch to best sword to attack.
- [ ] API option for determining what to attack (ex: ignore passive mobs)

PR #4846, opened 2025-08-10
Adding Sprint Jumping & Headhits to baritone for improved traversal. 

AverWasTaken:
showcase video -> https://www.youtube.com/watch?v=stUSnUsQpmU

----

jump-sprint: Automatically times jump-sprinting on flat segments to maximize speed while staying centered and stable.

looks ahead for turns, step-ups, drops, and low ceilings; switches to regular sprint or suppresses jumps when we prob shouldnt be doing cool shit.

diagonal and ascend cadence: supports flat diagonals without midair slowdown climbs one-block steps cleanly with speed-aware takeoff (wow so awesome)

midair momentum and headroom: keeps sprint in-air, avoids head-bonks, and can spam jump under low ceilings for extra speed. (speedrun type shit)

settings: allowJumpSprint, allowJumpSprintDiagonal, and allowHeadHits.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -809,6 +809,11 @@ public final class Settings {
     public final Setting<Boolean> sprintInWater = new Setting<>(true);
 
     /**
+     * Jump while sprinting on flat traverses for extra speed
+     */
+    public final Setting<Boolean> jumpsprint = new Setting<>(false);
+
+    /**
      * When GetToBlockProcess or MineProcess fails to calculate a path, instead of just giving up, mark the closest instance
      * of that block as "unreachable" and go towards the next closest. GetToBlock expands this search to the whole "vein"; MineProcess does not.
      * This is because MineProcess finds individual impossible blocks (like one block in a vein that has gravel on top then lava, so it can't break)
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -811,7 +811,17 @@ public final class Settings {
     /**
      * Jump while sprinting on flat traverses for extra speed
      */
-    public final Setting<Boolean> jumpsprint = new Setting<>(false);
+    public final Setting<Boolean> allowJumpSprint = new Setting<>(false);
+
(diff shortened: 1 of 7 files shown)


Murat65536:
There are a few issues I have with this PR:
- All PRs should be targetting the currently maintained version 1.19.4. They will then be ported over to later versions by others. Your PR targets 1.21.8, which just isn't really acceptable.
- Is there any reason you didn't try to look at `MovementParkour` instead? Your current implementation seems unnecessarily complicated and just from looking at it, very prone to error since it's not something that extends `Movement`. This is by far the largest gripe I have with this code. This also means it has no impact on cost, which is just BAD.
- This doesn't respect `allowSprint`.
- In `MovementAscend`, you removed the code to stop when the movement isn't running anymore.
- In `MovementDiagonal`, you added something that will set sprint to true if it meets a condition despite the line just above that setting sprint to true in all cases.
- You completely removed the y-axis check for `MovementTraverse`.
- The headhitters are done by holding jump instead of spamming it. To change this, you will need to edit the code a little but without this change, headhitters are basically useless.
- There are cases where it can just fall (see video)

https://github.com/user-attachments/assets/c961fd43-c397-41d5-8248-5f990cefc80d

Your current implementation... could use some work. I see no way this can even be considered to be merged before the whole thing is revamped to work with `MovementParkour`. I mean, another option is to just wait for me to do it. I started trying basically the same thing based in `MovementParkour` [here](https://github.com/Murat65536/baritone/tree/parkour-on-traverse) a few days ago. I haven't really devoted much time into it since I've mainly been focusing on #4838 but I should hopefully be able to start back on that branch soon.

Murat65536:
Here's my PR for the same idea: #4847.
What you're trying to do can be accomplished in a way that ~requires much less code,~ looks much cleaner, and doesn't fall. Keep in mind that it's still a draft PR and anything about it is subject to change at any time.

leijurv:
This looks pretty uncontrolled from the first few seconds of your video. Looks like it is hitting the side blocks (i.e. it would hit side blocks whatever they may be).

Issue #4538, opened 2024-10-28
1.21.1 - ATM10 - The Undergarden + Silent Gear + Baritone causes crash


Dokturek991:
## Some information
Operating system: Windows 11 Pro (23H2)
Java version: Java 8 Update 421
Minecraft version: 1.21.1
Baritone version: baritone-api-neoforge-1.10.2-9-g3b01f129
Other mods (if used): The Undergarden, Silent Gear (+lib)
Neoforge version: 21.1.72

## Exception, error or logs
I will provide three files, just to be sure. One is with both mods enabled and the other two are with one disabled.
[crash_silent_undergarden_on.txt](https://github.com/user-attachments/files/17540671/crash_silent_undergarden_on.txt)
[crash_silent_off.txt](https://github.com/user-attachments/files/17540672/crash_silent_off.txt)
[crash_undergarden_off.txt](https://github.com/user-attachments/files/17540673/crash_undergarden_off.txt)

## How to reproduce
Download ATM10 and add Baritone. You will not be able to create world as it always crashes game before it gets a chance. 
OR
Create custom modpack with said mods and add baritone. Same behaviour.

## Modified settings
None.

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [x] I have included logs, exceptions and / or steps to reproduce the issue.
- [x] I have not used any OwO's or UwU's in this issue.

I think that this is my first time ever posting an issue, so if I missed something, just let me know. Thank you for making this mod as it saves my time so I can grind less and spend more time with family after job.

Dokturek991:
Forgot to provide mod verions:
silent-gear-1.21.1-neoforge-4.0.8.
The_Undergarden-1.21.1-0.8.22

ZacSharp:
This is two different incompatibilities.
1. Undergarden seems to assume nobody will ever query the damage value of an `ItemStack` before a world is loaded. I'd blame that on them since Minecraft does construct `ItemStack`s before world load and assuming that no mod will query the damage of one of those stacks seems rather brittle.
2. ~Silentlib~ SilentGear somehow wants to construct a new `ItemStack` every time one of its items is asked for its damage, and Baritone reads the damage of every new `ItemStack`, leading to infinite recursion. While I think constructing a new `ItemStack` in `getDamage` is a weird thing to do, I also think this one is on Baritone to fix.

[...]

PR #4687, opened 2025-03-25
Calculate item hash only if needed

ZacSharp:
Fixes a stackoverflow with SilentGear.
The problem is that SilentGear has a complicated parts system and to get the damage value of an `ItemStack` it first constructs the relevant `PartInstance`, which involves defensively copying the `ItemStack` before storing it in the new `PartInstance`. Copying an `ItemStack` of course triggers our hash calculation in the constructor run for the copy, we query the damage value of the new stack, and the cycle repeats.

With this change we only calculate the hash once we actually need it (and if the hash happens to be 0 we do so every time it is needed).
An alternative would be to try copying the hash from the old stack to the new one, though given the Mixin hackery this would require I'm not sure whether that would actually be better. (Need to redirect the relevant constructor call and probably also need a `ThreadLocal` to pass the hash)

Fixes #4538
Fixes #4863

PR #4947, opened 2025-12-30
Fix for neoforge ItemStack init recursion

samhpr:
this fixes a crash on neoforge 1.21 where baritone’s ItemStack mixin can end up calling itself over and over

this happens when other mods access ItemStack data like durability or hashes while an ItemStack is still being created, which causes baritone’s init hook to re-enter and eventually stack overflow or crash on join

the fix adds a small threadlocal guard so the ItemStack init hook does nothing if it’s already running, which breaks the recursion without changing normal behavior

seen with mods like silent gear, but this should apply generally to neoforge setups

fix & related: https://github.com/cabaletta/baritone/issues/4538

### DIFF
--- a/src/launch/java/baritone/launch/mixins/MixinItemStack.java
+++ b/src/launch/java/baritone/launch/mixins/MixinItemStack.java
@@ -18,6 +18,7 @@
 package baritone.launch.mixins;
 
 import baritone.api.utils.accessor.IItemStack;
+import baritone.launch.util.ItemStackInitGuard;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import org.spongepowered.asm.mixin.Final;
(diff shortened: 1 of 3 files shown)


src/launch/java/baritone/launch/mixins/MixinItemStack.java
```diff
@@ -50,7 +51,15 @@ private void recalculateHash() {
             at = @At("RETURN")
     )
     private void onInit(CallbackInfo ci) {
+        if (ItemStackInitGuard.isEntered()) {
+            return;
+        }
+        ItemStackInitGuard.enter();
+        try {
         recalculateHash();
+        } finally {
```
ZacSharp, line 60:
```suggestion
        try {
            recalculateHash();
        } finally {
```
Slight formatting error.

src/launch/java/baritone/launch/util/ItemStackInitGuard.java
ZacSharp, line 1: I don't think we need a separate class for this. The thread local can just be an `@Unique static private final` field in `MixinItemStack` and the methods here inlined.

ZacSharp (commented):
Somewhat of a duplicate of https://github.com/cabaletta/baritone/pull/4687 but using a different approach. I prefer the approach in #4687, though I might be biased by being the author of #4687.

Also something seems to be wrong with your commit history.

leijurv:
merged #4687

PR #4953, opened 2026-01-04
Allow mining / replacing by tag

### LINKED ISSUE #1089: 'Block' parameters in commands should support tags

ZacSharp:
Now that I know we can use tags I figured I'd try how this would work out.

Closes #1089

### DIFF
--- a/src/api/java/baritone/api/command/datatypes/ForBlockOptionalMeta.java
+++ b/src/api/java/baritone/api/command/datatypes/ForBlockOptionalMeta.java
@@ -18,28 +18,13 @@
 package baritone.api.command.datatypes;
 
 import baritone.api.command.exception.CommandException;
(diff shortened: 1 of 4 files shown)


BaoLam-hue:
May I ask, what is the use of tags in mining?

PR #4968, opened 2026-02-16
elytra jump takeoff

rfresh2:
in 1.12, efly takeoff was decided by the server. 

to be reliable on high latency, baritone needed to walk off a cliff to get fall distance

but in modern mc, takeoff is controlled client side and latency is no longer an issue

so we can just jump in place and go immediately

### DIFF
--- a/src/main/java/baritone/process/ElytraProcess.java
+++ b/src/main/java/baritone/process/ElytraProcess.java
@@ -18,16 +18,12 @@
 package baritone.process;
 
 import baritone.Baritone;
-import baritone.api.IBaritone;
 import baritone.api.event.events.*;
 import baritone.api.event.events.type.EventState;
 import baritone.api.event.listener.AbstractGameEventListener;
 import baritone.api.pathing.goals.Goal;
 import baritone.api.pathing.goals.GoalBlock;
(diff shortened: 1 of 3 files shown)


src/main/java/baritone/process/ElytraProcess.java
```diff
@@ -205,74 +191,55 @@ public PathingCommand onTick(boolean calcFailed, boolean isSafeToCancel) {
...
-                            (fall.getSrc().x + fall.getDest().x) / 2,
-                            (fall.getSrc().y + fall.getDest().y) / 2,
-                            (fall.getSrc().z + fall.getDest().z) / 2
+            final boolean canStartFlying = !ctx.player().isOnGround() && !ctx.player().isFallFlying();
+
+            if (canStartFlying) {
+                this.state = State.START_FLYING;
+            } else {
+                // todo: would be better if we searched for a safe pos to takeoff from, and then pathed directly there
+                if (System.currentTimeMillis() - startedJumpStateTime > 2500) {
```
leijurv, line 218: I'd rather count ticks than millis

leijurv:
can i see a screen recording

PR #4979, opened 2026-03-15
renderGoalXZBeacon issue fix

Azzar-Aixel:
Fixed by implementing a custom Beacon Beam RenderPipeline using `RenderPipelines#register` with depth testing disabled. 
Because the `register` method is private and couldn't be exposed through an AccessWidener (no idea why the AccessWidener isn't working despite setup correctly), it is accessed via mixin instead.

A custom event (`RenderBlockEntitiesEvent`) was also introduced, and the beacon beam rendering helpers were ported from the Minecraft client source and adapted to work with the new pipeline.

`RenderBlockEntitiesEvent` is used instead of `RenderEvent` because rendering the beam in the general render stage causes noticeable misalignment when the camera FOV changes (e.g., sprinting) or when the camera moves. Rendering during the block entity stage ensures the correct pose stack and camera transforms are applied, keeping the beam aligned with the world.

Using `RenderEvent` 

image

Using `RenderBlockEntitiesEvent`

image

### DIFF
--- a/src/main/java/baritone/utils/IRenderer.java
+++ b/src/main/java/baritone/utils/IRenderer.java
@@ -28,18 +28,14 @@ import com.mojang.blaze3d.platform.DepthTestFunction;
 import com.mojang.blaze3d.platform.DestFactor;
 import com.mojang.blaze3d.platform.SourceFactor;
 import com.mojang.blaze3d.vertex.*;
-import com.mojang.math.Axis;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.RenderPipelines;
-import net.minecraft.client.renderer.SubmitNodeCollector;
 import net.minecraft.client.renderer.blockentity.BeaconRenderer;
 import net.minecraft.client.renderer.rendertype.RenderSetup;
 import net.minecraft.client.renderer.rendertype.RenderType;
 import net.minecraft.client.renderer.rendertype.RenderTypes;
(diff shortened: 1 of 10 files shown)


src/main/java/baritone/utils/IRenderer.java
```diff
@@ -173,4 +214,58 @@ static void emitLine(BufferBuilder bufferBuilder, PoseStack stack, Vec3 start, V
         emitLine(bufferBuilder, stack, start.x - vpX, start.y - vpY, start.z - vpZ, end.x - vpX, end.y - vpY, end.z - vpZ, lineWidth);
     }
 
+    static RenderType beaconBeam(Identifier identifier, boolean bl) {
+        return BEACON_BEAM.apply(identifier, bl);
+    }
+
+    static void submitBeaconBeam(PoseStack poseStack, SubmitNodeCollector submitNodeCollector, Identifier identifier, float f, float g, int i, int j, int k, float h, float l) {
```
leijurv: this looks pasted from decompiled code

Azzar-Aixel: yeah, this is a bad idea as i used the decompiled src code to render the beam, well shit... i have to find another way to code it's own beacon beam renderer.

Azzar-Aixel:
done, added a custom renderer for the beacon XZ goal instead of just rendering using the (copied) decompiled code.

leijurv:
okay I guess this is fine

PR #4983, opened 2026-03-30
neo 1.21.11 + update unimined & gradle

rfresh2:
neo build was broken on unimined 1.2.9 and dropped from the initial port


ZacSharp:
Do you know how close Unimined is to a release? I would strongly prefer not using `-SNAPSHOT` for a Baritone release.

Looks good otherwise.

rfresh2:
@wagyourtail ^

PR #4999, opened 2026-04-08
Fix BetterBlockPos leaking into block entity map, fixes #4307

### LINKED ISSUE #4307: crash on custom goal request
```java
case ShulkerSearchPathing: {
    if (baritone.getCustomGoalProcess().isActive()) {
        Helper.HELPER.logDirect("custom goal to path to shulker");
        return; // Wait to get there
    }

    curCheckingShulker = getShulkerToCheck();
    Helper.HELPER.logDirect("found current shulker: " + curCheckingShulker.toString());
    if (curCheckingShulker == null) {
        ...
    }

    Optional<Rotation> shulkerReachable = RotationUtils.reachable(ctx.player(), curCheckingShulker,
            ctx.playerController().getBlockReachDistance());

    if (shulkerReachable.isPresent()) {
        ...
    } else {
        Helper.HELPER.logDirect("Going to shulker");
        // crash happens here vvv
        baritone.getCustomGoalProcess().setGoalAndPath(new GoalBlock(getPathingSpotByShulker(curCheckingShulker)));
    }
    break;
}
```

This leads to this error 
```java
[19:38:34] [Render thread/ERROR]: Unreported exception thrown!
java.lang.IllegalArgumentException: Multiple entries with same key: class_2338{x=-322, y=63, z=-61}=net.minecraft.class_2627@5c367bff and BetterBlockPos{x=-322,y=63,z=-61}=net.minecraft.class_2627@32300150
``` 
and crashes the rendering thread so crashing the game.

mankool0:
In these cases minecraft's functions `getShape` and `getCollisionShape` get called with `BetterBlockPos` and minecrafts block entity map doesn't contain that position yet then you'll get a crash.

The `reachable` function has a leak in `BuilderProcess::toBreakNearPlayer`:
```java
if (!(curr.getBlock() instanceof AirBlock) && !(curr.getBlock() == Blocks.WATER || curr.getBlock() == Blocks.LAVA) && !valid(curr, desired, false)) {
    BetterBlockPos pos = new BetterBlockPos(x, y, z);
    Optional<Rotation> rot = RotationUtils.reachable(ctx, pos, ctx.playerController().getBlockReachDistance());
    if (rot.isPresent()) {
        return Optional.of(new Tuple<>(pos, rot.get()));
    }
}
```

And `calculateBlockCenter` has a leak in `MovementTraverse::updateState` as `dest` is `protected final BetterBlockPos dest;`:
```java
// combine the yaw to the center of the destination, and the pitch to the specific block we're trying to break
// it's safe to do this since the two blocks we break (in a traverse) are right on top of each other and so will have the same yaw
float yawToDest = RotationUtils.calcRotationFromVec3d(ctx.playerHead(), VecUtils.calculateBlockCenter(ctx.world(), dest), ctx.playerRotations()).getYaw();
```

This fixes issue #4307

### DIFF
--- a/src/api/java/baritone/api/utils/RotationUtils.java
+++ b/src/api/java/baritone/api/utils/RotationUtils.java
@@ -175,6 +175,10 @@ public final class RotationUtils {
     }
 
     public static Optional<Rotation> reachable(IPlayerContext ctx, BlockPos pos, double blockReachDistance, boolean wouldSneak) {
+        // Prevent BetterBlockPos from leaking into Minecraft's block entity map
+        if (pos instanceof BetterBlockPos) {
+            pos = new BlockPos(pos.getX(), pos.getY(), pos.getZ());
+        }
         if (BaritoneAPI.getSettings().remainWithExistingLookDirection.value && ctx.isLookingAt(pos)) {
             /*
              * why add 0.0001?
--- a/src/api/java/baritone/api/utils/VecUtils.java
+++ b/src/api/java/baritone/api/utils/VecUtils.java
@@ -43,6 +43,10 @@ public final class VecUtils {
      * @see #getBlockPosCenter(BlockPos)
      */
     public static Vec3 calculateBlockCenter(Level world, BlockPos pos) {
+        // Prevent BetterBlockPos from leaking into Minecraft's block entity map
+        if (pos instanceof BetterBlockPos) {
+            pos = new BlockPos(pos.getX(), pos.getY(), pos.getZ());
+        }
         BlockState b = world.getBlockState(pos);
         VoxelShape shape = b.getCollisionShape(world, pos);
         if (shape.isEmpty()) {


ZacSharp (approved):
Looks good.

Feel like #4501 would be more correct, but it's been sitting long enough, so might as well try to get this one in.

ZacSharp:
Me is stupid, and did not notice you were pring this against 1.21.11. The dev branch is still 1.19.4 and will be 1.21.4 (probably once Unimined 1.4.2 releases) so you should target this at either of those branches (preferably 1.19.4, if possible).

mankool0:
Sounds good, I'll make a new pr for 1.19.4. I just looked and saw that this was the most recently updated one, and thought it was the correct one.

mankool0:
Closing in favor of the `1.19.4` branch PR #5006

PR #5015, opened 2026-04-28
Elytra in overworld (pt2)

### LINKED ISSUE #4187: Add support for auto-flight with Elytras in the overworld and the End.

underscore-zi:
Continues off the work in #4698 
Closes #4187


underscore-zi:
There are a few new commits since the last update in the old PR. Still addressing that same lag issues I was encountering. I managed to get a local reproducer and trace it down in more detail.

Fix is split across three commits:

1. Skipping the tick instead of blocking if the nether-pathfinder read lock isn't available. Not the nicest solution but there isn't a clean way I could think of to do any work without read access. 
2. Searching for a landing spot will split its work across multiple ticks instead of blocking. This was an especially noticeable issue when flying over lava pools or oceans and chunk loading was slow.
3. Not blocking while hoping async solveAngles call will complete. There was 250ms timeout (I added that to prevent deadlocking) on the `get` for that future. In the worst case scenario this would block for 250ms, fail, and then repeat the work synchronously. 

The third point is the one that I feel the least confident in as a solution. I've tested it, and not noticed any new issues, in-fact it feels a bit smoother during some tighter flights though that could just be in my head.


[0] Sample of the choppy flight experience: 

https://github.com/user-attachments/assets/d79301db-322b-49b8-acd0-865b73303d31

[1] Fixed/smooth flight:

https://github.com/user-attachments/assets/c6653048-5caf-4212-9328-7c377fae0844

LHM056awa:
Nice feature!

PR #5016, opened 2026-04-28
feat: Add farmUsingSelection setting and implementation

jayc331:
as the title says


meqativ:
why not add it as a subcommand to #sel
e.g. #sel farm

jayc331:
@meqativ that didn't seem appropriate as sel subcommands are more for operations on the selection itself

PR #5042, opened 2026-05-30
Add BranchMineProcess for automated branch mining

### LINKED ISSUE #773: Branch mine
Wow this literally hasn't been suggested before? tf

## Describe your suggestion
Okay so legitmine was made with the principle that it would make you look more legit to admins.
In all honesty that's a load of baloney. If I'm watching a player and they xray towards a cave, and take specific directions to go to a small block of diamond that was outcropped, that's sus af.

Branchmine would do literally that: branchmine. Dig down to Y=11, dig a tunnel - trace back, did that way. Then come back in the middle, move 3 blocks to the left, repeat. Get any ores that crop up in the middle. Maybe even fill in the tunnels with a single block at the hallway to keep creepers from coming back in at you.

image

## Settings
Distances the branches go, space between branches, what Y level, how wide the "main shaft" should be.

## Context

You're not going to find all of the diamonds everywhere. You're not expected to - that's sus. Your goal is to mine like a legit human so that when people ask you where you got 2 stacks of diamonds, you can show them the mines under your base and the miles of branch-mining. 

## Final checklist

### LINKED ISSUE #3393: Strip mining feature for 1.18.2
Hello I noticed there is no way to mine with baritone that's not cheating..... or even hits the grey area between cheating and not, So looking online i found this video that almost does it , https://www.youtube.com/watch?v=WXjP6KEOhS0&t=520s and trying those settings today results in https://www.youtube.com/watch?v=yzxN5osLiW4&t=7s
 now we used to strip mine by going 2 blocks and mining that section out as far as you can reach up, left, and right

and it would be great to expand legit miner to do this, but staying on that center value like tunneling but getting every ore, so we can do something like #stripmine [xyz value or pulls from ~~~] [all (all ores), or by specific Minecraft id]

### LINKED ISSUE #3759: strip mining command/mode
## Describe your suggestion
When strip mining mode is activated: the bot should start `#tunnel`ing. if an ore is found, mine the whole vein then continue tunneling. if the tunnel can't be continued for some reason, go a couple blocks to the side and start tunneling backwards

## Settings
a list of ores should either be specified via the command like `#mine` (e.g. `#stripmine diamond_ore`) or via a setting
`stripMineYLevel`: if set, stripmine will first drop to this Y level before beginning mining
`stripMineOffset`: the amount of blocks between each strip. should default to 2 for human-like mining though a larger distance might be better, hence the setting
`stripMineLength`: if set, baritone will turn around after mining this many blocks instead of just going straight forever
image
`stripMineDistance`: if set, maximum distance in blocks an ore has to be from the strip mine before baritone will mine it. set to 1 for legit human mine, any higher values basically use x-ray but still look legit if an admin is looking at your mines in spectator mode

## Context
for servers that have anti-xray, strip mining is much more efficient (and safer) than `#mine`, but having to pause every time I want to is mildly annoying enough and is a common enough usage I think its worth it
`#tunnel` already does the work of avoiding lava and being smart
related to #3393 and #773

## Final checklist

ThomasAusHH:
## Summary
Adds a fully automated branch mining process with serpentine pattern generation, multi-layer support, and integrated ore vein collection.

## New Command
- `#branchmine` (alias: `#bm`)

Usage with selection:
#sel pos1
#sel pos2
#branchmine

Usage with coordinates:
#branchmine x1 y1 z1 x2 y2 z2

## Configuration
All settings start with #set:

- branchMineInterval (default: 3) - Distance between parallel lanes
- branchMineTunnelHeight (default: 2) - Height of the mined tunnel
- branchMineOreCollection (default: true) - Collect exposed ores automatically
- branchMineScanInterval (default: 20) - Ticks between ore scans
- branchMineStaircase (default: true) - Descend via staircase to next layer
- branchMineLayerHeight (default: 3) - Vertical distance between layers
- branchMineStaggerOffset (default: 1) - Stagger offset for multi-layer patterns

Example: #branchMineInterval 5

## Features
- Serpentine pattern: Mines forward, steps sideways, mines back
- Ore vein mining: Scans 4-block radius after breaking an ore for adjacent ores
- Active block breaking: Uses LookBehavior + tool switching + left-click
- Safe ore timeout: Gives up on unreachable ores after 6 seconds

Fixes #3759
Fixes #773
Fixes #3393

### DIFF
--- a/src/api/java/baritone/api/IBaritone.java
+++ b/src/api/java/baritone/api/IBaritone.java
@@ -95,6 +95,12 @@ public interface IBaritone {
     IElytraProcess getElytraProcess();
 
     /**
+     * @return The {@link IBranchMineProcess} instance
+     * @see IBranchMineProcess
+     */
+    IBranchMineProcess getBranchMineProcess();
+
+    /**
      * @return The {@link IWorldProvider} instance
      * @see IWorldProvider
      */
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -1553,6 +1553,41 @@ public final class Settings {
     public final Setting<Boolean> allowWalkOnMagmaBlocks = new Setting<>(false);
 
     /**
+     * Distance between branch tunnels in branch mine mode
+     */
+    public final Setting<Integer> branchMineInterval = new Setting<>(3);
+
+    /**
+     * Height of branch mine tunnels (1 = 1x1, 2 = 1x2)
+     */
+    public final Setting<Integer> branchMineTunnelHeight = new Setting<>(2);
+
+    /**
+     * Collect visible ores while branch mining
+     */
+    public final Setting<Boolean> branchMineOreCollection = new Setting<>(true);
+
+    /**
+     * Ticks between ore scans during branch mining
+     */
+    public final Setting<Integer> branchMineScanInterval = new Setting<>(20);
+
+    /**
+     * Build a staircase between layers in branch mine mode
+     */
+    public final Setting<Boolean> branchMineStaircase = new Setting<>(true);
+
+    /**
+     * Vertical distance between branch mine layers
+     */
+    public final Setting<Integer> branchMineLayerHeight = new Setting<>(2);
+
+    /**
+     * Branch offset per layer for maximum coverage (staggering)
+     */
+    public final Setting<Integer> branchMineStaggerOffset = new Setting<>(1);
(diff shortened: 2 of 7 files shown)


Samemantlt:
I just tried. It prettty works!

But I have an issue. When it mine first forward tunnel it ignores ores, but after first turn it mines it.
May be its because it marks all blocks of first forward tunnel as red outlined, so ore detection is in the end of the queue.

PR #5043, opened 2026-06-02
Add configurable humanized look rotations

quatrecentdouze:
## Summary

This PR adds a configurable humanized look rotation system directly on Baritone.

Instead of snapping instantly to every target rotation, Baritone can now smooth camera movement in a more natural way while preserving the logical rotations needed for pathing, mining, placing, farming, and elytra movement.

## What Changed

- Added new `humanizeLook` settings to control look speed, acceleration, jitter, overshoot, elytra speed, and movement posture.
- Extracted the old aim processing logic into a dedicated `AimProcessor`.
- Added `HumanizedRotationSmoother` for stateful, testable rotation smoothing.
- Added `VisualRotationHelper` for visual-only camera interpolation and movement posture.
- Refactored `LookBehavior` to track:
  - effective/server rotations used for Baritone logic
  - visible camera rotations shown to the player
- Updated reachability probing to use `peekRotationForReachability`, so blocks are not considered unreachable just because the humanized rotation has not arrived yet.
- Added unit tests for the smoother and visual rotation helper.

## Important Behavior

Block interactions use humanized effective rotations, so mining and placing wait until Baritone is actually aimed correctly.

Movement/pathfinder rotations intentionally remain immediate and unhumanized. This is important because delaying movement yaw can make pathing unstable or cause circular movement. For normal movement, only the visible camera is smoothed toward the path heading.

This keeps path execution reliable while making Baritone-controlled look behavior feel much less robotic.

### DIFF
--- a/.gitmessage
+++ b/.gitmessage
@@ -1,5 +1,7 @@
 <emoji> <title> (<ticket>)
 
+# trigger workfow
+
 # 📝 Update README.md (WD-1234)
 # ✅ Add unit test for inputs (WD-1234)
 
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -782,6 +782,71 @@ public final class Settings {
     public final Setting<Integer> smoothLookTicks = new Setting<>(5);
 
     /**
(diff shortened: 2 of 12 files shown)


Camille1358:
when it will be added ?

PR #5047, opened 2026-06-07
Fix elytra hitbox collapse when motion vector is negative

### LINKED ISSUE #5044: Baritone error in Elytra Behaviour
## Some information 
Operating system: Windows
Java version: Java21
Minecraft version: 1.21.11
Baritone version: Baritone-1.17.0
Other mods: 

## Exception, error or logs
I find a bug(maybe) in ElytraBehaviour.java
https://github.com/cabaletta/baritone/blob/775f4ca97f64bba5780f0f012485dce20b36fb44/src/main/java/baritone/process/elytra/ElytraBehavior.java#L1193
I think hitbox should move(motion), not inflate(motion)
the box will be empty if moving towards negative axises
this error may  led to a crash into walls/lava

## How to reproduce
.

## Final checklist

CompileRider:
AABB.inflate(dx,dy,dz) adds dx to BOTH minX and maxX. When motion components are negative, this collapses or inverts the bounding box, causing false collision detection with walls and lava during elytra flight.

Use AABB.move(dx,dy,dz) instead, which translates the entire box without distorting its dimensions. The .inflate(0.01) safety padding is preserved.

Fixes #5044

UwU


leijurv:
yeah sure ill merge this just delete that test

ZacSharp:
Are speeds above one block per tick a problem or do Minecraft physics not check for collisions between the current and next positions? This check now assumes the player can move through lava without touching it by simply moving fast enough.

CompileRider:
> Are speeds above one block per tick a problem or do Minecraft physics not check for collisions between the current and next positions? This check now assumes the player can move through lava without touching it by simply moving fast enough.


You are right. While AABB.move(dx, dy, dz) fixes the bounding box collapse/inversion when motion is negative, it indeed introduces a 'tunneling' problem. It only checks the hitbox at the destination, meaning at high speeds (common with Elytra), the bot could fly through thin walls or lava streams without detecting them.

The original use of inflate() was likely an attempt to create a 'swept volume' check, but inflate() is bidirectional and fails with negative vectors. The correct fix should be using AABB.expandTowards(motion.x, motion.y, motion.z). This method expands the box specifically in the direction of movement, correctly covering the entire path for that tick without the risk of collapsing the box or skipping intermediate obstacles.

CompileRider:
Since this PR is already merged, it would be best to address this 'tunneling' improvement in a new issue. I'll open one to track the implementation of expandTowards to ensure full collision coverage at high speeds while maintaining the fix for negative motion vectors.

PR #5054, opened 2026-06-14
Add support for nether vines v2

ZacSharp:
Directly based on https://github.com/cabaletta/baritone/pull/4111. This is a crazy long lived branch, but somehow still merges cleanly.

* Most movements explicitly refuse to start on normal vines or ladders. Now they also refuse to start on nether vines.
* `MovementFall` now knows that nether vines reset fall damage, so Baritone can climb down on them
* `MovementPillar` and `MovementTraverse` now press the jump button instead of walking into an adjacent block
  * Traversing on vines/ladders is much less janky
  * Climbing up unsupported vines is possible
  * It can climb up mixed vines/ladders without getting stuck against the bottom face of a ladder
* `allowVines` now applies to nether vines as well, including the unreliability

Would really like if someone else could also try breaking this. I don't trust myself to find the edge case regressions I missed.

### DIFF
--- a/src/main/java/baritone/pathing/movement/MovementHelper.java
+++ b/src/main/java/baritone/pathing/movement/MovementHelper.java
@@ -522,7 +522,7 @@ public interface MovementHelper extends ActionCosts, Helper {
      */
     static boolean mustBeSolidToWalkOn(CalculationContext context, int x, int y, int z, BlockState state) {
         Block block = state.getBlock();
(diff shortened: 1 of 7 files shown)


ZacSharp:
> "scorbett123 and others added 8 commits 3 years ago"

That's one commit created three years ago and seven created today. GitHub sure knows how to not be misleading.

PR #5082, opened 2026-07-29
Update to 26.2 and stabilize mining aim

N-Echelibe:
This updates Baritone to the 26.2 development baseline and includes a fix for legit mining aim.

The mining rotation previously used roughly fixed angular offsets. The farther away a block is, the more that same angle moves the crosshair across the target, so at longer reach distances or awkward viewing angles it could drift off the block. The adjustment is now scaled by distance so the crosshair moves by roughly the same amount in world space regardless of how far away the block is.

It also keeps the current block target stable while that block is actively being broken. Previously, target reevaluation could switch blocks mid-break, repeatedly reset progress, and leave mining stuck until the player intervened.

The fix is intentionally lightweight. Distance scaling is a constant-time calculation, and the target is only held during an active break; it does not add any new scans or pathfinding work.

Testing: added focused coverage for distance-scaled aim behavior and completed the project tests and Fabric build.


ZacSharp:
This never left draft status, there are conflicts and we have a release for 26.2, closing.
(This is also 10x more changes than our 26.2 port, so something is clearly wrong here)

Issue #4972, opened 2026-02-20
Baritone breaks blocks that i blocked.


wojtekszzzz6-star:
## Some information
Operating system: windows 10
Java version: 21.0.7
Minecraft version: 1.21.11
Baritone version: I dont know
Other mods (if used): Sodium

## Exception, error or logs
Please find your `latest.log` or `debug.log` in this folder and attach it to the issue

[latest.log](https://github.com/user-attachments/files/25452987/latest.log)

## How to reproduce
I set sand as block to disallow and avoid breaking, I used #sel set stone_bricks. It started mining sand that was in way.

## Modified settings
To get the modified settings run `#modified` in game

allowPlace True
allowInventory true
blockBreakAdditionalPenalty 100000.0
blocksToAvoid sand
blocksToDisallowBreaking sand
blocksToAvoidBreaking crafting_table,furnace,chest,trapped_chest,sand,shulker_box
buildSkipBlocks sand
rightClickSpeed 0
randomLooking113 0.0
randomLooking 0.0
primaryTimeoutMS 250
buildInLayers true
placeIncorrectBlockPenaltyMultiplier 10000.0

## Final checklist
- [x] I know how to properly use check boxes 
- [x ] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [ x] I have included logs, exceptions and / or steps to reproduce the issue.
- [ x] I have not used any OwO's or UwU's in this issue.

wojtekszzzz6-star:
BARITONE VERSION: null

ZacSharp:
`blocksToDisallowBreaking` (and `allowBreak`, `allowPlace`, etc.) affect what Baritone does while walking, not what the builder (or farming) does as part of completing its job.
If you want to stop the builder from replacing some blocks with air you can use `buildIgnoreBlocks`, but iirc that will not stop Baritone from replacing those blocks with something else. There might be a setting to completely stop the builder from breaking some blocks, but I could not find it and also do not remember it existing.

Issue #4812, opened 2025-07-10
 Custom blocksToAvoid penalty


radicore:
## Custom blocksToAvoid penalty
Being able to customise a blocks penalty value as an extension to blocksToAvoid

## Settings
- You can add a block penalty by specifying blocksToAvoid [block] [number]
- Default / max would be COST_INF by not specifying a number
- Min value would be 1.000 which would remove the block from the blocksToAvoid list

## Context
Pathing through water would have a higher cost because it is slower (as baritone hovers rather than swims through water) than walking on land.

Having a customisable penalty means it could walk on that block to reach its goal as a last resort instead of failing on the spot. (example being simply adding it to blocksToAvoid would not find a path if that was the only option) 

Another one could be setting it to preferring soul sand if wearing soul speed boots

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have not used any OwO's or UwU's in this issue.

PR #5103, opened 2026-08-27
Make pathing respect blocksToAvoid when breaking blocks

### LINKED ISSUE #4675: Tunnel command doesnt respect blocks to avoid
## How to reproduce
#blocksToAvoid cave_air
then #tunnel

the tunnel will not avoid cave_air

Image

## Modified settings
Image

cryst6lkush:
Fixes #4675
Related: #4972 (same root cause, closed without a fix), #4812 (per-block penalty proposal — partially addressed by the new toggle)

## Problem

`blocksToAvoid` was only consulted in `MovementHelper.canWalkThroughBlockState`, i.e. when deciding whether Baritone is willing to *walk through* a block without mining it first. Any goal that digs — most visibly `#tunnel` — works by *breaking* every block in its way, and `getMiningDurationTicks` never looked at the setting, so avoided blocks were broken at normal cost.

For the reporter's exact repro (`#blocksToAvoid cave_air`) the check was even unreachable while walking: `cave_air` extends `AirBlock`, and the `AirBlock -> YES` early return fired before the avoid check.

## Changes

- **`MovementHelper.canWalkThroughBlockState`** — moved the `blocksToAvoid` check above the `AirBlock` early return, so air-family blocks like `cave_air` are actually treated as not walkable-through when avoided.
- **`MovementHelper.getMiningDurationTicks`** — breaking a block in `blocksToAvoid` now goes through the new `blocksToAvoidBreakPenalty` setting:
  - `true` (default): break cost of avoided blocks is multiplied by 10, so A* strongly prefers routes that don't require breaking them but still succeeds when unavoidable.
  - `false`: strict mode — breaking an avoided block costs `COST_INF`, so Baritone paths around or fails instead, same semantics as `blocksToDisallowBreaking`.
- **`MovementHelper.fullyPassableBlockState`** — replaced the hardcoded `Blocks.TRIPWIRE` (added by #4885) with the setting check, so parkour moves and path-execution sanity checks stay consistent with walkability.
- **`Settings.blocksToAvoidBreakPenalty`** (new boolean, default `true`) with javadoc covering both modes and their caveats.

## Behavior notes

- Default settings are unchanged except that paths which would break an avoided block now pay a 10x cost multiplier; tripwire handling matches previous behavior exactly since tripwire is the only default entry.
- `#tunnel` without arguments uses `GoalStrictDirection`, which offers no lateral alternative — in penalty mode tunneling through a cheap-to-break avoided block may still happen (with reduced preference); set `blocksToAvoidBreakPenalty false` to refuse outright.
- In strict mode `#mine` ignores targets whose block is in `blocksToAvoid` (documented in the setting javadoc).

## Testing

- `./gradlew compileJava` and `./gradlew test` pass (41 tests, 0 failures) on JDK 17 / Gradle 8.2.
- In-game verification: `#blocksToAvoid cave_air` + `#tunnel` now avoids cave air where an alternative path exists and respects the strict/penalty modes as described above; plain `#tunnel` through normal terrain unchanged.

### DIFF
--- a/src/api/java/baritone/api/Settings.java
+++ b/src/api/java/baritone/api/Settings.java
@@ -264,6 +264,17 @@ public final class Settings {
     public final Setting<Double> avoidBreakingMultiplier = new Setting<>(.1);
 
     /**
+     * How Baritone treats breaking a block in {@link #blocksToAvoid}:
+     * true = pay a large break cost penalty, pathing around the block when possible but breaking it if unavoidable;
+     * false = refuse to break it entirely (infinite cost), like {@link #blocksToDisallowBreaking}.
+     * <p>
+     * Note that a goal which forces Baritone in a straight line (e.g. {@code #tunnel}) may have no path around an
+     * avoided block, in which case even the penalty mode will break it. Strict mode also makes {@code #mine} ignore
+     * targets whose block is in {@link #blocksToAvoid}.
+     */
+    public final Setting<Boolean> blocksToAvoidBreakPenalty = new Setting<>(true);
+
+    /**
      * A list of blocks to be treated as if they're air.
      * <p>
      * If a schematic asks for air at a certain position, and that position currently contains a block on this list, it will be treated as correct.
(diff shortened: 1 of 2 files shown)


ZacSharp:
There is `blocksToAvoidBreaking` to penalize breaking specific blocks and `blocksToDisallowBreaking` to completely stop path execution from breaking specific blocks; neither is the purpose of `blocksToAvoid`.

PR #5105, opened 2026-08-27
Add optional head hitter sprint jumping

AverWasTaken:
New `headHitters` setting, off by default like other optional behavior settings.

When enabled, Baritone sprint jumps in 1x2 corridors and under any low ceiling on flat traverses. The sprint jump speed boost applies on the first ticks of the jump before we bonk our head, so hopping through a tunnel is faster than plain sprinting.

Implementation is entirely in `PathExecutor.shouldSprintNextTick` since that's where the sprint policy already lives. It forces `Input.JUMP` the same way the traverse to straight ascend skip does, and because it runs after `movement.update()` reasserts the inputs, the forced jump only lasts one tick so there's no state to clean up.

Some guards it has, mostly found by breaking it in game:
* it waits until we're fully inside the corridor (same idea as `skipNow`) so we don't jump a block early and bonk on the face of the entrance block, which just stops us
* no headhitting while sneaking (magma edge safety) or while the movement still has blocks to break (breaking is like 5x slower when you're jumping)
* no headhitting when the path goes down right after the movement, or when the two blocks past the destination have no floor. The bonk momentum carries an extra block or two and will happily throw us off a ledge next to the goal
* that ledge check originally had an off by one that crashed the client when the current movement was the last on the path, also gone now

This is execution only, the cost model is untouched (same as `sprintAscends` and `overshootTraverse`), so path choice doesn't change, we just arrive a bit sooner.

Tested in a dev env on fabric: tunnels, overhangs, goals next to dropoffs.

### DIFF
--- a/src/main/java/baritone/pathing/path/PathExecutor.java
+++ b/src/main/java/baritone/pathing/path/PathExecutor.java
@@ -373,6 +373,9 @@ public class PathExecutor implements IPathExecutor, Helper {
 
         // if the movement requested sprinting, then we're done
         if (requested) {
+            if (Baritone.settings().headHitters.value) {
+                headHitJump(current);
+            }
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/baritone/pathing/path/PathExecutor.java
```diff
@@ -487,6 +490,61 @@ private boolean shouldSprintNextTick() {
...
+        }
+        return true;
+    }
+
+    private boolean clearOfLedgesAhead(IMovement current, BlockPos dir) {
+        // momentum from the head bonk can carry us an extra block or two, so don't headhit when there's a ledge or
+        // dropoff ahead, e.g. when the goal is right next to one
+        if (pathPosition < path.length() - 2 && path.movements().get(pathPosition + 1).getDirection().getY() < 0) {
+            return false; // the path goes down right after this movement, don't add any momentum
+        }
```
ZacSharp: Could potentially make this stricter and require that the next two blocks in the current direction must also be part of the path, so momentum never sends us off path.

src/main/java/baritone/pathing/path/PathExecutor.java
```diff
@@ -487,6 +490,61 @@ private boolean shouldSprintNextTick() {
...
+            return false; // breaking is like 5x slower when you're jumping
+        }
+        // not while sneaking either, e.g. walking on magma, a jump would break the sneak and the edge safety with it
+        return !behavior.baritone.getInputOverrideHandler().isInputForcedDown(Input.SNEAK);
+    }
+
+    private boolean underHeadBonkCeiling(BlockPos dir, BetterBlockPos feet) {
+        if (MovementHelper.fullyPassable(ctx, feet.above(2))) {
+            return false; // not under a ceiling yet; jumping now would bonk on the face of the ceiling block ahead and stop us
+        }
```
ZacSharp: Things like trap doors are not fully passable, yet you cannot (always) head bonk against them. Maybe check whether the block to bonk against is a normal cube (iirc `MovementHelper` also checks that somewhere, might be you can steal that).

ZacSharp (commented):
Still have to try it in game, so comments are purely based on predicted behavior after reading the code.

src/main/java/baritone/pathing/path/PathExecutor.java
AverWasTaken: fixed in 5b6c3cf7, clearOfLedgesAhead now requires the next two movements to go in the same direction, so the blocks momentum would take us to are always part of the path. that also covers the path turning or ending there, and the floor checks got dropped since the path already guarantees those blocks are walkable

src/main/java/baritone/pathing/path/PathExecutor.java
AverWasTaken: also fixed in 5b6c3cf7, the ceiling now has to be a normal cube (stole isBlockNormalCube from MovementHelper like you suggested) so trapdoors and friends won't trigger it anymore

src/main/java/baritone/pathing/path/PathExecutor.java
ZacSharp: Path positions do not need to be supported if the corresponding movement would place a block or doesn't need support (traverse likes to do that kind of thing with throwaways, ladders, vines, frostwalker), but I failed to actually make it fall off, so fine by me.

ZacSharp (approved):
Sorry for taking so long for a <15 minute review. I'm happy now.

ZacSharp:
Oh also that force push makes me happy :)

AverWasTaken:
glad u like it mwah

AverWasTaken:
@leijurv lgtm?

AverWasTaken:
@ZacSharp cbd182b says the main dev branch is 1.21.4 now, so I've ported this over there. Same changes, just ported (only real tweak was isOnGround() → onGround() with the new mappings). Also fixed your last comment inline: clearOfLedgesAhead now bails when the support under either of the next two positions isn't physically there yet (throwaway placement, frostwalker ice that hasn't frozen yet, ladders/vines that have no floor at all), since bonk momentum can arrive before the support does. Couldn't make it fall off myself either, but now it's guaranteed either way.

Renaming the head branch for the move closed this PR as a side effect and GitHub doesn't let it reopen, so the port lives on in #5120. plz re review <3

ZacSharp:
Whoops, my bad on not noticing that. Thanks for fixing.

New pr looks good and leijurv was faster anyway, so no review needed.

PR #5112, opened 2026-09-09
Add biome and structure locate commands for 26.2

Dvow:
Add biome and structure locate commands for 26.2

Adds commands to find biomes and structures beyond loaded chunks and travel to them:

```text
#locate biome cherry_grove
#locate structure end city
#locate structure village
#locate structure stronghold 10000
```

Singleplayer uses the world’s seed and generator automatically. Multiplayer requires a known seed set with `#locate seed <seed>`. Numeric and text seeds follow Minecraft’s parsing rules, and settings are saved per world and dimension.

Structure searches use Minecraft’s placement rules and structure generation, including biome checks, weighted variants and stronghold rings. Structure IDs, tags and spaced names are supported. Both search types stay in the current dimension.

Searches run in the background and cancel on stop, replacement or world changes. Before the final approach, the process checks loaded terrain for standing or swimming positions. Biome arrival requires the actual biome to match; structure arrival is reported as reaching the predicted area.

Small biomes can fall between samples, and older chunks or custom generation can differ from predictions. Finding an End City does not guarantee a ship or a reachable route.

Includes completion, usage docs and regression tests. Build changes remove an unused Parchment repository and support Java 25 installations without `jmods` during ProGuard packaging.

### Testing

- Fabric build and all 79 tests passed on Java 25.
- Structure predictions were compared against Minecraft’s own generation for End Cities, villages, strongholds, fortresses and bastions.
- Tests cover seed parsing, settings, search bounds, cancellation, world changes and destination checks.
- Both biome and structure locating and travel were tested successfully in-game on Fabric.

Other loaders have not been validated.


ZacSharp:
I do not thing this is a good idea. This feature simply does not work on many servers - specifically 2b2t has custom terrain generation - and in singleplayer it is just an overly convoluted and bug prone way of running the built in `/locate` command. This also depends strongly on internals of Minecrafts resource management and world generation, which is bad. Doing similar hacks to load loot tables has lead to plenty of problems in the past.

Besides that, there clearly is some LLM overengineering and some unrelated changes in this pr.

PR #5115, opened 2026-09-11
Update Baritone version in README

Soyodk:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Baritone
+# Baritone 1.21.11
 <p align="center">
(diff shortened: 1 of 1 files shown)


ZacSharp:
I don't think we need the Minecraft version number in the readme header, but I'm also not quite sure whether that really is the reason you opened this pull request.
If you wanted to trigger pull request checks to get a CI run you probably noticed already that those do not run for first-time contributors. You can enable Actions in your own repository and trigger a CI run by pushing a dummy commit just to your own repository. Though currently the 1.21.11 release is also up to date, so you can just use that.

PR #5116, opened 2026-09-11
Stop worker threads from keeping the game running after exit

kekkykekkek:
The worker pool used non-daemon threads and the chunk packer and periodic cache save never finish, so the process stayed alive after Minecraft shut down until the client shutdown watchdog killed it with a crash report.

Use daemon threads, and stop the pool in a shutdown hook that waits briefly for running tasks such as cache saves.


leijurv:
#5123

Issue #31, opened 2018-08-17
Automatically disable auto-jump while pathing is active


leijurv:
We jump ourselves, and it messes up MovementDiagonal edging around blocks (it jumps on top because it thinks we want to, for example if we were edging around magma this would cause us to take damage and lose the path, and continue to take damage while recalculating)

HowardStark:
Fixed by commit d86dc62.

PR #5117, opened 2026-09-12
Replace the native nether-pathfinder with a Java port

### LINKED ISSUE #4494: Add a option to allow unsupported cpus to use efly
## Describe your suggestion
Add a option to use elytra fly with unsupported cpus.

## Settings
The command being used twice to overide the message about unsuported cpu.

## Context
This would allow a wider amount of people to use this feature

## Final checklist

### LINKED ISSUE #5046: Unexpected exit during nether pathng
## Some information
Operating system: Windows
Java version: Java21
Minecraft version: 1.21.11
Baritone version: baritone-1.15
Other mods: My private mode(which modify player's boundingBox during elytra fly)

## Exception, error or logs
```
[02:04:25] [pool-10-thread-4/INFO] (Minecraft) [STDOUT]: Loaded region successfully in 835ms
raytrace whiffed
```
then game exit without any mc crash report

## How to reproduce
Can not reproduce stably

## Final checklist

0Mattias:
Fifteen commits: the three below, then the changes from the reviews, one commit per comment.

1. `baritone.process.elytra.pathfinder`: the whole library in plain Java, one class per C++ file, with its tests and oracle files (what the native library answered for 625 generated chunks and 4000 rays; the port reproduces every one, hit positions to the bit). 34 tests, junit 4 like the rest.
2. The seam. `NetherPathfinderContext` gets a `Chunk` where it had a pointer, `BlockStateOctreeInterface` caches one, and the dependency, the babbaj maven repo, the fabric `include`, the forge/neoforge `shadowCommon`, the proguard keep rule and the natives for each platform go. Every system is supported now, so `NullElytraProcess` and the check in `ElytraProcess.create` go too. `elytraCustomAllocator` stays as a deprecated no-op so settings files still load.
3. A fix to the region reader. Baritone writes a cached chunk's two bits per block with `BitSet.toByteArray()`, lowest bit first; the native reader took each byte's bits from the top down (`baritone.cpp`, `get2Bits`), so every aligned run of four blocks along x came back mirrored in any chunk a search took from the region cache. The port had copied it; now it reads what baritone writes, with a test built the way `CachedChunk` builds a chunk.

Since the first push, from babbaj's and ZacSharp's reviews: the FFI leftovers are gone (arguments packed into longs, the port's own `BlockPos` and `Face`, the raytrace wrappers, the duplicate region magic), `Raytracer` returns the hit position instead of a boolean, the block update asks `isAir()`, the x8 summary is a `byte[]` per section and stays exact on every clear (the `exact` flag is gone, no whole-chunk writer clears a block), `Chunk` says what keeps its readers and writers apart instead of describing a race, a chunk's origin is a `final boolean fromCaller`, `insertChunkData` and `setChunkState` are deleted, `NodePos` holds three ints, the set of generated chunks is a fastutil `LongOpenHashSet`, and the neighbour order is kept as the native library had it so the A* search expands the same way.

Left alone on purpose: the read-write lock and the two executors. The port doesn't need them (chunks are objects, the table is a `ConcurrentHashMap`, lookups, inserts, culls, searches and rays can all run at once), but dropping them is its own change.

Three inputs the native library couldn't take are handled now, because the elytra solver does send them: a zero-length ray (native: `raytrace whiffed`, `exit(696969)`), a NaN or infinite coordinate (native hangs on an infinity in z; the port throws `IllegalArgumentException`), and a ray ending exactly on a voxel corner, which every ray at a path node is (native could step into a node the ray never enters and exit; the port terminates, tested).

Same machine and inputs as the tables on #31:

| | native | java |
|---|---|---|
| 200k rays through real terrain, 30-150 blocks | 0.75 us/ray | 0.56-0.67 us/ray |
| `pathFind` over 800 blocks, 20 runs | 0.39 ms avg | 0.43-0.45 ms avg |
| main.cpp's 100k block generating search | 3.0 s | 2.9-3.1 s |

Tried for real in my fabric fork (0Mattias/cheesecake#19, same seam): unit tests plus four in-world elytra flights, overworld above the build limit, auto-jump, nether below and above the roof, all pass with the port. Here: `./gradlew build` on JDK 21, all four loaders, tests included.


src/api/java/baritone/api/Settings.java
```diff
@@ -1550,8 +1550,10 @@ public final class Settings {
     public final Setting<Boolean> elytraChatSpam = new Setting<>(false);
 
     /**
-     * May reduce memory usage by using a custom allocator for pathfinding
+     * Does nothing. The pathfinder is Java now and has no allocator of its own; kept so that a
+     * settings file naming it still loads.
      */
+    @Deprecated
     public final Setting<Boolean> elytraCustomAllocator = new Setting<>(true);
```
babbaj: I think this one can be safely removed

src/main/java/baritone/process/elytra/pathfinder/BaritoneRegion.java
```diff
@@ -0,0 +1,125 @@
...
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.zip.GZIPInputStream;
+
+/** Reads the chunks of one of Baritone's cached region files ({@code r.X.Z.bcr}). */
+final class BaritoneRegion {
+
+    private static final int MAGIC = 456022911;
```
babbaj: make baritone.cache.CachedRegion.CACHED_REGION_MAGIC public and reference that instead of having this duplicate field

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+    // How the raytracer will treat chunks that aren't actually observed.
+    public static final int CACHE_MISS_GENERATE = 0;
+    public static final int CACHE_MISS_AIR = 1;
+    public static final int CACHE_MISS_SOLID = 2;
+
+    public static final int DIMENSION_OVERWORLD = 0;
+    public static final int DIMENSION_NETHER = 1;
+    public static final int DIMENSION_END = 2;
+
+    static final int STATE_FROM_JAVA = 0;
```
babbaj: "from java" is now a silly name now that both sides are written in java.  maybe "from caller" would be better? im not really sure what to call this.

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+ * the chunks it has been given or has generated, and a search that can be cancelled.
+ * <p>
+ * The method names are the ones the native API had, with objects where it had pointers: a chunk
+ * is a {@link Chunk} that stays valid for as long as it is referenced, and there is nothing to free.
+ * Lookups, inserts, the search and rays may all run at the same time from different threads.
+ */
+public final class NetherPathfinder implements AutoCloseable {
+
+    // How the raytracer will treat chunks that aren't actually observed.
+    public static final int CACHE_MISS_GENERATE = 0;
```
babbaj: these can just be an enum now

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+    public static final int CACHE_MISS_AIR = 1;
+    public static final int CACHE_MISS_SOLID = 2;
+
+    public static final int DIMENSION_OVERWORLD = 0;
```
babbaj: same as above

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+        this.cancelFlag.set(false);
+        final Size size = atLeastX4 ? Size.X4 : Size.X2;
+        final NodePos start = PathFinder.findAir(this, size, new BlockPos(x1, y1, z1), defaultAirElseGenerate);
+        final NodePos goal = PathFinder.findAir(this, size, new BlockPos(x2, y2, z2), defaultAirElseGenerate);
+        final PathFinder.Path path = PathFinder.findPathSegment(this, start, goal, atLeastX4, failTimeoutInMillis, defaultAirElseGenerate, fakeChunkCost);
+        if (path == null) {
+            return null;
+        }
+        final List<BlockPos> blocks = refine ? Raytracer.refine(this, path.blocks) : path.blocks;
+        final long[] packed = new long[blocks.size()];
```
babbaj: Packing into longs was only necessary for FFI. We can just directly use the List<BlockPos> now.

src/main/java/baritone/process/elytra/pathfinder/PathFinder.java
```diff
@@ -0,0 +1,406 @@
...
+        ctx.tryLoadRegion(startZero.chunkX(), startZero.chunkZ());
+        startNode.cost = 0;
+        startNode.combinedCost = startNode.estimatedCostToGoal;
+        s.openSet.insert(startNode);
+        ctx.getRealChunkFromCacheOrFakeChunkMaybeGen(startZero.chunkX(), startZero.chunkZ(), fakeChunkMode);
+
+        s.bestSoFar = startNode;
+        s.bestHeuristicSoFar = startNode.estimatedCostToGoal;
+
+        final long startTime = System.nanoTime();
```
babbaj: should be using currentTimeMillis

src/main/java/baritone/process/elytra/pathfinder/BlockPos.java
```diff
@@ -0,0 +1,126 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.process.elytra.pathfinder;
+
+/** An integer position. Immutable. */
+final class BlockPos {
```
babbaj: we should be using minecraft's BlockPos

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,216 @@
...
+ * scan of 64 or 8 longs.
+ * <p>
+ * Reads and writes are plain (not synchronized), as they were in the native library: a reader
+ * that races a writer may see a partly written slab, which is the same as before, and nothing
+ * worse can happen.
+ */
+public final class Chunk {
+
+    public static final int HEIGHT = 384;
+    public static final int SLABS = HEIGHT / 16;
```
babbaj: they should be called sections

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+    long tryLoadRegion(int cx, int cz) {
+        if (this.baritoneCache == null) {
+            return 0;
+        }
+        final int regionX = cx >> 5;
+        final int regionZ = cz >> 5;
+        if (!this.checkedRegions.add(key(regionX, regionZ))) {
+            return 0;
+        }
+        final long t1 = System.nanoTime();
```
babbaj: also should be currentTimeMillis

src/main/java/baritone/process/elytra/pathfinder/Raytracer.java
```diff
@@ -0,0 +1,379 @@
...
+        final int childLevel = level - 1;
+        final int childBytes = CHILD_BYTES[level];
+        int currNode = firstNode(tx0, ty0, tz0, txm, tym, tzm);
+        do {
+            final int i;
+            final double r;
+            switch (currNode) {
+                case 0:
+                    i = a;
+                    r = procSubtree(a, ox, oy, oz, targetLen, tx0, ty0, tz0, txm, tym, tzm, childLevel, nx + ((i & 4) != 0 ? half : 0), ny + ((i & 2) != 0 ? half : 0), nz + ((i & 1) != 0 ? half : 0), slab, filled, level == 1 ? off : off + i * childBytes, i);
```
babbaj: this is pretty unreadable. id like the arguments between `childLevel` and `slab` to be computed before entering the switch because they appear to always be the same and with a comment explaining them.

0Mattias:
wait realized something, fixing now

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    // ---- rays ----
+
+    /**
+     * Traces {@code inputs} segments, start[i*3..] to end[i*3..]. hitsOut[i] says whether segment i
+     * hit a solid block, and if hitPosOutCanBeNull is given, where. Points and the segments between
+     * them must have 0 <= y < 384.
+     */
+    public void raytrace(int fakeChunkMode, int inputs, double[] start, double[] end, boolean[] hitsOut, double[] hitPosOutCanBeNull) {
```
babbaj: this function can be removed now that we can just call Raytracer.raytrace directly.

the array arguments were also an ffi optimization so wherever arrays were constructed to call this can be simplified

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+            hitsOut[i] = Raytracer.raytrace(this, start[i * 3], start[i * 3 + 1], start[i * 3 + 2], end[i * 3], end[i * 3 + 1], end[i * 3 + 2], fakeChunkMode, hitPosOutCanBeNull, i * 3);
+        }
+    }
+
+    /**
+     * With anyIfTrueElseAll, the index of the first segment that is clear, else the index of the
+     * first that is blocked; -1 if there is none. So -1 means "none clear" in the first mode and
+     * "all clear" in the second.
+     */
+    public int isVisibleMulti(int fakeChunkMode, int inputs, double[] start, double[] end, boolean anyIfTrueElseAll) {
```
babbaj: for the same reasons as above this function can be removed/move to the file it was called

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,385 @@
...
+                }
+            } else if (!anyIfTrueElseAll) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    /** Whether there is line of sight between the two points. */
+    public boolean isVisible(int fakeChunkMode, double x1, double y1, double z1, double x2, double y2, double z2) {
```
babbaj: another redundant function

src/main/java/baritone/process/elytra/pathfinder/Raytracer.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /**
+     * Traces the segment from (fx, fy, fz) to (tx, ty, tz). Returns true if it hit a solid block;
+     * the hit position is then written to {@code hitOut} at {@code hitIndex}, if the array is not null.
+     * The two points must be inside 0 <= y < 384, as must be the segment between them. A segment
+     * of no length is a point, which is a hit if it is inside a block; a coordinate that is not a
+     * finite number is refused.
+     */
+    static boolean raytrace(NetherPathfinder ctx, double fx, double fy, double fz, double tx, double ty, double tz, int fakeChunkMode, double[] hitOut, int hitIndex) {
```
babbaj: instead of a boolean this should return the hit position if there's a hit, otherwise null. the last 2 arguments can be removed

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,216 @@
...
+        final int off = x2Offset(x, y, z);
+        final long mask = 1L << (((off & 7) << 3) + bitIndex(x, y, z));
+        final int x8 = x8Index(x, y, z);
+        if (solid) {
+            // the x8 is marked before its block is set, so a reader racing this write never skips a block it can see
+            this.filled[y >> 4] |= 1 << x8;
+            s[off >>> 3] |= mask;
+        } else {
+            s[off >>> 3] &= ~mask;
+            if (allZero(s, x8 * (X8_BYTES / 8), X8_BYTES / 8)) {
```
babbaj: why??

babbaj (changes requested):

0Mattias:
Thanks for the review. All of it (including your new comments) is now applied and pushed here. Quick overview of how the port fits together first, since a few of your points touch the seams, then each comment in order, then what I checked.

### How the port is put together

The Java port lives in `baritone.process.elytra.pathfinder` and follows the native library file for file. `Chunk` is the same octree layout as `Chunk.h`: 24 sections of 512 bytes, one bit per block, plus a byte per section that says which x8 cubes may hold a block. `ChunkGeneratorHell` and the two noise generators produce Nether terrain from the seed. `PathFinder` is the A* over cubes of 2 to 16 blocks, with the binary heap. `Raytracer` is the Revelles octree traversal from `Refiner.cpp`. `BaritoneRegion` reads Baritone's `.bcr` region cache. `NetherPathfinder` is the context: the chunk table, the search, and the cancel flag.

The integration is thin on purpose. `NetherPathfinderContext` keeps the methods Baritone already called (`pathFindAsync`, `raytrace`, `insertChunk`, `cullFarChunks`, etc.) but holds a `NetherPathfinder` object instead of a native pointer. `BlockStateOctreeInterface` reads chunks through it. There is no JNI and no `Unsafe`, nothing to load, so `ElytraProcess` no longer needs `isSupported` or a null process.

Correctness is checked against recorded output from the native library: chunk hashes for a generated world, 4,000 rays with their hit positions, and searches over the same terrain. Those tests did not change in this round and still match bit for bit.

### Your comments, in order

1. **`Settings.java`, `elytraCustomAllocator`.** Removed, including the deprecation stub. It only ever selected the native allocator.

2. **`BaritoneRegion`, duplicate magic number.** `CachedRegion.CACHED_REGION_MAGIC` is public now and `BaritoneRegion` uses it. The local copy is gone.

3. **"from java" naming.** Took your suggestion: `STATE_FROM_CALLER`, `hasChunkFromCaller`, and a `fromCaller` parameter on `setChunkState`. The distinction is between chunks the game handed in and chunks the port made up (generated, or assumed air). "From caller" reads fine for that.

4. **`CACHE_MISS_*` as an enum.** Now `NetherPathfinder.CacheMiss { GENERATE, AIR, SOLID }`. The range check on the int is gone with it.

5. **`DIMENSION_*` as an enum.** Now `NetherPathfinder.Dimension { OVERWORLD, NETHER, END }`. `BaritoneRegion.dimensionOf` returns the enum, or null for a directory that is not a cache, instead of `-1`.

6. **Packing into longs.** Gone. `PathSegment` holds a `List<BlockPos>`, `pathFind` returns it directly, and `UnpackedSegment.from` maps it to `BetterBlockPos`. The packing constants, `packBlockPos`, and the round-trip test for them are deleted.

7. and 10. **`currentTimeMillis`.** Done in both places: the search timeouts in `PathFinder.findPathSegment` and the region-read time in `tryLoadRegion`, which is subtracted from the timeouts and is in milliseconds too now. For the record, `nanoTime` was there because it is monotonic and `currentTimeMillis` can jump with the wall clock. At 30-second timeouts that does not matter.

8. **Minecraft's `BlockPos`.** The port's own `BlockPos` class is deleted and `net.minecraft.core.BlockPos` is used everywhere. Two small things moved: `Face` got an `offset(pos, n)` method for what the old class did with a face, and the floor is `Mth.floor`.

9. **Slabs are sections.** Renamed everywhere in `Chunk` and `Raytracer`: `SECTIONS`, `sections`, `section(y)`, `SECTION_LONGS`, the parameters, and the comments.

11. **The unreadable switch in `Raytracer`.** The three arguments every case computed the same way (the child's origin on x, y and z) plus the child's byte offset are now computed once before the switch, from the octant index `i = currNode ^ a`, with a comment explaining the reflection and what each bit of `i` means. Each case now only differs in which t-parameters it passes and which node comes next, which is the part that actually varies.

12., 13. and 14. **`NetherPathfinder.raytrace`, `isVisibleMulti`, `isVisible`.** All three removed. Callers call `Raytracer.raytrace` directly. The array versions of `raytrace(count, src, dst, ...)` on `NetherPathfinderContext` stay, because they are Baritone's own API and `ElytraBehavior` sends its hitbox rays through them in batches. They now loop and ask the raytracer one ray at a time, so the array handling stops at that seam instead of running through the port.

15. **`Raytracer.raytrace` returning a boolean with out-parameters.** It returns the hit position as a `Vec3`, or null when the ray reaches its end. The `hitOut` array and index parameters are gone. The oracle test compares against the returned position.

16. **`Chunk`, "why??"** That line cleared the x8 summary bit when the last block in that cube was cleared, and to know that it had to scan the cube's eight longs on every clear. Why it was there: the `filled` byte per section says which x8 cubes hold a block, and I had kept it exact on clears so `isEmptyX8` would be exact too. But nothing needs it exact. Every reader only uses the bit to skip a cube it knows is empty, so a set bit that outlives its blocks costs a reader one scan and can never give a wrong answer. So I dropped the clear-side maintenance. The bit is set when a block is set and reset only by `fillSection`. The javadoc now says the bit means "may hold a block", and the test checks that a clear leaves it set while the exact query sees the block gone.

### Also in this push

Two things that are not from the review. `pathFind` used to clear the cancel flag on the way in. That meant a `cancel()` arriving while a search was still queued behind the lock, which is exactly the one `destroy()` sends, got wiped, and that search then ran to its full timeout while holding the lock. It is cleared on the way out now, so a cancel is honoured by the search it was aimed at and does not leak into the next one. `CancelTest` covers it. And since visibility is answered per ray now, the batch methods answer exactly what the rays say, one at a time, instead of a batch-level shortcut.

### Verified

`./gradlew :test`: 77 tests, all passing, including the chunk hash and 4,000 ray comparisons against the native library. `./gradlew :fabric:build` is green, and the jars carry the port with no native library and no leftover `BlockPos` class. The same code is flying in my Fabric fork, where four in-world elytra runs (Overworld above the build limit, auto-jump, Nether below the roof and above it) pass end to end.

Let me know what you think, I can change any of this as needed.

src/main/java/baritone/cache/CachedRegion.java
```diff
@@ -44,7 +44,8 @@ public final class CachedRegion implements ICachedRegion {
     /**
      * Magic value to detect invalid cache files, or incompatible cache files saved in an old version of Baritone
      */
-    private static final int CACHED_REGION_MAGIC = 456022911;
+    // Public: the elytra pathfinder's region reader checks the same header.
```
babbaj: remove the comment

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -143,10 +136,10 @@ public void setBlock(int x, int y, int z, boolean solid) {
             this.filled[y >> 4] |= 1 << x8;
             s[off >>> 3] |= mask;
         } else {
+            // The x8's bit in filled stays. It means the cube may hold a block, so one that
+            // outlives its blocks costs a reader a scan of the cube and never a wrong answer;
+            // fillSection resets it.
             s[off >>> 3] &= ~mask;
-            if (allZero(s, x8 * (X8_BYTES / 8), X8_BYTES / 8)) {
```
babbaj: This is still needed because queueBlockUpdate may set random blocks to air but for performance this check should be done when updating individual blocks from queueBlockUpdate. In all other places blocks are set the entire chunk is being set at once and in those cases this check is not needed. Because this function is a bit complicated instead of creating a separate function just add an argument to this one that conditionally runs this code that is only set to try by queueBlockUpdate. then for convenience add another setBlock function with the same signature as we currently have but passes false for that new argument.

src/main/java/baritone/process/elytra/pathfinder/PathSegment.java
```diff
@@ -17,13 +17,16 @@
...
 package baritone.process.elytra.pathfinder;
 
+import java.util.List;
+import net.minecraft.core.BlockPos;
+
 public class PathSegment {
     public final boolean finished;
-    public final long[] packed;
-
+    /** The blocks of the path, in order. */
```
babbaj: remove the comment

babbaj (changes requested):

0Mattias:
Thanks again for the review. Pushed 8827e157. `setBlock` takes an `exact` flag now: a clear that empties its x8 also clears the cube's bit in `filled`, and only `queueBlockUpdate` passes true. The four-arg `setBlock` passes false, so the whole-chunk fills skip the scan. `ChunkTest` covers both kinds of clear. The two comments are gone, and I dropped four unused imports while I was in there.

babbaj (approved):

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -139,13 +128,14 @@ public void queueBlockUpdate(BlockChangeEvent event) {
...
             writeLock.lock();
             try {
-                long ptr = NetherPathfinder.getChunk(this.context, chunkPos.x, chunkPos.z);
-                if (ptr == 0) return; // this shouldn't ever happen
+                final Chunk chunk = this.context.getChunk(chunkPos.x, chunkPos.z);
+                if (chunk == null) return; // this shouldn't ever happen
                 event.getBlocks().forEach(pair -> {
                     BlockPos pos = pair.first().below(minY);
                     if (pos.getY() < 0 || pos.getY() >= 384) return;
                     boolean isSolid = pair.second() != AIR_BLOCK_STATE;
```
babbaj: while we're here we should also replace this with .isAir()

babbaj (changes requested):

src/main/java/baritone/process/elytra/pathfinder/Face.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package baritone.process.elytra.pathfinder;
+
+import net.minecraft.core.BlockPos;
+
+enum Face {
```
babbaj: this can be deleted and replaced with minecraft's `Direction`

babbaj (changes requested):

0Mattias:
Both in, pushed eb62fc04. `isAir()` in the block update and `AIR_BLOCK_STATE` is gone, and `Face` is deleted, `PathFinder` takes a `Direction` now and moves with `BlockPos.relative`. Just a heads up, I kept the neighbour array in the old order (UP, DOWN, NORTH, SOUTH, EAST, WEST) instead of `Direction.values()` so the A* search expands the same way it did before. Tests green, and the four in-world flights in my fork pass too.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,231 @@
...
+    static final int X8_BYTES = 64;
+    static final int X4_BYTES = 8;
+
+    /** A chunk with no blocks. Shared; writes to it throw. */
+    public static final Chunk AIR = new Chunk(false);
+    /** A chunk with every block solid. Shared; writes to it throw. */
+    public static final Chunk SOLID = new Chunk(true);
+
+    private final long[][] sections = new long[SECTIONS][];
+    /** Bit i of entry s is set once a block is set in x8 cube i of section s. fillSection resets it, an exact setBlock clears it with the cube's last block, and a plain clear leaves it: set means the cube may hold a block, clear that it does not. A section that is null has 0. */
```
ZacSharp: I had a hard time reading this even though the statement is rather simple. Maybe drop the description of which methods change it and just state what the bits mean. Something like
> Bit i of entry s indicates whether the x8 cube of section s might contain blocks. An unset bit means the cube is empty. A section that is null has 0.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,231 @@
...
+    private final int[] filled = new int[SECTIONS];
```
ZacSharp: Why is this an int array? Doesn't a section have just eight x8 cubes so we only need eight bits per entry?

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public boolean isEmpty(Size size, int x, int y, int z) {
+        switch (size) {
+            case X1: return !isSolid(x, y, z);
+            case X2: return isEmptyX2(x, y, z);
+            case X4: return isEmptyX4(x, y, z);
+            case X8: return isEmptyX8(x, y, z);
+            default: return isEmptyX16(y);
+        }
+    }
```
ZacSharp: This can return `false` for empty x8 cubes.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public boolean isEmptyX16(int y) {
+        return this.filled[y >> 4] == 0;
+    }
+
+    public boolean isEmptyX8(int x, int y, int z) {
+        return (this.filled[y >> 4] & (1 << x8Index(x, y, z))) == 0;
+    }
```
ZacSharp: This can return `false` for empty cubes.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
```diff
@@ -0,0 +1,231 @@
...
+            }
+            s = this.sections[y >> 4] = new long[SECTION_LONGS];
+        }
+        final int off = x2Offset(x, y, z);
+        final long mask = 1L << (((off & 7) << 3) + bitIndex(x, y, z));
+        final int x8 = x8Index(x, y, z);
+        if (solid) {
+            // the x8 is marked before its block is set, so a reader racing this write never skips a block it can see
+            this.filled[y >> 4] |= 1 << x8;
+            s[off >>> 3] |= mask;
```
ZacSharp: Is the JIT compiler / CPU not allowed to reorder the writes?
Might be we can get away without any extra synchronization, but I'm not familiar enough with this code to check.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
ZacSharp, line 192: According to the specification of the `filled` flags these can return false for empty x8 and x16 while the smaller sizes always work, is that intentional?
At a call site I'd certainly not expect this behavior ~and the pathfinder seems to rely on not getting false negatives~.

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,313 @@
...
+
+    static final int STATE_FROM_CALLER = 0;
+    static final int STATE_FAKE = 1; // could be generated or just air
+
+    /** A chunk in the table, where it is, and whether it came from the game or was made up. */
+    static final class Entry {
+        final Chunk chunk;
+        final int x;
+        final int z;
+        volatile int state;
```
ZacSharp: Should this be an enum or an `isFake`/`fromCaller` bool?

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+
+    // ---- the chunk table ----
+
+    /**
+     * Inserts a chunk from the game, replacing any chunk at that position. {@code data} has one
+     * boolean per block, indexed {@code y << 8 | z << 4 | x}, 16 * 16 * 256 of them (384 in the overworld).
+     */
+    public void insertChunkData(int chunkX, int chunkZ, boolean[] data) {
```
ZacSharp: This method appears to be unused. EDIT: it's used in tests. Maybe GitHub optimized some off-screen files away so browser search didn't find the usage.

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /** The chunk at (x, z), or null. */
+    public Chunk getChunk(int x, int z) {
+        final Entry e = this.chunks.get(key(x, z));
+        return e != null ? e.chunk : null;
+    }
+
+    /** Marks the chunk at (x, z) as from the game or as made up. Returns true if the chunk existed and the change was made. */
+    public boolean setChunkState(int x, int z, boolean fromCaller) {
```
ZacSharp: Perhaps mark this as only needed for tests?

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
```diff
@@ -0,0 +1,313 @@
...
+        final int regionZ = cz >> 5;
+        if (!this.checkedRegions.add(key(regionX, regionZ))) {
+            return 0;
+        }
+        final long t1 = System.currentTimeMillis();
+        final boolean read = BaritoneRegion.load(this.baritoneCache, regionX, regionZ,
+                (x, z, chunk) -> this.chunks.putIfAbsent(key(x, z), new Entry(STATE_FROM_CALLER, chunk, x, z)));
+        return read ? System.currentTimeMillis() - t1 : 0;
+    }
+}
```
ZacSharp: The comment says "for the search and rays" but unless I missed something all of these methods are only used in tests. Are they actually needed?

src/main/java/baritone/process/elytra/pathfinder/NodePos.java
```diff
@@ -0,0 +1,63 @@
...
+package baritone.process.elytra.pathfinder;
+
+import net.minecraft.core.BlockPos;
+
+/** A cube of the octree: its size and its position in units of that size. */
+final class NodePos {
+
+    final Size size;
+    /** absolute position divided by the width */
+    private final BlockPos pos;
```
ZacSharp: Idk whether the compiler is smart enough to notice this object doesn't escape and can be inlined. Might as well split it into `x`,`y`,`z` fields, it's not used for more than `getX`,`getY`,`getZ` anyway.

src/main/java/baritone/process/elytra/pathfinder/PathFinder.java
```diff
@@ -0,0 +1,410 @@
...
+     * cancelled.
+     */
+    static Path findPathSegment(NetherPathfinder ctx, NodePos start, NodePos goal, boolean x4Min, int timeoutMs, boolean airIfFake, double fakeChunkCost) {
+        final NetherPathfinder.CacheMiss fakeChunkMode = airIfFake ? NetherPathfinder.CacheMiss.AIR : NetherPathfinder.CacheMiss.GENERATE;
+        final Size minSize = x4Min ? Size.X4 : Size.X2;
+        final BlockPos goalCenter = goal.absolutePosCenter();
+        final BlockPos startCenter = start.absolutePosCenter();
+
+        final Search s = new Search(goalCenter, startCenter, fakeChunkCost);
+        final Set<Long> doneFull = new HashSet<>();
```
ZacSharp: Should this be some kind of `LongSet`?

src/main/java/baritone/process/elytra/pathfinder/Raytracer.java
```diff
@@ -0,0 +1,369 @@
...
+import net.minecraft.util.Mth;
+import java.util.ArrayList;
+import java.util.List;
+import net.minecraft.world.phys.Vec3;
+
+/**
+ * A ray through the chunk octrees: "An efficient parametric algorithm for octree traversal"
+ * (Revelles, Urena, Lastra), as the native Refiner.cpp had it. The traversal only walks positive
+ * directions, so the ray is reflected around each x16 node's centre and the child indices are
+ * flipped back with {@code a}.
```
ZacSharp: What's `a`?

src/main/java/baritone/process/elytra/pathfinder/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with Baritone.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * A pathfinder and raytracer for flying the Nether by elytra: babbaj's nether-pathfinder
+ * (https://github.com/babbaj/nether-pathfinder), which Baritone loaded as a native library
+ * through JNI, in plain Java. Each class is the C++ file of the same name, ported one for one,
+ * and the tests hold the terrain generator and the raytracer to the native library's own
+ * recorded answers over a few hundred generated chunks and a few thousand rays. Nothing in here
+ * refers to Minecraft; the seam is {@link baritone.process.elytra.NetherPathfinderContext},
```
ZacSharp: It uses `BlockPos` and `Direction`, no?

ZacSharp (commented):
I'm not at all familiar with the C++ codebase so I cannot say much about the port. I've tried to not leave comments which likely apply to the C++ code as well.

I've mostly skipped over the tests.

One thing to perhaps care about later is that this spams allocations. Some years ago benchmarks showed the main pathfinder was faster after removing every single `BlockPos` allocation. The JVM might since have gotten better at handling this kind of short-lived-allocation spam, but it might be worth a try once things have settled.

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
ZacSharp: I'd suggest "real", in line with the existing "fake chunk" terminology, but "from caller" works as well.

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
ZacSharp: Is there actually any place which calls this with `solid=false` and `exact=false` outside tests? From a quick search it looks like all callers either call `setBlock(x, y, z, solid, true)` or know that they are writing to an empty chunk and only call `setBlock(x, y, z, true)`, suppressing calls with `solid=false`.

babbaj: im pretty sure no

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
babbaj: It shouldn't matter but id like to know why fable thinks there might be a race condition

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
babbaj: if `filled` disagrees with the actual data then that's a bug with whatever wrote to the chunk

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
babbaj: yes but i just noticed setChunkState never gets called so this should also be final and not volatile

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
babbaj: it is not used and can be removed

src/main/java/baritone/process/elytra/pathfinder/NetherPathfinder.java
babbaj: it should be deleted

src/main/java/baritone/process/elytra/pathfinder/PathFinder.java
babbaj: yes

src/main/java/baritone/process/elytra/pathfinder/package-info.java
babbaj: yes, the comment is outdated

src/main/java/baritone/process/elytra/pathfinder/Chunk.java
babbaj: I believe you are correct

0Mattias:
Thanks both. All eight are in, pushed 1a3c2113..63a5813b, one commit per comment:

- `filled` is a `byte[]`, one byte per section, and its comment just says what the bits mean.
- The `exact` flag is gone. No whole-chunk writer ever clears a block, as Zac found, so the flag never skipped a scan for anyone, and without it the summary is exact on every clear: `isEmptyX8` and `isEmptyX16` can no longer say "not empty" for an empty cube. If they ever disagree with the blocks, that is a bug in whatever wrote them, as babbaj says.
- The "x8 is marked before its block" comment is gone too. To answer babbaj: there is no race. The comment was written as if `Chunk` had to be safe on its own, like the native code, but every reader in Baritone holds `NetherPathfinderContext`'s read lock and every writer its write lock, so nothing can watch a block being set. The class comment says that now instead.
- `Entry.state` is a `final boolean fromCaller`, `setChunkState` and `insertChunkData` are deleted, and the test helper generates its chunks itself. I kept "from caller" since it matches `hasChunkFromCaller`.
- `NodePos` holds three ints instead of a `BlockPos`, `doneFull` is a fastutil `LongOpenHashSet`, and the package comment names the Minecraft classes it uses.

On the allocations: the `NodePos` change drops one object per node the search creates. The rest I would rather look at once the review has settled, with a benchmark, as you say. Tests green, and the four in-world flights in my fork pass too.

babbaj:
There's still a lot of BlockPos objects being created, especially where they are made just to construct NodePos objects.

src/main/java/baritone/process/elytra/NetherPathfinderContext.java
```diff
@@ -46,33 +45,22 @@
...
-            throw new RuntimeException(ex);
-        }
-    }
-    private static final BlockState AIR_BLOCK_STATE = Blocks.AIR.defaultBlockState();
-    // This lock must be held while there are active pointers to chunks in java,
-    // but we just hold it for the entire tick so we don't have to think much about it.
+    // The native library needed this lock held while there were pointers to its chunks in Java.
+    // The port needs none of that -- a chunk is an object that stays valid for whoever holds it,
+    // and the table takes lookups, inserts and culls from any thread at once -- and the lock is
+    // kept so that the threads still run in the order they always have.
```
ZacSharp, line 56: Isn't this the lock that prevents writing to a chunk while search is reading it?

0Mattias:
Both in, pushed 63a5813b..40b4e3f9:

- **BlockPos** (babbaj): a `NodePos` is built from three ints now and gives its corner back as ints, so the search loop, the neighbour expansion, `findAir` and the heuristic don't make a `BlockPos` anywhere. The four sub cubes on a face are visited in place instead of returned in an array. The only `BlockPos` left is what leaves the search: the path and its start and goal.
- **The lock** (Zac): yes, it is, and that comment was wrong. Packing fills a chunk after the table already holds it, and a block update changes a chunk a search may be reading, so the lock is what keeps them apart. The comment now says who takes which side, and `Chunk`'s points to it.
- Merged 1.21.4 in. The only conflict was #5126's line in the native `writeChunkData`, which this PR deletes; `fillSection` already fills all 512 bytes.

The output doesn't change: 390 searches over generated terrain (the ones that finish before the 500 ms timeout) and 3000 `findAir` calls give the same paths and cubes, block for block. Over those searches:

| | before | after |
|---|---|---|
| allocated | 1348 MB | 891 MB |
| time | 1.26–1.32 s | 1.13–1.17 s |
| `BlockPos` and `BlockPos[]`, share of allocations | 37% | 0.05% (the path) |

Most of what's left is `NodePos` itself (69%, one per neighbour, as the map key) and boxed `Long` keys into the chunk table (12%). I'd leave those for the benchmark pass.

One correction to what I said last time: not every reader holds the read lock. `ElytraBehavior.setPath` checks the landing spot with a ray on the game thread without it, same as upstream today. With the port that ray can see a stale block but can't fail, since every read takes its section once and null checks it. With the native library, my fork traced its "raytrace whiffed" exits to that same ray (0Mattias/cheesecake#16), and I fixed it there by running the check on a worker under the read lock. Happy to send that as its own PR if you want it.

Tests green (79 with the merge), and the four in-world flights in my fork pass.

babbaj (approved):

ZacSharp:
Very welcome side effect: CI artifacts size just decreased by 83% (43.1MB to 7.37MB) :tada:

PR #5118, opened 2026-09-12
add automatic torch placement

### LINKED ISSUE #971: [Suggestion] Auto-place Torches
Good day people! I would like to suggest an auto-place function for torches so that while mining, you won't worry mobs spawning behind you or previous tunnels you built :)

#autotorch (on/off auto torch)
#autotorchradius (specifies what light level will the torch be placed on)

Adding this suggestion will make less worries for mobs while afk'ing. Although #avoidance has this, adding light while mining is better. It also guides your way.

kekkykekkek:
TorchProcess looks for air blocks within a small radius whose block light is at or below a threshold and that can support a torch, then aims and places one. It's modelled on BackfillProcess. Spacing falls out of the light level rather than needing its own setting once a torch is placed, what it lights up stops being a candidate.

Settings: autoTorch (off by default), autoTorchLightLevel (default 7), autoTorchRadius (default 4).

MovementHelper.attemptToPlaceABlock now takes the item to place, so a caller can place something other than a throwaway block. The existing signature delegates to the new one, so pathing behaviour is unchanged.

Built and tested in game on 26.2.
Closes [cabaletta/baritone#971](https://github.com/cabaletta/baritone/issues/971)


src/api/java/baritone/api/Settings.java
```diff
@@ -668,6 +668,21 @@ public final class Settings {
...
+
+    /**
+     * Place a torch when the block light level is at or below this
+     */
+    public final Setting<Integer> autoTorchLightLevel = new Setting<>(7);
+
+    /**
+     * How far from the player to look for dark spots to place torches in
+     */
+    public final Setting<Integer> autoTorchRadius = new Setting<>(4);
```
ZacSharp: This should either mention that `autoTorchRadius` has to be less than the `blockReachDistance`, or you could just reuse the reach distance instead of adding a new setting.

src/main/java/baritone/process/TorchProcess.java
```diff
@@ -0,0 +1,171 @@
...
+        if (recentlyPlaced.containsKey(pos)) {
+            return false;
+        }
+        if (!ctx.world().getBlockState(pos).isAir()) {
+            return false;
+        }
+        if (ctx.world().getBrightness(LightLayer.BLOCK, pos) > Baritone.settings().autoTorchLightLevel.value) {
+            return false;
+        }
+        return Blocks.TORCH.defaultBlockState().canSurvive(ctx.world(), pos);
```
ZacSharp: Does this work with wall torches?

src/main/java/baritone/process/TorchProcess.java
```diff
@@ -0,0 +1,171 @@
...
+            return false;
+        }
+        recentlyPlaced.values().removeIf(tick -> now() - tick > PLACEMENT_COOLDOWN);
+        if (now() - lastPlacement < PLACEMENT_COOLDOWN) {
+            return false;
+        }
+        if (!selectTorch(false)) {
+            return false;
+        }
+        return !darkSpots().isEmpty();
```
ZacSharp, line 90: This scans twice per tick. While I agree that that's not pretty either, maybe assign the scan result to a field which is then read by `onTick`?

src/api/java/baritone/api/Settings.java
```diff
@@ -668,6 +668,21 @@ public final class Settings {
      */
     public final Setting<Boolean> backfill = new Setting<>(false);
 
+    /**
+     * Place torches on dark blocks near the player
+     */
+    public final Setting<Boolean> autoTorch = new Setting<>(false);
```
ZacSharp, line 674: Could actually get rid of this setting by making `autoTorchLightLevel > 15` disable the feature. But there might be benefit to having a separate off switch, so idc.

ZacSharp (changes requested):
This pull request should target the 1.21.4 branch. Features are added there first and then merged to the branches for newer version.

kekkykekkek:
okay, ill work on getting onto the 1.21.4 branch now. ill also do some testing for things you have mentioned like the wall torches.

kekkykekkek:
i believe i have done the change to 1.21.4 correctly, if not let me know. 
many thanks

src/api/java/baritone/api/Settings.java
kekkykekkek: it should now use block reach distance

src/main/java/baritone/process/TorchProcess.java
kekkykekkek: it didn't, does now

ZacSharp (commented):
Still scans twice per tick. Other than that it looks fine.

ZacSharp:
https://github.com/user-attachments/assets/87031812-4df1-4f4e-bcc0-5e2e534e636e

Don't think it's supposed to work like that. I doesn't reliably let go of rightclick and places too many torches.

kekkykekkek:
What is the light level set at?

ZacSharp:
Default value, so seven.

PR #5120, opened 2026-09-14
Add optional head hitter sprint jumping

AverWasTaken:
Supersedes #5105: per cbd182b the primary development branch is `1.21.4` now, so the changes were ported onto it (only real adjustment was the Mojang-mapping rename `isOnGround()` → `onGround()`). The old PR closed as a side effect of the branch move and GitHub doesn't let it reopen, hence this one. `clearOfLedgesAhead` also now bails when the support under the next two positions isn't physically there yet, addressing the last review comment on the old PR.

---

New `headHitters` setting, off by default like other optional behavior settings.

When enabled, Baritone sprint jumps in 1x2 corridors and under any low ceiling on flat traverses. The sprint jump speed boost applies on the first ticks of the jump before we bonk our head, so hopping through a tunnel is faster than plain sprinting.

Implementation is entirely in `PathExecutor.shouldSprintNextTick` since that's where the sprint policy already lives. It forces `Input.JUMP` the same way the traverse to straight ascend skip does, and because it runs after `movement.update()` reasserts the inputs, the forced jump only lasts one tick so there's no state to clean up.

Some guards it has, mostly found by breaking it in game:
* it waits until we're fully inside the corridor (same idea as `skipNow`) so we don't jump a block early and bonk on the face of the entrance block, which just stops us
* no headhitting while sneaking (magma edge safety) or while the movement still has blocks to break (breaking is like 5x slower when you're jumping)
* no headhitting when the path goes down right after the movement, or when the two blocks past the destination have no floor. The bonk momentum carries an extra block or two and will happily throw us off a ledge next to the goal
* that ledge check originally had an off by one that crashed the client when the current movement was the last on the path, also gone now

This is execution only, the cost model is untouched (same as `sprintAscends` and `overshootTraverse`), so path choice doesn't change, we just arrive a bit sooner.

Tested in a dev env on fabric: tunnels, overhangs, goals next to dropoffs.


leijurv:
This looks cool. Anyone else want to give an opinion?

AverWasTaken:
#5105 zac said it was cool

AverWasTaken:
^^ removed the "Sprint jump in a 1x2 corridor" from features.md as this PR would.. well.. add it. yippe

0alter0:
I approve of this message
- Lil Skittle 2026

PR #5122, opened 2026-09-15
Compare block state property values with equals, not reference identity

pirateben820:
### What

`BuilderProcess.sameBlockstate` compares each property's value with `!=`:

```java
Map<Property<?>, Comparable<?>> map1 = first.getValues();
Map<Property<?>, Comparable<?>> map2 = second.getValues();
for (Property<?> prop : map1.keySet()) {
    if (map1.get(prop) != map2.get(prop)
```

The values are boxed `Comparable`s, so this is reference identity. It is fine for `EnumProperty` and
`BooleanProperty`, whose values are singletons. For an `IntegerProperty` it happens to work today because every
vanilla integer property stays inside the `-128..127` range the JVM's `Integer` cache interns — I checked: the largest,
`age`, tops out at 25. Nothing in the comparison guarantees that, and identity is not what is meant here; equality is.

`Objects.equals` to match the three other comparisons in this file. Plain `equals` would also be correct, since both
states share one `StateDefinition` and the lookups cannot return null.

This branch is only reached when `buildIgnoreDirection` or `buildIgnoreProperties` is set (both default off);
otherwise the early `first.equals(second)` return is taken.

### Verification

Compiles on this branch (`./gradlew compileJava`, Gradle 8.7, Temurin JDK 21.0.12.1). GitHub's own runs are waiting on first-contributor approval. Not run in game - it is a one-line change to a comparison.

### Authorship

This patch and this description were written by an AI assistant (Claude, via Claude Code) working under my direction,
while I was debugging my own mod, which drives Baritone from a language model. I am submitting it and I answer for it.
If the project would rather not take AI-written changes, say so and I will close it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pirateben820:
Retargeted to 1.21.4 as requested on #5123. The change had to be re-expressed rather than rebased: on 1.21.4 the comparison is `map1.get(prop) != map2.get(prop)` over `getValues()`, so it is `Objects.equals` there. The description now describes that code; the earlier one described the 26.2 shape. Compiles on 1.21.4 (Gradle 8.7, JDK 21).

PR #5123, opened 2026-09-15
Let the game exit: daemon threads for the shared pool, and wait for the final cache save

pirateben820:
### The symptom

With Baritone installed the client does not exit when the window closes. It sits in shutdown until Mojang's own
`ClientShutdownWatchdog` kills it about fifteen seconds later, with

```
java.lang.Error: Watchdog (Client shutdown from post-main)
```

and on Windows the leftover `javaw` holds a lock on the mods folder until then.

### The cause

`Baritone`'s shared pool is created with the default thread factory, so every worker is **non-daemon** — and none of
them ever exit:

```java
threadPool = new ThreadPoolExecutor(4, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS, new SynchronousQueue<>());
```

- Core size is 4 and `allowCoreThreadTimeOut` is never set, so four workers park in `take()` for the life of the JVM
  whether or not they have work.
- Two of them are permanently busy anyway: `CachedWorld`'s `PackerThread` (`while (true) { toPackQueue.take(); … }`)
  and its ten-minute save loop.

Once `Minecraft.main` returns those are the only non-daemon threads left, so the JVM cannot exit. A thread dump at
shutdown shows exactly them:

```
"pool-4-thread-1" prio=5 WAITING on ...LinkedBlockingQueue.take
        at baritone.o$a.run(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker
"pool-4-thread-2" TIMED_WAITING
"pool-4-thread-3" WAITING on ...SynchronousQueue$Transferer
"pool-4-thread-4" WAITING on ...SynchronousQueue$Transferer
```

(If the elytra process has been used in the session there are also its own single-thread executors —
`ElytraBehavior.solverExecutor` and the two in `NetherPathfinderContext` — which this PR does not touch.)

### The change

A daemon `ThreadFactory` for the pool. The threads are also named `Baritone-N`, so a dump reads as such rather than
`pool-4-thread-N`. That is the whole diff.

### The trade-off, and #5116

Daemon threads mean a periodic cache save that happens to be mid-write at exit can be cut off. `CachedRegion.save`
writes in place, so that could leave a truncated region file; `CachedRegion.load` already tolerates a bad file, and the
region is re-cached.

**#5116 by @kekkykekkek predates this PR by four days, fixes the same hang, and also adds a JVM shutdown hook
(`shutdownNow` + a bounded `awaitTermination`) that covers that trade-off.** I did not find it before opening this,
and I should have. If you would rather take #5116, please close this one. If a version on 1.21.4 is useful, I am happy
for this to carry it, with credit to that PR.

### What has and has not been verified

Reproduced on the `26.2` branch (Baritone 1.19.0, Minecraft 26.2, Fabric Loader 0.19.5, Java 25), where the patch
compiles. I have not run a patched jar in game. This 1.21.4 branch compiles as well (`./gradlew compileJava`, Gradle 8.7, Temurin JDK 21.0.12.1). GitHub's own runs are waiting on first-contributor approval, so that local build is the only compile of this exact tree so far.

An earlier revision of this PR also made `WorldData.onClose` wait for the final save. That was wrong on two counts —
`onClose` is not reached at window close (the quit path goes through `disconnect`, which never calls `setLevel`), and
where it *is* reached, on dimension change and rejoin, the wait would have blocked the client thread — so it is gone.

### Authorship

This patch and this description were written by an AI assistant (Claude, via Claude Code) working under my direction,
while I was debugging my own mod, which drives Baritone from a language model. I am submitting it and I answer for it.
I had it independently reviewed after opening, which is what turned up the `onClose` mistake and #5116. If the project
would rather not take AI-written changes, say so and I will close it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


leijurv:
rebase on main dev branch 1.21.4

pirateben820:
Rebased onto 1.21.4 — thanks. Same two files, applied cleanly; the pool and `WorldData.onClose` are identical on that branch.

I also retargeted #5122 (the `sameBlockstate` property comparison) to 1.21.4. That one needed reworking rather than rebasing: on 1.21.4 the comparison is written as `map1.get(prop) != map2.get(prop)` over `getValues()`, so it is `Objects.equals` there.

pirateben820:
Two corrections to this PR, both from an independent review I had done after opening it.

1. **#5116 by @kekkykekkek predates this by four days and fixes the same hang** — daemon threads plus a shutdown hook. I missed it and should have searched first. If you would rather take that one, please close this; if a 1.21.4 version is useful, I am happy for this to carry it with credit.
2. The earlier revision here also made `WorldData.onClose` wait for the final cache save. That was wrong: `onClose` is not reached at window close (the quit path goes through `disconnect`, which never calls `setLevel`), and where it *is* reached — dimension change, rejoin — the wait would have blocked the client thread for up to 10 s. It is removed. What remains is the daemon factory, which is the part that actually fixes the exit, with the comment corrected to name the idle core threads as well as the two loops.

The description is rewritten to match, and this branch compiles on 1.21.4 (Gradle 8.7, JDK 21).

PR #5124, opened 2026-09-16
Make the pathfinder consider more movements per second

AverWasTaken:
## What this PR does

Caches the stuff A* was recomputing over and over and removes a guava cache lookup from the hot path. Same paths, found 1.2x to 1.6x faster. No cost function changed!

## Why it should be merged

More movements considered per second means the search gets further inside the same `primaryTimeoutMS` before it has to hand back a segment, so fewer short segments and fewer "no path found" on stuff that was actually reachable. And it does that without touching the cost model, so the executor's per-tick cost verification still agrees with the planner exactly. Everything new is a memo in front of code that already existed, which makes it easy to bisect if something looks off.

## What changed

A* was spending most of its time doing the same thing over and over. The block under a node gets read by 21 of the 22 movements out of it, its break cost gets computed by the four descends and then again by the neighbours' descends, and every one of those goes through the registry's identity hash map just to find flags we already computed.

The bigger ones:

- Nodes live in 16x8x16 cells now, with only the cells in a hash map. The flat position -> node map was the hottest thing in the profile, every lookup was a cache miss into a multi megabyte key array. Successors are right next to the node being expanded so they mostly land in the cell we just touched.
- `canPlaceAgainst` went through `Block.isShapeFullBlock`, which is a guava cache that allocates a queue node on every hit. 8% of the search and nearly half of all allocation, mostly to ask whether air is a full cube. It's a precomputed flag now. get rekt guava
- The precomputed flags are stored on the `BlockState` itself via a small mixin instead of a `Reference2IntOpenHashMap` lookup, and the table is shared across searches instead of starting empty every tick. If the mixin isn't applied it falls back to the old lookup.
- The pathing thread's `BlockStateInterface` caches block lookups by position, and `getMiningDurationTicks` is memoized by position too.
- The heap keeps its keys in a parallel `double[]` so the sifts don't chase pointers. That one alone was ~15% which honestly surprised me.

## Results

Numbers are from an offline benchmark running the real `AStarPathFinder` against a synthetic 1024x1024 world put into actual `LevelChunkSection`s, 3 second budget, median nodes per second:

| scenario | before | after |
|---|---|---|
| surface goto | 129k | 152k |
| surface goto with favoring | 129k | 163k |
| underground goto | 130k | 185k |
| mine-style composite, 64 goals | 127k | 210k |
| descend to y=-40 | 133k | 185k |

So 1.2x on flat ground and up to 1.6x anywhere it has to dig. yippee. The runs swing ~8% so don't stare at any single row too hard. Also the mixin path can't run in the benchmark (no mixins there), that column is measuring the fallback, in game should be a bit better still.

## Testing

Existing unit tests pass (the open set tests check the heap against the linked list one), plus the usual light smoke testing in a dev env.

---

> AI was used in this PR to verify the math behind the benchmark comparisons, sanity-check assumptions around the caching changes, and help review the implementation for correctness. Performance testing was done and validated manually.


leijurv:
very cool.

`java.util.Arrays.fill` instead import Arrays plz

Can u have ur guy also check if we can get a win from making the costs not be doubles? Float is certainly good enough, and honestly, costs could probably work just fine as quantized ints? Maybe "1" would mean 0.001 seconds or some such.

leijurv:
Also why guard against the mixin failing to apply? Is that realistic?

AverWasTaken:
Unrelated PR??? @leijurv ^^ anyways im working on ur questions

AverWasTaken:
fixed the `Arrays` import.

on the mixin guard: it isn't realistic in game, the mixin always applies there. the fallback exists because the pathfinder can now run in a plain jvm with no mixins (that's how the benchmark works, and it's handy for tests). i dropped the identity cache since that was the ugly part, what's left is one `static final boolean` checked once at class init so it's free either way. if you'd rather it just hard cast and blow up without the mixin i can do that too.

on float / quantized ints: i tried float on the search side only (`PathNode` cost fields, the heap keys, the bestSoFar array) with the movement functions still returning doubles and casting at the boundary. same machine, 4 runs each, median nodes/s:

| scenario | double | float |
|---|---|---|
| surface goto | 187k | 198k |
| surface goto + favoring | 182k | 196k |
| underground goto | 186k | 208k |
| descend to y=-40 | 180k | 186k |

so about 5-10%. it's real but it's not the block reads, which are still most of the time. precision looks fine, costs top out in the thousands of ticks and `MIN_IMPROVEMENT` is 0.01, float has ~1e-4 to spare there. the one thing to watch is `Path.runBackwards` taking `min(calculateCost, nodeCostDiff)`, with float sums that diff can come in a hair under the true cost, which only ever makes the executor's check stricter so it's harmless.

quantized ints would probably land in the same place (same 4 bytes, integer compares are a little cheaper) but they touch the whole api, `IMovement.getCost`, `ActionCosts`, `Goal.heuristic` are all doubles. felt like its own PR rather than piling it on this one. happy to do it as a follow up if you want it, i'd go float on the search side first since it's contained to three classes.

PR #5114, opened 2026-09-11
26.2 Elytra Boost Module Integration and 512 Byte Fix

GPSolo:
This pull request adds a one-line fix to fix nether navigation and prevent falsely identifying incomplete air pockets as paths and also adds a MeteorClientCompat to allow baritone to use ElytraBoost directly, second changes were made purely to my own necessity and are not necessary, as even with ElytraBoost enabled, using a firework uses the ElytraBoost module and prevents fireworks from being consumed. I do not expect or advise the changes for the ElytraBoost module to be merged, unless the team thinks it has some merit, the main purpose of the pull request is the 512 Byte Fix.

PR #5125, opened 2026-09-16
Move from 2 Byte Nether Pathfinding to 512 Byte Temporarily

GPSolo:
New PR Made in Response to #5114, this PR includes only the 512 Byte changes. This should be considered to temporarily fix pathfinding issues in 26.2 until the new implementation.

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -25,7 +25,7 @@
   ],
   "depends": {
     "fabricloader": ">=0.19.3",
-    "minecraft": ["26.1.*"]
+    "minecraft": ["26.2"]
   },
   "custom": {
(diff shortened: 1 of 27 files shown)


AverWasTaken:
you said this "includes only the 512 Byte changes" but the diff also adds the `elytraBoostModule` setting to Settings.java. nothing in this PR reads that setting, it's a leftover from #5114 pointing at meteor's Elytra Boost module which isn't here. dead setting in a one line fix PR. drop the Settings.java hunk and this is just the NetherPathfinderContext change.

leijurv:
does this belong on 1.21.4?

GPSolo:
> does this belong on 1.21.4?

It's on 26.2 branch?

Fixed the Settings.java last PR is #5126

Issue #5111, opened 2026-09-06
Selection Clear Area does not automove and gets stuck


johanbebeboy:
## Some information
Operating system: Windows 11 24H2
Java version: java-runtime-epsilon
Minecraft version: 26.2
Baritone version: 1.19.0
Other mods (if used): 
Meteor 26.2 
Xaeros minimap
3d skin layers
chat heads
cloth config 26.2
distant horizons
durability tooltip
entityculling
essential-container
fabricapi
fabriclanguagekotlin
ferritecore
gammautils
immediatelyfast
iris
lithium
modmenu
placeholderapi
simplevoicechat
sodium
SuperMartijn642's Config Library

## Modified settings
RandomLooking = 0
RandomLooking133 = 0

## Final checklist
- [x] I know how to properly use check boxes 
- [x] I have included the version of Minecraft I'm running, baritone's version and forge mods (if used).
- [] I have included logs, exceptions and / or steps to reproduce the issue.
- [x] I have not used any OwO's or UwU's in this issue.

## Image
<img width="657" height="402" alt="Image" src="https://github.com/user-attachments/assets/f35007cd-aad3-4f38-8eab-41cf7706e02a" />

## Comment
i dont know why this happens, just want to clear out an area using selection ClearArea

ZacSharp:
Just some random ideas because I don't know what causes this
1. Is there cave air around?
2. Have you tried removing some of your mods? Either distant horizons or another mod with similar features was incompatible in the past. Sodium requires explicit support code in Baritone, which might have broken at some point.

PR #5128, opened 2026-09-17
Fix builder stalling on satisfied break goals

Zepo4ka:
Fixes #5111.

## Root cause

BuilderProcess could return a composite containing a GoalBreak that already considered the player's current position to be in-goal. This happens after the direct break scan has found no ray-reachable target. Since GoalComposite uses any-member semantics, PathingBehavior saw the whole builder goal as already satisfied and did not begin path calculation, leaving selection clear/fill-air operations active but stationary.

## Fix

Exclude break goals already satisfied at the player's current feet when assembling the builder path goal. Direct breaking has already been attempted at that point, so those goals cannot currently make progress. Other break targets remain in the composite and start normal pathing; skipped positions are reconsidered after the player moves.

The change is local to BuilderProcess and does not alter normal goto/path behavior or process arbitration.

## Testing

- Added BuilderProcessTest.satisfiedUnreachableBreakGoalDoesNotPreventPathing
- `./gradlew test` (42 tests, 0 failures)
- `./gradlew :forge:build` (Minecraft 26.2 Forge standalone distribution succeeds)


ZacSharp:
1. Wrong branch
2. Isn't this just going to loop instead?
3. That helper should be private.
4. Your LLM might only be capable of adding and running useless unit tests, but that doesn't mean changes do not need in-game testing.

PR #5129, opened 2026-09-17
Write cache and waypoint files atomically

kekkykekkek:
CachedRegion and WaypointCollection opened the real file for writing, which truncates it before the new contents arrive, so a save that does not finish leaves a half written file behind. Waypoint loading catches the resulting exception and ignores it, silently dropping every waypoint of that tag, and a region load logs a stack trace and loses the cache.

Write to a sibling .tmp file and move it into place instead, so an interrupted save leaves the previous file intact. The empty region file that save created up front is gone too, since it would itself be left behind as a corrupt file if a first save were cut off.


leijurv:
lowkey slopped up but fine

PR #5132, opened 2026-09-20
Polish the path, goal and search rendering

AverWasTaken:
# Polish the path, goal and search rendering (wow so pretty)

---

## What this PR does

Same floating line, same goal box, same colors more or less, but antialiased, rounded at the corners and faded at the ends, and the path calculation grows and fades on screen instead of snapping to a different path every frame. Rendering only, nothing in pathing is touched.

## Why it should be merged

The renderer is the one part of Baritone everybody looks at all day, and it's still 5 pixel aliased gl lines with gaps at every joint and a search display that flickers between unrelated paths. This keeps the look people know and cleans it up. It's all behind two settings, `renderPathRibbon` and `renderBoxFill`, and with both off the old code runs exactly as it did, so it's easy to compare and easy to back out. The new drawing lives in two new classes, the old line renderer is kept whole as the fallback.

## What changed

The path was `RENDERTYPE_LINES`, which can't fade across its width, aliases, and leaves a notch at every joint once it's thick. The search display drew "best path so far" and "path to the most recent node" fresh each frame, but that's sixty samples a second out of a few hundred thousand nodes a second, so what you actually saw was two unrelated paths teleporting around.

- `PathRibbon`: the path as a strip of quads facing the camera. Width is still `pathRenderLineWidthPixels` in real screen pixels (converted per vertex from view depth and the projection matrix, so it follows the fov), with a one pixel feather on each edge. Corners get two rounds of Chaikin, joints are shared vertices so no gaps, and the bit of path behind the player fades out over 2.5 blocks. `fadePath` works as before.
- Steps are drawn across, up, across. The old straight diagonal from the bottom of an ascend to the top passes 0.21 of a block deep through the corner of the block being stepped onto, which is very visible with `renderPathIgnoreDepth` off. The riser sits over the lower block, 0.2 short of the step's face, and that's enough that the rounded corner never enters it. Checked numerically for ascend, descend, the diagonal versions, falls and staircases.
- `SearchGlow`: remembers what it drew last frame. The best path eases toward its new length so the tip glides, and when the search switches branch the dropped one fades out from the fork over 450 ms while the new one grows out of it. The most recent nodes are thin light blue twigs, each drawn from its latest fork with anything already on screen (the best path or an older twig), so the whole thing is one connected tree and nothing gets traced twice. Samples that would need more than 20 nodes to connect are skipped, so there's never a long line thrown across the screen. Twigs hanging off a dropped branch go down with it.
- Goal box and the break / place / walk into boxes get a faint fill and thinner, more solid outlines. The goal's fill fades upward from the floor and breathes on the same two second cycle as the rings. `drawGoal` now collects its boxes first and draws after, one pass of quads and one of lines, which also replaces the nullable buffer + `setupRender` batching for composites.
- `renderPathAnimated`: a faint bit of extra opacity drifting along the path toward the goal, so you can tell which way it goes. An earlier version had proper bright pulses and they were distracting the moment you tried to look at anything else, so this is deliberately barely there.
- Default colors are softer versions of the same hues (current path a deeper red, search in two blues, goal a minty green). Anyone who set their own keeps them. This is the one part that changes the look for people who don't touch settings, so if you'd rather keep the primaries i can drop it from the PR.

New settings: `renderPathRibbon`, `renderPathAnimated`, `renderBoxFill`, all default on. `renderPathAsLine` only applies with the ribbon off.

## Preview

https://github.com/user-attachments/assets/38f1d925-1a96-4213-8608-3d1bba9bb6eb

## Testing

Unit tests green (44, clean checkout, JDK 21) and the main source set compiles. Looked at in a dev client over a lot of rounds of tweaking, mostly walking paths with climbs in them, the search display, and `renderPathIgnoreDepth` on and off. The elytra path also goes through the ribbon via the public `drawPath` and I haven't flown with it yet, so that one is untested. Known limit: where the line crosses over itself on screen with depth ignored it blends twice, same as the gl lines did.


AverWasTaken:
oh btw this change isnt integrated into #5127 so uhhh ill fix that later if we ever merge it

AverWasTaken:
+857 lines was a bit much for this change so I trimmed it down as much as I could :( 

vv

leijurv:
okay i'm NGL i really do like the spastic flickering blue lines while its computing though 😇 

general cleanups are welcome. i'm not against making it look cute in general

however i do recall a similar PR in the long distant past - the reason why I didn't like it then was that the current path renderer has an actual height to it - it's not just a line that faces the camera, it faces the vertical plane of the world. 

i would want to see way more recordings. can you post more videos, each 10mb i kno, but at full resolution? even if the framerate is lower or the clips are shorter

AverWasTaken:
flickering search is back as the default. the smooth one (best path growing, recent nodes branching off it) is behind a new setting now, `renderSearchSmooth`, default false. it only does anything with `renderPathRibbon` on, otherwise you get the old gl lines for everything like before

on the height: yeah that was a fair catch, v3 was just a camera facing line. it has the frickin thingy back now, same 0.03 tall rectangle standing in the vertical plane of the path as the old renderer, just drawn as one soft band instead of 4 gl lines. from the side you see the full height on top of the line width, from straight above (or on the vertical bits of steps) it's just the line. `renderPathAsLine` still turns the height off

preview: https://www.youtube.com/watch?v=1MoT2F7Dx90

AverWasTaken:
and lowkey we should prob just leave the default for search smooth to on since not everyone cares really about the cool looking computing of the path ( no offense ) that way u can just toggle what u want but it looks clean for 90% of people.

